PR #8, opened 2019-03-13
Add support for configuring an import callback

mwylde:
This PR lets clients customize import resolution ([as supported in the C library](https://github.com/google/jsonnet/blob/30f07a2684ebc40d800bb1948e1f5987e5fb02e3/include/libjsonnet.h#L67)). This is accomplished with an optional param to the evaluator that resolves a (scope, import path) to a filesystem path.

We use this to support non-filesystem paths (in particular HTTP), and it seems like the equivalent support in the C and go APIs are fairly widely used.


lihaoyi-databricks:
looks great

Issue #14, opened 2019-04-15
Incomplete cli interface implementation


FUSAKLA:
Hi, thanks for the initiative the speed is amazing!

I'd love to migrate to this jsonnet implementation but it does not completely implement the cmd CLI interface.

It's missing:
 - `-c / --create-output-dirs  Automatically creates all parent directories for files`
 - `-S / --string           Expect a string, manifest as plain text`

Also it's missing support for multiple file output and the `std.manifestYamlDoc` (the yaml is not important).

It there any intention for implementing those? It would be really appreciated.

Issue #9, opened 2019-03-18
the default CLI output is incomplete , shows a lot of '???'


xask00:
# ./sjsonnet.jar 
error: Need to pass in a jsonnet file to evaluate
usage: sjsonnet [sjsonnet-options] script-file

  -i, --interactive  Run Mill in interactive mode, suitable for opening REPLs and taking user input
  -n, --indent       How much to indent your output JSON
  -J, --jpath        Specify an additional library search dir (right-most wins)
  -o, --output-file  Write to the output file rather than stdout
  -V, --ext-str      ???
  --ext-str-file     ???
  --ext-code         ???
  --ext-code-file    ???
  -A, --tla-str      ???
  --tla-str-file     ???
  --tla-code         ???
  --tla-code-file    ???

lihaoyi:
Haha yeah, this is just me being sloppy. These flags are all fully implemented, just not documented properly. A PR to copy over the messages from the Go or C++ Jsonnet help output would be welcome!

PR #15, opened 2019-04-22
Implement `std.manifestDocYaml` and `create-output-dirs`, `string`, `multi` command line options

pvid:
Closes https://github.com/databricks/sjsonnet/issues/14 https://github.com/databricks/sjsonnet/issues/9

This pull request implements new features:

* implements `std.manifestYamlDoc` and `std.manifestYamlStream`
* `create-output-dirs`, `string`, `multi` command line options
* documents previously undocumented flags (replaces `???`)

and fixes some differences between the reference implementation and `sjsonnet`:

* `std.substr` does not care about being out of bounds - it just produces an empty output for characters out of bounds
* `std.manifestJson` does not contain spaces between `\n` characters in empty objects and arrays
* incorrect escaping of `~` has been fixed in https://github.com/google/jsonnet/pull/564
* `std.trace` outputs to stderr

NOTICE: a potentially breaking change is the addition of the `create-output-dirs` flag. `sjsonnet` previously behaved as if it were enabled, which differed from default behaviour of `jsonnet`

The number of lines changes seems huge, but most of it is caused by addition of tests for `std.manifestYamlDoc` in ` sjsonnet/test/resources/test_suite/stdlib.jsonnet`

### DIFF
--- a/sjsonnet/src/sjsonnet/Cli.scala
+++ b/sjsonnet/src/sjsonnet/Cli.scala
@@ -48,7 +48,7 @@ object Cli{
     ),
     Arg[Config, String](
       "ext-str", Some('V'),
-      "???",
+      "<var>[=<val>] Provide 'external' variable as string. 'If <val> is omitted, get from environment var <var>",
       (c, v) => v split('=') match{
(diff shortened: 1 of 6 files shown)


FUSAKLA:
Great, thanks a lot @pvid for tackling this!

lihaoyi-databricks:
Looks good to me, thanks!

Issue #36, opened 2019-10-01
std.sort does not accept second parameter


javaduke:
When trying to sort an array of objects using `std.sort` with second parameter, e.g.
`std.sort(myArray, function(x) x.weight)` I'm getting this error:

> sjsonnet.Error: Too many args, function has 1 parameter(s)

javaduke:
and looks like there are many more functions that don't take key function parameter.

yehaolan:
Same with `std.set(...)`

My jsonnet file has following code:
```
local a = {
  name: "123",
  id: "abc",
};

local b = {
  name: "123",
  id: "aaa",
};

local l = [a, b];

std.set(l, function (o) std.toString(o))
```
Then, I will get `sjsonnet.Error: Too many args, function has 1 parameter(s)`

lihaoyi-databricks:
This appears to be a new feature in jsonnet, with many of the `std.*` functions now taking an additional `keyF` argument with a default value of `id`.

@sparkprime @sbarzowski do you guys have any test suite for calls to these functions with the additional `keyF` parameter? I didn't see any such calls in the https://github.com/google/jsonnet/blob/master/test_suite/stdlib.jsonnet test files, but if you guys have a test suite I'd rather re-use yours rather than trying to come up with my own

sbarzowski:
Hmmm... it appears in the benchmark, in the docs and tests specific to go-jsonnet, but it's missing from the main test suite. I have created an issue to add some: https://github.com/google/jsonnet/issues/716

javaduke:
> This appears to be a new feature in jsonnet, with many of the `std.*` functions now taking an additional `keyF` argument with a default value of `id`.

@lihaoyi I've implemented this functionality in my fork, if you want, feel free to take a look, and I can create a PR for you.
The functions I modified are:
`sort()`
`uniq()`
`set()`
`setMember()`
`setUnion()`
`setInter()`
`setDiff()`

lihaoyi-databricks:
@javaduke yes I would love a pull request!

PR #40, opened 2019-10-09
Key functions

javaduke:
@lihaoyi-databricks As discussed, this PR fixes the issue #36. I also made a few minor optimizations to the code.

I will have a separate PR which adds regex functions to the sjsonnet but it will require Mill issue #601 fix (which I already have a PR for: https://github.com/lihaoyi/mill/pull/706)


lihaoyi-databricks:
Looks good to me, thanks!

PR #44, opened 2019-10-21
Non-Boolean Conditions fix

javaduke:
An expression like [x for x in [1,2,3] if 'a'] in Jsonnet results in error "Condition must be boolean, got string" - but sjsonnet does not fail. This PR fixes it.


lihaoyi:
@javaduke can you submit the PR without all the unrelated reformatings?

javaduke:
Ah, sorry, will do. Apparently IntelliJ does all this automatically, I didn't even realize the formatting is different. Sorry about it.

lihaoyi:
Looks good, thanks!

PR #48, opened 2019-12-01
Added GZip Standard Function

agrawroh:
## Description
Added a standard function to gzip the given data (String/Bytes Array) and return a Base64 encoded gzipped string back.

## Usage
`std.gzip(...)`


lihaoyi-databricks:
Looks good to me!

@sparkprime do we have any process for standardizing this sort of standard library addition? I'm gonna merge this for now since we'll use this internally, but I wouldn't be surprised if other people would want something similar and it would be nice to keep our standard libraries in sync

PR #52, opened 2020-02-12
Preserving Ordering of Fields: Alternative

fugu13:
I figured it was easier to show some of the possible adjustments than talk about them, here's the old code I wrote (now rebased onto the most recent version)


fugu13:
Opened to the wrong place by default, closing!

Issue #67, opened 2020-05-20
std.rstripChars not supported


igorwwwwwwwwwwwwwwwwwwww:
I tried to run sjsonnet on our existing setup and got:

```
sjsonnet.Error: Field does not exist: rstripChars
```

It's from the stdlib: https://github.com/google/jsonnet/blob/29fd57097fba5bb9fa90038e855b1bd1ee0901c0/stdlib/std.jsonnet#L63.

There's probably other things missing for parity too, e.g. https://github.com/databricks/sjsonnet/issues/66.

lihaoyi-databricks:
Fixed by https://github.com/databricks/sjsonnet/pull/79

PR #79, opened 2020-07-13
Strip Chars functions implemented

javaduke:
Addresses https://github.com/databricks/sjsonnet/issues/67 , https://github.com/databricks/sjsonnet/issues/80,

### DIFF
--- a/sjsonnet/src/sjsonnet/Std.scala
+++ b/sjsonnet/src/sjsonnet/Std.scala
@@ -326,13 +326,13 @@ object Std {
     },
 
     builtin("rstripChars", "str", "chars"){ (ev, fs, str: String, chars: String) =>
-      rstripChars(str, chars.toList)
+      str.replaceAll("[" + chars + "]+$", "")
     },
     builtin("lstripChars", "str", "chars"){ (ev, fs, str: String, chars: String) =>
-      lstripChars(str, chars.toList)
+      str.replaceAll("^[" + chars + "]+", "")
     },
     builtin("stripChars", "str", "chars"){ (ev, fs, str: String, chars: String) =>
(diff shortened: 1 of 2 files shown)


sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -324,6 +324,17 @@ object Std {
...
     },
+
+    builtin("rstripChars", "str", "chars"){ (ev, fs, str: String, chars: String) =>
+      str.replaceAll("[" + chars + "]+$", "")
+    },
+    builtin("lstripChars", "str", "chars"){ (ev, fs, str: String, chars: String) =>
+      str.replaceAll("^[" + chars + "]+", "")
+    },
+    builtin("stripChars", "str", "chars"){ (ev, fs, str: String, chars: String) =>
+      str.replaceAll("[" + chars + "]+$", "").replaceAll("^[" + chars + "]+", "")
```
lihaoyi-databricks, line 335: Would I be right to say that this won't work if `chars` contains `]`? I think an approach using `str.dropWhile(!chars.contains(_)).reverse.dropWhile(!chars.contains(_)).reverse` would avoid that problem

lihaoyi-databricks: Using `indexWhere(!chars.contains(_))` and `lastIndexWhere(chars.contains(_))` to shorten the string would probably work too, and much faster than `dropWhile` and `takeWhile`. We should also make sure to convert `chars` into a `Set` to ensure fast lookup

javaduke: Ah, you were absolutely right, my bad - but `Regex.quote()` takes care of all special chars :)

lihaoyi-databricks:
Looks reasonable to me I think

Issue #80, opened 2020-07-13
Missing implementation of std.member


markj-db:
https://jsonnet.org/ref/stdlib.html

std.member(arr, x)
Available since version 0.15.0.
Returns whether x occurs in arr. Argument arr may be an array or a string.

This is used in https://github.com/jsonnet-libs/xtd

Version 0.15.0 introduced several new functions in the std library:

New stdlib functions:
std.member
std.clamp
std.stripChars, std.lstripChars, std.stripChars
std.flatMap
std.repeat

https://github.com/google/jsonnet/releases

markj-db:
This is another incompatibility with standard jsonnet https://github.com/databricks/sjsonnet/issues/73

markj-db:
And is a duplicate of https://github.com/databricks/sjsonnet/issues/66

lihaoyi-databricks:
Fixed by https://github.com/databricks/sjsonnet/pull/81

PR #82, opened 2020-07-15
std.flatMap implemented

javaduke:
https://github.com/databricks/sjsonnet/issues/80

### DIFF
--- a/sjsonnet/src/sjsonnet/Std.scala
+++ b/sjsonnet/src/sjsonnet/Std.scala
@@ -295,7 +295,7 @@ object Std {
             builder.append(
               fres match {
                 case fstr: Val.Str => fstr.value
-                case Val.Null => throw Error.Delegate("flatMap func result must not be null")
+                case Val.Null => ""
                 case x => throw Error.Delegate("flatMap func must return string, got " + fres.asInstanceOf[Val].prettyName)
               }
             )
--- a/sjsonnet/src/sjsonnet/Std.scala
+++ b/sjsonnet/src/sjsonnet/Std.scala
@@ -274,6 +274,40 @@ object Std {
         }
       )
     },
+    builtin("flatMap", "func", "arr"){ (ev, fs, func: Applyer, arr: Val) =>
+      val res: Val = arr match {
+        case a: Val.Arr =>
+          val arrs = Val.Arr(
+            a.value.map{ i =>
+              Val.Lazy(func.apply(i))
+            }
+          )
+          val out = collection.mutable.Buffer.empty[Val.Lazy]
+          for(x <- arrs.value){
+            x.force match{
+              case Val.Null => // do nothing
+              case Val.Arr(v) => out.appendAll(v)
+              case x => out.append(Val.Lazy(x))
+            }
+          }
+          Val.Arr(out.toSeq)
+
+        case s: Val.Str =>
+          val builder = new StringBuilder()
+          for (c: Char <- s.value) {
+            val fres = func.apply(Val.Lazy(Val.Str(c.toString)))
+            builder.append(
+              fres match {
+                case fstr: Val.Str => fstr.value
+                case x => Materializer(fres)(ev).toString
+              }
+            )
+          }
+          Val.Str(builder.toString)
+      }
+      res
+    },
+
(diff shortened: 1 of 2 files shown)


sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -274,6 +274,40 @@ object Std {
...
+        case a: Val.Arr =>
+          val arrs = Val.Arr(
+            a.value.map{ i =>
+              Val.Lazy(func.apply(i))
+            }
+          )
+          val out = collection.mutable.Buffer.empty[Val.Lazy]
+          for(x <- arrs.value){
+            x.force match{
+              case Val.Null => // do nothing
```
lihaoyi-databricks, line 288: From what I can tell, this `case Val.Null =>` check is unnecessary, since jsonnet.org simply errors out if the second param `arr` is an array and `func` returns `null`. Given that, we might be able to implement this simply as 

```scala
case a: Val.Arr => 
  val results = a.value.map(func.appy(i).asInstanceOf[Val.Arr].value)
  val arrResults = results.collect{case a: Val.Arr => a}
  if (arrResults.size != results.size) {
    val arrResults = results.filter{!_.isInstanceOf[Val.Arr]}
    throw new Error.Delegate("flatMap func must return an array, not " + arrResults.mkString(", "))
  }
  Val.Arr(results.flaten)
```

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -274,6 +274,40 @@ object Std {
...
+          Val.Arr(out.toSeq)
+
+        case s: Val.Str =>
+          val builder = new StringBuilder()
+          for (c: Char <- s.value) {
+            val fres = func.apply(Val.Lazy(Val.Str(c.toString)))
+            builder.append(
+              fres match {
+                case fstr: Val.Str => fstr.value
+                case x => Materializer(fres)(ev).toString
```
lihaoyi-databricks, line 302: 1. We do not need this catch-all clause, since when the second parameter `arr` is a string, `func` returninganything except a string or `null` is an error

2. We *do* need a case to handle `null` here, since it appears that when the second parameter `arr` is a string, returning `null` means nothing gets added to the output string.

javaduke: Well technically nothing stops a user from implementing a function that returns something other than string. I think catch-all should throw an error saying "The response of the function parameter must be a string" and I will add a null check. What do you think?

lihaoyi-databricks: My proposed behavior above is according to what the upstream google/jsonnet behavior is, from what I can tell on jsonnet.org.

```jsonnet
> std.flatMap(function (x) 123, "helo")
RUNTIME ERROR: expected string but arr[0] was number
	std.jsonnet:271:7-77	function <anonymous>
	example1.jsonnet:1:1-38	
```

Although it's always possible to be a bit more accepting, my goal for this repo is to try and match google/jsonnet behavior as much as possible, including in the error cases on a best effort basis.

lihaoyi-databricks:
@javaduke on jsonnet.org, `flatMap` on a `String` can accept the function returning `null`, and it treats it as removing that character from the string. We should preserve this behavior

```jsonnet
> std.flatMap(function (x) if x == " " then null else x, "a b c d e")
"abcde"
```

sjsonnet/src/sjsonnet/Std.scala
javaduke: Yes, agree completely, it is all fixed now and I added test for this type of error.

lihaoyi-databricks:
Here's a bit of an odd behavior on jsonnet.org:

```jsonnet
> std.flatMap(function (x) if x == " " then null else x, ["a", " ", "b"])
"[ ]anullb"
```

It seems we can return strings even when `flatMap`ing over lists

javaduke:
> It seems we can return strings even when `flatMap`ing over lists, and it uses the logic from `+` to concatenate them together

Hmm, I do understand why this happens, but is it a correct behavior or just poorly written code?

lihaoyi-databricks:
@javaduke good question, let's just ignore that edge case for now and if anyone wants it we can add it later

lihaoyi-databricks:
@javaduke this looks good, if you're ok with it I'll merge it

javaduke:
All looks good on my side :) Feel free to merge :)

PR #84, opened 2020-07-28
Use passed streams instead of standard streams

t0il3ts0ap:
On overriding the print stream, errors are still getting printed on console.


szeiger:
I ran into the same bug and fixed it in be895232e1c7b203f52a9d467c0e81236253bdfa last week before I saw this PR so I'm closing this. I suppose it's still marked as mergeable because it's identical to my version.

PR #85, opened 2020-08-21
Write output directly to disk with -o

szeiger:
This prevents caching the entire output string in memory, reducing the heap usage for our largest output file of 7.2 MB from 101 MB to 93 MB in my test.

Only JSON is supported for now. YAML requires more work because the YamlRenderer is more closely tied to the underlying StringWriter, so it still has to buffer.

### DIFF
--- a/sjsonnet/src-jvm/sjsonnet/SjsonnetMain.scala
+++ b/sjsonnet/src-jvm/sjsonnet/SjsonnetMain.scala
@@ -1,6 +1,7 @@
 package sjsonnet
 
-import java.io.{BufferedOutputStream, InputStream, OutputStreamWriter, PrintStream, StringWriter}
+import java.io.{BufferedOutputStream, InputStream, OutputStreamWriter, PrintStream, StringWriter, Writer}
+import java.nio.charset.StandardCharsets
 import java.nio.file.NoSuchFileException
 
 import sjsonnet.Cli.Config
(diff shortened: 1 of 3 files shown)


sjsonnet/src-jvm/sjsonnet/SjsonnetMain.scala
```diff
@@ -105,15 +106,25 @@ object SjsonnetMain {
...
+    def writeFile(f: os.RelPath, contents: String): Either[String, Unit] =
+      handleWriteFile(os.write.over(os.Path(f, wd), contents, createFolders = config.createDirs))
+
+    def writeToFile[U](f: os.RelPath, contents: Writer => Either[String, U]): Either[String, Unit] =
+      handleWriteFile(os.write.over.outputStream(os.Path(f, wd), createFolders = config.createDirs)).flatMap { out =>
+        try {
+          val buf = new BufferedOutputStream(out)
+          val wr = new OutputStreamWriter(buf, StandardCharsets.UTF_8)
+          val u = contents(wr)
+          wr.flush()
```
lihaoyi-databricks, line 124: Do we need to call `buf.flush()` here?

lihaoyi-databricks:
One comment, looks good to me otherwise. Feel free to merge once that's resolved!

sjsonnet/src-jvm/sjsonnet/SjsonnetMain.scala
szeiger: No, flushing (like closing) is transitive, you only need to call it once for the first stream/writer in the pipeline.

Issue #69, opened 2020-06-03
Sjsonnet does not error when duplicate fields are present in an object


mgyucht:
Reproduction:
```
$ cat test.jsonnet
{ a: 1, a: 2 }
$ ./sjsonnet.jar test.jsonnet
{
   "a": 2
}
$ jsonnet test.jsonnet
STATIC ERROR: test.jsonnet:1:9: duplicate field: a
```

PR #86, opened 2020-08-24
Detect duplicate fields in the parser

szeiger:
- The error message has a rather unhelpful position but this is consistent with the way other static errors are currently handled by Sjsonnet. The original Jsonnet parser does it better.

- There are two different places in the Jsonnet codebase where duplicate fields are detected: Statically in the parser and dynamically in the VM. It's not clear to me without digging much deeper into their C++ implementation when the second one is supposed to kick in. There is no specification or test case for the dynamic detection.

Fixes https://github.com/databricks/sjsonnet/issues/69

### DIFF
--- a/sjsonnet/src/sjsonnet/Parser.scala
+++ b/sjsonnet/src/sjsonnet/Parser.scala
@@ -270,7 +270,18 @@ object Parser{
 
   def objinside[_: P]: P[Expr.ObjBody] = P(
     member.rep(sep = ",") ~ ",".? ~ (forspec ~ compspec).?
-  ).map{
+  ).flatMap { case t @ (exprs, _) =>
+    val seen = collection.mutable.Set.empty[String]
+    var overlap: String = null
+    exprs.foreach {
+      case Expr.Member.Field(_, Expr.FieldName.Fixed(n), _, _, _, _) =>
+        if(seen(n)) overlap = n
+        else seen.add(n)
+      case _ =>
+    }
+    if (overlap == null) Pass(t)
+    else Fail.opaque("no duplicate field: " + overlap)
+  }.map{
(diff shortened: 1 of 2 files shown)


sjsonnet/test/src-jvm/sjsonnet/ErrorTests.scala
```diff
@@ -303,5 +303,8 @@ object ErrorTests extends TestSuite{
 //        |    at .(sjsonnet/test/resources/test_suite/error.invariant.simple3.jsonnet:18:10)
 //        |""".stripMargin
 //    )
+    test("duplicate_fields") - check(
```
lihaoyi-databricks, line 306: Could we put this in `EvaluatorTests` instead? Currently all the test cases in `test_suite/` that are used in `FileTests.scala` and `ErrorTests.scala` are directly vendored from the google/jsonnet implementation, so it's best to keep our own custom additions separate so next time we update the vendored code we don't accidentally lose our additions

lihaoyi-databricks:
One small comment, but feel free to merge once resolved

sjsonnet/test/src-jvm/sjsonnet/ErrorTests.scala
szeiger: `EvaluatorTests` seems like a strange place for a parser test.

I also assumed that the test suite should be the original Jsonnet one, but then I saw that it had a few changes applied over time and it doesn't match the current Jsonnet test suite.

lihaoyi-databricks: How about `ParserTests`? Not much in there now, but if this is a parser test then it seems like it belongs

szeiger:
Test moved to ParserTests.

Issue #87, opened 2020-08-24
Disallow adjacent object literals


szeiger:
Requested by @iuliand-db:

> Forbid syntactic sugar that allows omitting the `+` to merge to objects when both LHS and RHS are dictionary literals. That’s basically never what you want, and usually indicates a comma between the two literals.

Since Jsonnet allows this, simply making it an error is probably not an option, but we could add a warning or a strict mode (neither of which currently exists in Sjsonnet).

PR #88, opened 2020-08-25
Add strict mode

szeiger:
This makes adjacent object literals an error. Fixes https://github.com/databricks/sjsonnet/issues/87


iuliand-db:
I wonder if this isn't too strict, perhaps we should disallow it only inside lists? Wondering how well this would fare on our existing body of jsonnet. :)

lihaoyi-databricks:
@iuliand-db I'd say this is probably OK, as we have the flexibility to adjust our existing body of Jsonnet accordingly. The golden ZZZ files should make it pretty easy to notice all the places where we need to tweak, and it should be easy to add a `+` in the cases where it's intentional.

PR #89, opened 2020-08-25
Update for release 0.2.7

szeiger:
@lihaoyi-databricks Any dependencies we should bump?


szeiger:
Merging and creating a new release as just discussed with @lihaoyi-databricks

PR #90, opened 2020-09-07
Add a pretty YAML output format

lihaoyi-databricks:
A lot of tools seem to like YAML, and directly outputting YAML might be significantly simpler and more performant than passing it through some third-party converter.

YAML has many ways of expressing the same data, so I somewhat arbitrarily picked  the PyYAML output style as a standard to follow. This style tends to leave strings and dictionary-keys unquoted where-ever possible, and tries to pick intelligently between the different string formats (naked, `'`-delimited, `"`-delimited, `|`, `|+`, `|-`, etc.), as well as wrapping them to about 80 characters wide (though not exactly). Internally we had previously used Python/PyYAML to generate YAMLs from the JSON output, and Python is generally pretty prevalent so we're probably not the only ones

YAML output is under a new flag `--yaml-out`. I also took the opportunity to refactor `SjsonnetMain.scala` to make it a bit more readable: split out nested functions into top-level functions, kill some unused parameters, etc.

I've optimized `PrettyYamlRenderer` sufficiently that it doesn't really turn up on my JProfiler profiles when I run this on our Databricks configuration codebase. It could always be micro-optimized some more, but it's probably fast enough for now.

In terms of compatibility, when re-generating 17,020,496 lines of YAML in our config codebase, there were 3 lines of diff related to PyYAML and this PR handling leading `---`s and trailing `...` differently in certain cases. Other than that the output of the two implementations are byte-for-byte identical.

In our internal config generation pipeline, our rough benchmarks indicate this is about a 2x speedup removing the intermediate JSON in the Jsonnet -> JSON -> YAML pipeline.


sjsonnet/src/sjsonnet/Materializer.scala
```diff
@@ -29,7 +29,7 @@ object Materializer {
         val arrVisitor = visitor.visitArray(xs.length, -1)
         for(x <- xs) {
           arrVisitor.visitValue(
-            apply0(x.force, visitor),
```
lihaoyi-databricks, line 32: This was a bug in the existing implementations of `Materializer.scala`. The contract of `upickle.core.Visitor` is that you're meant to call `.subVisitor` and pass that in to recursive calls, to give the implementer a chance to hook in. This fixes `Materializer.scala` to be compliant

sjsonnet/src/sjsonnet/YamlRenderer.scala
```diff
@@ -0,0 +1,554 @@
...
+
+import java.io.{StringWriter, Writer}
+import java.util.regex.Pattern
+
+import upickle.core.{ArrVisitor, ObjVisitor}
+import ujson.BaseRenderer
+
+
+
+class YamlRenderer(out: StringWriter = new java.io.StringWriter(), indentArrayInObject: Boolean = false,
```
lihaoyi-databricks, line 11: I broke `YamlRenderer` into a separate file unchanged, since `Renderer.scala` was getting a bit large

sjsonnet/src/sjsonnet/YamlRenderer.scala
```diff
@@ -0,0 +1,554 @@
...
+
+/**
+ * A version of YamlRenderer that tries its best to make the output YAML as
+ * pretty as possible: unquoted strings, de-dented lists, etc. Follows the PyYAML
+ * style. Also adds the ability to stream writes to a generic `Writer`.
+ */
+class PrettyYamlRenderer(out: Writer = new java.io.StringWriter(),
+                         indentArrayInObject: Boolean = false,
+                         indent: Int = 2,
+                         idealWidth: Int = 80) extends BaseRenderer[Writer](out, indent){
```
lihaoyi-databricks, line 129: There's some duplication between this and `YamlRenderer`, but both implementations are pretty gnarly and I didn't manage to unify them

sjsonnet/src-jvm/sjsonnet/SjsonnetMain.scala
```diff
@@ -159,41 +174,39 @@ object SjsonnetMain {
...
+                val suffix = if (isScalar(single)) "\n..." else ""
+                single.transform(renderer).toString + suffix
+              case multiple =>
+                multiple.zipWithIndex
+                  .map{
+                    case (v, i) =>
+                      val renderer = rendererForConfig(new StringWriter(), config)
+                      val rendered = v.transform(renderer).toString
+                      if (isScalar(v)) "--- " + rendered
+                      else if (i != 0) "---\n" + rendered
```
lihaoyi-databricks, line 193: This changes the streaming JSON style for scalar values to `"--- "`-prefixed rather than `"---\n"`-prefixed. While the old style was compatible, this new style is a lot more compact and more closely matches what other implementations like PyYAML do

sjsonnet/src-jvm/sjsonnet/SjsonnetMain.scala
```diff
@@ -159,41 +174,39 @@ object SjsonnetMain {
...
-                val docs = renderedFiles.collect{case Right(yaml) => yaml}
-                val stream = if (docs.isEmpty) "" else docs.mkString("---\n", "\n---\n", "\n...")
-                Right[String, String](stream)
-              case errs =>
-                Left[String, String]("rendering errors:\n" + errs.mkString("\n"))
+            val stream = arr.value.toSeq match{
+              case Nil => ""
+              case Seq(single) =>
+                val renderer = rendererForConfig(new StringWriter(), config)
+                val suffix = if (isScalar(single)) "\n..." else ""
```
lihaoyi-databricks, line 184: Now we only add the trailing `...` to single-element outputs, matching what PyYAML does

sjsonnet/src/sjsonnet/YamlRenderer.scala
```diff
@@ -0,0 +1,554 @@
...
+  var newlineBuffered = false
+  var dashBuffered = false
+  var afterColon = false
```
lihaoyi-databricks, line 132: `YamlRenderer` uses backtracking to go back and delete any spaces that are followed by a newline. In order to make `PrettyYamlRenderer` work with opaque `Writer`s that might not support such backtracking, I instead buffer up the space-after-colon in this `afterColon` boolean and only flush it to `out` when I know I am writing something that's not a newline following the colon

sjsonnet/src/sjsonnet/YamlRenderer.scala
```diff
@@ -0,0 +1,554 @@
...
+  var afterKey = false
+  var topLevel = true
+  var leftHandPrefixOffset = 0
```
lihaoyi-databricks, line 135: This new variable is used to store the width of the LHS of `<key>:` syntax, so that if it's followed by a string that needs wrapping, I can subtract the width of `<key>` to make sure the string wraps at the appropriate place

sjsonnet/src/sjsonnet/YamlRenderer.scala
```diff
@@ -0,0 +1,554 @@
...
+  var firstElementInArray = false
```
lihaoyi-databricks, line 136: This variable is used to support rendering the start of nested arrays inline:

```yaml
- - a
  - b
- - c
  - d
```

Rather than the status quo:

```yaml
-
  - a
  - b
-
  - c
  - d
```

sjsonnet/src-jvm/sjsonnet/SjsonnetMain.scala
```diff
@@ -99,34 +134,14 @@ object SjsonnetMain {
...
       config.tlaBinding,
       OsPath(wd),
-      importer = resolveImport(
-        config.jpaths.map(os.Path(_, wd)).map(OsPath(_)),
-        allowedInputs
-      ),
+      importer = importer match{
+        case Some(i) => (wd: Path, str: String) => i(wd, str).map(p => (OsPath(p), os.read(p)))
+        case None => resolveImport(config.jpaths.map(os.Path(_, wd)).map(OsPath(_)), allowedInputs)
+      },
```
lihaoyi-databricks, line 140: This fixes a regression where we accidentally stopped using `importer` once we started using `config.jpaths` to validate imports. This change makes it such that `importer` takes precedence over `config.jpaths` if passed in

sjsonnet/src/sjsonnet/YamlRenderer.scala
```diff
@@ -0,0 +1,554 @@
...
+    }
+    dashBuffered = false
+    newlineBuffered = false
+    dashBuffered = false
+  }
+  override def visitArray(length: Int, index: Int) = new ArrVisitor[Writer, Writer] {
+    var empty = true
+    val dedentInObject = afterKey && !indentArrayInObject
+    def subVisitor = {
+      if (empty){
```
lihaoyi-databricks, line 216: In both `visitArray` and `visitObject`, we move some of the logic into when `subVisitor` is called the first time in order to avoid it being run if the array or object is empty. This is because empty arrays and objects are rendered as inline `[]`s or `{}`s, pretty different from the multi-line `key:`-delimited object or `- `-delimited list syntax, but due to the nature of the visitor we do not know whether an array or object is empty until someone calls `visitEnd`

sjsonnet/src/sjsonnet/PrettyYamlRenderer.scala
```diff
@@ -0,0 +1,434 @@
...
+
+import ujson.BaseRenderer
+import upickle.core.{ArrVisitor, ObjVisitor}
+
+/**
+ * A version of YamlRenderer that tries its best to make the output YAML as
+ * pretty as possible: unquoted strings, de-dented lists, etc. Follows the PyYAML
+ * style. Also adds the ability to stream writes to a generic `Writer`.
+ */
+class PrettyYamlRenderer(out: Writer = new java.io.StringWriter(),
```
lihaoyi-databricks, line 14: There's some duplication between this and YamlRenderer, but both implementations are pretty gnarly and I didn't manage to unify them

sjsonnet/src/sjsonnet/PrettyYamlRenderer.scala
```diff
@@ -0,0 +1,434 @@
...
+                         indentArrayInObject: Boolean = false,
+                         indent: Int = 2,
+                         idealWidth: Int = 80) extends BaseRenderer[Writer](out, indent){
+  var newlineBuffered = false
+  var dashBuffered = false
+  var afterColon = false
```
lihaoyi-databricks, line 20: `YamlRenderer` uses backtracking to go back and delete any spaces that are followed by a newline. In order to make `PrettyYamlRenderer` work with opaque Writers that might not support such backtracking, I instead buffer up the space-after-colon in this `afterColon` boolean and only flush it to `out` when I know I am writing something that's not a newline following the colon

sjsonnet/src/sjsonnet/PrettyYamlRenderer.scala
```diff
@@ -0,0 +1,434 @@
...
+  var afterKey = false
+  var topLevel = true
+  var leftHandPrefixOffset = 0
```
lihaoyi-databricks, line 23: This new variable is used to store the width of the LHS of `<key>:` syntax, so that if it's followed by a string that needs wrapping, I can subtract the width of `<key>` to make sure the string wraps at the appropriate place

sjsonnet/src/sjsonnet/PrettyYamlRenderer.scala
```diff
@@ -0,0 +1,434 @@
...
+  var firstElementInArray = false
```
lihaoyi-databricks, line 24: This variable is used to support rendering the start of nested arrays inline:

```yaml
- - a
  - b
- - c
  - d
```

Rather than the status quo:

```yaml
-
  - a
  - b
-
  - c
  - d
```

sjsonnet/src/sjsonnet/PrettyYamlRenderer.scala
```diff
@@ -0,0 +1,434 @@
...
+      out.append("- ")
+    }
+    dashBuffered = false
+    newlineBuffered = false
+    dashBuffered = false
+  }
+  override def visitArray(length: Int, index: Int) = new ArrVisitor[Writer, Writer] {
+    var empty = true
+    val dedentInObject = afterKey && !indentArrayInObject
+    def subVisitor = {
```
lihaoyi-databricks, line 103: In both `visitArray` and `visitObject`, we move some of the logic into when `subVisitor` is called the first time in order to avoid it being run if the collection is empty. This is because empty collection are rendered as inline `[]`s or `{}`s, pretty different from the multi-line `key:`-delimited object or `- `-delimited list syntax, e.g. needing a newline after the delimiter for non-empty collections but a space after the delimiter for empty collections.

Due to the nature of the visitor API we do not know whether an array or object is empty until someone calls `visitEnd`, so we need to shoehorn this logic into `subVisitor` for the non-empty case and `visitEnd` for the empty case.

sjsonnet/src/sjsonnet/PrettyYamlRenderer.scala
```diff
@@ -0,0 +1,434 @@
...
+import upickle.core.{ArrVisitor, ObjVisitor}
+
+/**
+ * A version of YamlRenderer that tries its best to make the output YAML as
+ * pretty as possible: unquoted strings, de-dented lists, etc. Follows the PyYAML
+ * style. Also adds the ability to stream writes to a generic `Writer`.
+ */
+class PrettyYamlRenderer(out: Writer = new java.io.StringWriter(),
+                         indentArrayInObject: Boolean = false,
+                         indent: Int = 2,
```
szeiger, line 16: `indent` gets initialized with `config.indent` which defaults to 3. Should the default depend on the output format instead?

sjsonnet/src/sjsonnet/PrettyYamlRenderer.scala
```diff
@@ -0,0 +1,434 @@
...
+                         idealWidth: Int = 80) extends BaseRenderer[Writer](out, indent){
+  var newlineBuffered = false
+  var dashBuffered = false
+  var afterColon = false
+  var afterKey = false
+  var topLevel = true
+  var leftHandPrefixOffset = 0
+  var firstElementInArray = false
+  val newlinePattern = Pattern.compile("\n")
```
szeiger, line 25: There's already a copy of this in `YamlRenderer` and it is used below (along with other `YamlRenderer` methods)

sjsonnet/src-jvm/sjsonnet/SjsonnetMain.scala
```diff
@@ -159,41 +174,39 @@ object SjsonnetMain {
...
-          new Renderer(wr, indent = config.indent)
-        )
-        config.outputFile match{
-          case None =>
-            materialize(new StringWriter).map(_.toString)
-          case Some(f) =>
-            val filePath = os.FilePath(f) match{
-              case _: os.Path => os.Path(f).relativeTo(os.pwd)
-              case _ => os.RelPath(f)
+            config.outputFile match{
```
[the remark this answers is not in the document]
lihaoyi-databricks: I didn't convert this code path to streaming output generation because it didn't come up in the profiles. We can do that in a follow up, since it's kind of orthogonal to the whole direct-YAML-generation thing

sjsonnet/src/sjsonnet/PrettyYamlRenderer.scala
lihaoyi-databricks: Good catch, it's dead

sjsonnet/src-jvm/sjsonnet/SjsonnetMain.scala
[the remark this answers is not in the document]
lihaoyi-databricks: Good idea, I'll do that

sjsonnet/src-jvm/sjsonnet/SjsonnetMain.scala
lihaoyi-databricks: Not sure, but in either case leading `---`s and trailing `...`s are optional in YAML docs so either way is valid YAML

sjsonnet/src/sjsonnet/PrettyYamlRenderer.scala
lihaoyi-databricks: Good question, we currently explicitly pass in `--indent 2` in our own codebase, so it's a bit of a moot point. In general, we never initialize any of the `FooRenderer`s without passing in an explicit `indent` anywhere in this codebase, so maybe we can just get rid of the default...

Issue #83, opened 2020-07-18
Wrong error message calling function in other file with wrong arguments


lihaoyi-databricks:
bar.jsonnet

```jsonnet
function(x) x
```

foo.jsonnet

```jsonnet
local f = import "bar.jsonnet";
f(1, x = 2)
```

```text
$ ./sjsonnet.jar foo.jsonnet
sjsonnet.Error: Function parameter f passed more than once
    at .(foo.jsonnet:2:2)
    at .(foo.jsonnet:2:2)
```

PR #97, opened 2020-11-04
Fix error message for too many arguments with at least one named arg

edsilfer:
## What changes are proposed in this pull request?
This PR fixes [issue 83](https://github.com/databricks/sjsonnet/issues/83) by changing the error message thrown when two parameters are passed to a function that was expecting only one. The message now matches the official [jsonnet.org](https://jsonnet.org) specification.

## How is this tested?
- A new unit test with a reproducible scenario of the issue
- `./mill "sjsonnet[2.13.3].jvm.test.testLocal"`
- Manually tested to check the behavior when importing the function from another file

### DIFF
--- a/sjsonnet/src/sjsonnet/Val.scala
+++ b/sjsonnet/src/sjsonnet/Val.scala
@@ -287,7 +287,8 @@ object Val{
       Error.failIfNonEmpty(
         repeats,
         outerOffset,
-        (plural, names) => s"Function parameter$plural $names passed more than once"
+        (plural, names) => s"binding parameter a second time: $names",
+        Some(defSiteFileScope)
       )
 
       Error.failIfNonEmpty(
--- a/sjsonnet/test/src-jvm/sjsonnet/ErrorTests.scala
+++ b/sjsonnet/test/src-jvm/sjsonnet/ErrorTests.scala
@@ -139,7 +139,7 @@ object ErrorTests extends TestSuite{
     )
 
     test("function_duplicate_arg") - check(
-      """sjsonnet.Error: Function parameter x passed more than once
+      """sjsonnet.Error: binding parameter a second time: x
         |    at .(sjsonnet/test/resources/test_suite/error.function_duplicate_arg.jsonnet:17:21)
         |    at .(sjsonnet/test/resources/test_suite/error.function_duplicate_arg.jsonnet:17:21)
         |""".stripMargin
@@ -147,12 +147,6 @@ object ErrorTests extends TestSuite{
     test("function_duplicate_param") - check(
       """Parse error: Expected no duplicate parameter: x:17:14, found ") x\n"""".stripMargin
     )
-//    test("function_infinite_default") - check(
-//      """sjsonnet.Error: Parameter passed more than once: x
-//        |    at .(sjsonnet/test/resources/test_suite/error.function_duplicate_arg.jsonnet:17:2)
-//        |    at .(sjsonnet/test/resources/test_suite/error.function_duplicate_arg.jsonnet:17:21)
-//        |""".stripMargin
-//    )
     test("function_too_many_args") - check(
       """sjsonnet.Error: Too many args, function has 2 parameter(s)
         |    at .(sjsonnet/test/resources/test_suite/error.function_too_many_args.jsonnet:19:4)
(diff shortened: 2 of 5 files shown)


sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -212,6 +212,13 @@ object Val{
         .defaults
         .map{ case (index, default) => (index, Lazy(evalDefault(default, newScope, evaluator)))}
 
+      // Check if expected args match passed args before  
+      if (params.args.length < args.length) {
```
iuliand-db, line 216: Why not make this more general, if the number of arguments `!=`, instead of only for too many arguments? The error would be something like `"Wrong number of arguments, expected ${params.args.length} but ${args.length} found"`

sjsonnet/test/src-jvm/sjsonnet/ErrorTests.scala
```diff
@@ -288,20 +288,12 @@ object ErrorTests extends TestSuite{
          |    at .(sjsonnet/test/resources/imports/error.too_many_arg.jsonnet:3:6)
          |""".stripMargin
     )
-    //    test("overflow") - check(
```
iuliand-db, line 291: @lihaoyi-databricks was there a reason to have these commented out? If yes, we should add a `TODO:` so we know why.

lihaoyi-databricks: Honestly I don't remember, but I assume it's because I never managed to get them passing. Looking at https://github.com/databricks/sjsonnet/commit/c874391d3f81915001e2d2eae26f54cc33a2c998 which introduced the other tests in this suite, it appears these tests were introduced already-commented-out, so we could probably just delete them

sjsonnet/src/sjsonnet/Val.scala
edsilfer: Apparently, there are valid use cases in which the number might be different, however, I couldn't catch any case where the number is greater. If we change the comparison to `!=`, the following tests start to fail:

```
X sjsonnet.StdWithKeyFTests.stdSetMemberWithKeyF 185ms
X sjsonnet.StdWithKeyFTests.stdSortWithKeyF 1ms
X sjsonnet.StdWithKeyFTests.stdUniqWithKeyF 1ms
X sjsonnet.StdWithKeyFTests.stdSetWithKeyF 1ms
X sjsonnet.StdWithKeyFTests.stdSetUnionWithKeyF 1ms
X sjsonnet.StdWithKeyFTests.stdSetInterWithKeyF 1ms
X sjsonnet.StdWithKeyFTests.stdSetDiffWithKeyF 1ms
X sjsonnet.FileTests.functions 11ms
X sjsonnet.FileTests.invariant_manifest 2ms
X sjsonnet.FileTests.parsing_edge_cases 2ms
X sjsonnet.FileTests.stdlib 99ms
X sjsonnet.FileTests.tla.simple 1ms
X sjsonnet.PreserveOrderTests.preserveOrderYaml 0ms
+ sjsonnet.PreserveOrderTests.preserveOrderXml 0ms
X sjsonnet.PreserveOrderTests.preserveOrderPreservesSetMembership 0ms
X sjsonnet.ErrorTests.invariant.option 1ms
X sjsonnet.ErrorTests.import_wrong_nr_args 1ms
X sjsonnet.ErrorTests.too_many_arg_with_named_arg 0ms
+ sjsonnet.FormatTests.X 2ms
X sjsonnet.EvaluatorTests.functions 0ms
X sjsonnet.EvaluatorTests.members 1ms
X sjsonnet.EvaluatorTests.topLevel 0ms
X sjsonnet.EvaluatorTests.lazy 1ms
X sjsonnet.EvaluatorTests.unboundParam 1ms
```

The minority seems to be because there is now order for the checks, so tests that were asserting against wrongly bound arguments, but passing too many args to the function call, fail with the wrong error stack trace. But the majority seems to be a valid case where the number of arguments does not match - I can't really tell why. For example:

```
test("functions") {
      eval("(function(x) x)(1)") ==> ujson.Num(1)
      eval("function() 1") ==> ujson.Num(1)
      eval("function(a=1) a") ==> ujson.Num(1)
      eval("(function(x, y = x + 1) y)(x = 10)") ==> ujson.Num(11)
}
```

I'm not sure if this is expected or not

iuliand-db: Could it be that my suggestion was too naive? Maybe the number of arguments at the call site should be tested against the number of parameters without a default value? Not sure if the implementation follows the specification (with a desugaring step that rewrites all functions to have a default value). Anyway, if it's getting too hairy let's leave it as it was

edsilfer: Following up offline, later I can place our conclusions here for the sake of documenting it

iuliand-db (approved):

lihaoyi-databricks (approved):

Issue #99, opened 2020-11-06
Duplicated fields are allowed when building object via list comprehension


edsilfer:
## What is the issue?
When assembling an object via list comprehension, duplicated files are allowed when they should not - see the same example running on [sjsonnet.org](https://sjsonnet.org)

## How to reproduce?
1. `echo '{ ["bar"]: x for x in  [37, -987]}' > foo.jsonnet`
2. `./mill "sjsonnet[2.13.3]".jvm.run foo.jsonnet`
3. Expected:
```
RUNTIME ERROR: duplicate field name: "bar"
	example1.jsonnet:1:1-35	
```
4. Actual
```
{
   "bar": -987
}
```

PR #100, opened 2020-11-09
Do not allow duplicate field in object when evaluating list list comprehension

edsilfer:
## What changes are proposed in this pull request?
This PR fixes [issue 99](https://github.com/databricks/sjsonnet/issues/99) by issuing an error in case an `Expr.Str` is matching with an `Expr.ForSpec` with a value greater than 1. This condition denotes that the `Expr.Str` field value would be overridden in each interaction of the `Expr.ForSpec`

## How is this tested?
- Two new unit test with a positive and negative scenario emulating the issue
- Run `./mill "sjsonnet[2.13.3].jvm.test.testLocal"` to make sure that all the tests are still passing

### DIFF
--- a/sjsonnet/src/sjsonnet/Parser.scala
+++ b/sjsonnet/src/sjsonnet/Parser.scala
@@ -291,6 +291,18 @@ object Parser{
         exprs(preLocals.length)
       val postLocals = exprs.drop(preLocals.length+1).takeWhile(_.isInstanceOf[Expr.Member.BindStmt])
         .map(_.asInstanceOf[Expr.Member.BindStmt])
+      
+      /* 
+       * Prevent duplicate fields in list comprehension. See: https://github.com/databricks/sjsonnet/issues/99
+       * 
+       * If comps._1 is a forspec with value greater than one lhs cannot be a Expr.Str
+       * Otherwise the field value will be overriden by the multiple iterations of forspec
+       */
+      (lhs, comps) match {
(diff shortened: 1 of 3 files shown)


sjsonnet/src/sjsonnet/Parser.scala
```diff
@@ -291,6 +291,20 @@ object Parser{
...
+       * 
+       * If comps._1 is a forspec with value greater than one lhs cannot be a Expr.Str
+       * Otherwise the field value will be overriden by the multiple iterations of forspec
+       */
+      if (lhs.isInstanceOf[Expr.Str] &&
+          comps._1.isInstanceOf[Expr.ForSpec] && 
+          comps._1.asInstanceOf[Expr.ForSpec].cond.isInstanceOf[Expr.Arr] &&
+          comps._1.asInstanceOf[Expr.ForSpec].cond.asInstanceOf[Expr.Arr].value.length > 1
+      ) {
+        Fail.opaque(s"""no duplicate field: "${lhs.asInstanceOf[Expr.Str].value}" """)
```
lihaoyi-databricks, line 306: This probably can be more nicely written as:

```scala
(lhs, comps) match {
  case (Expr.Str(_, _), (_, Expr.ForSpec(_, _, Expr.Arr(_, values)) ) if values.length > 1 =>  Fail.opaque(...)
  case _ => // do nothing
}
```

lihaoyi-databricks (approved):
One small nit, but otherwise looks good

sjsonnet/src/sjsonnet/Parser.scala
edsilfer: Good call, updated

PR #104, opened 2020-11-25
Add objectValues and objectValuesAll to stdlib

### LINKED ISSUE #103: Add support for std.objectValues and std.objectValuesAll
This was added recently to jsonnet here: https://github.com/google/jsonnet/pull/858.

I could probably take a crack at this at some point.

DWvanGeest:
This fixes #103. I believe my implementation takes the same approach (in Scala) as the jsonnet implementation in https://github.com/google/jsonnet/pull/858.

Tested by running `mill __.test.test`, everything passes.

### DIFF
--- a/sjsonnet/src/sjsonnet/Std.scala
+++ b/sjsonnet/src/sjsonnet/Std.scala
@@ -55,26 +55,20 @@ object Std {
       v1.getVisibleKeys().get(v2).isDefined
     },
     builtin("objectFields", "o"){ (ev, fs, v1: Val.Obj) =>
-      val keys = v1.getVisibleKeys()
-        .collect{case (k, false) => k}
-        .toSeq
-      val maybeSorted = if(ev.preserveOrder) {
-        keys
-      } else {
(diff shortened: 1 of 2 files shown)


sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -1019,4 +1013,34 @@ object Std {
...
+    if(ev.preserveOrder) {
+      keys
+    } else {
+      keys.sorted
+    }
+  }
+  
+  def getObjValuesFromKeys(ev: EvalScope, fs: FileScope, v1: Val.Obj, keys: Seq[String]): Val.Arr = {
+    Val.Arr(keys.map { k =>
+      Val.Lazy(v1.value(k, -1)(fs, ev))
```
DWvanGeest, line 1043: I was a bit unsure about the `offset` argument here, but saw `-1` used elsewhere. Let me know if it should be different.

lihaoyi-databricks (approved):

PR #119, opened 2021-04-12
Some additional performance optimization

szeiger:
The latest result from 0.4.0 was:

```
[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  751.757 ± 6.697  ms/op
```

This PR brings it down to:

```
[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  310.339 ± 1.590  ms/op
```

The main optimizations are:
- New scope handling: `ValScope` is now symbol-based instead of name-based. New bindings are appended at the end with no shadowing; `self`, `$` and `super` are treated like regular variables in the scope.
- A new `StaticOptimizer` which uses a single AST transformation after parsing to implement several of the following optimizations (per source file; stored in the parse cache for reuse)
- Static binding of standard library calls
- Arity-specific function calls for standard library functions (`ApplyBuiltin`) and user-defined functions (`Apply`)
- Static resolution of named arguments and defaults in function calls
- Remove several unnecessary object allocations and indirections (`Applyer`, lambdas in members & standard library functions)
- No more `ValScope` allocation for standard library calls
- Partial application and peephole optimizations of standard library calls via `Builtin.specialize` (e.g. pre-compiled patterns in `format` and `strReplaceAll`; specialized implementation of `length(filter(...))`)
- Static application of functions when all parameters are literals
- Inlining of literals
- General optimizations of various standard library functions
- Make `Val` a subclass of `Lazy` to avoid unnecessary wrappers when a `Lazy` is required for an already computed `Val`
- Allow arrays of literals to be treated as literals (similar to static objects introduced in the last round of optimizations)
- Faster dispatch in the `Evaluator` methods by using a `tableswitch`-based dispatch for operators and a `lookupswitch` ordered by frequency (in our benchmarks) for node types
- Avoid unnecessary multiple loading of imported files
- New `Renderer` implementations based on the latest ujson
- New JSON parsing directly to Sjsonnet `Val` without an intermediate usjon AST

New features:
- `ExprTransform` and `ScopedExprTransform` for implementing tree transforms (used by the optimizer and some benchmarks)
- Benchmarks for parser, optimizer and materializer
- A profiler for gathering jfr-like profiling data at the level of AST evaluation


sjsonnet/src/sjsonnet/Evaluator.scala
```diff
@@ -314,73 +337,120 @@ class Evaluator(parseCache: collection.mutable.HashMap[(Path, String), fastparse
       )
   }
 
-  def visitBinaryOp(pos: Position, lhs: Expr, op: BinaryOp.Op, rhs: Expr)(implicit scope: ValScope) = {
-    (visitExpr(lhs), op, visitExpr(rhs)) match {
+  def visitBinaryOp(pos: Position, lhs: Expr, op: Int, rhs: Expr)(implicit scope: ValScope) = {
+    val l = visitExpr(lhs)
+    val r = visitExpr(rhs)
+    def fail() = Error.fail(s"Unknown binary operation: ${l.prettyName} ${Expr.BinaryOp.name(op)} ${r.prettyName}", pos)
+    op match {
```
lihaoyi-databricks, line 344: Should this be annotated `: @switch` to ensure the `tableswitch`-based compilation isn't accidentally broken?

sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -16,14 +16,27 @@ import scala.reflect.ClassTag
...
   def prettyName: String
+
   def cast[T: ClassTag: PrettyNamed] =
     if (implicitly[ClassTag[T]].runtimeClass.isInstance(this)) this.asInstanceOf[T]
     else throw new Error.Delegate(
       "Expected " + implicitly[PrettyNamed[T]].s + ", found " + prettyName
     )
-  def pos: Position
+
+  private[this] def failAs(err: String): Nothing =
```
lihaoyi-databricks, line 28: Should we make this take an implicit `T: PrettyNamed` for the error message, rather than passing in the `err` string manually? That would help ensure that we're consistent with our names for the various `Val.*` classes

sjsonnet/test/src/sjsonnet/ParserTests.scala
```diff
@@ -13,13 +13,13 @@ object ParserTests extends TestSuite{
...
       parse("1 * 2 + 3") ==>
-        BinaryOp(pos(6), BinaryOp(pos(2), Num(pos(0), 1), BinaryOp.`*`, Num(pos(4), 2)), BinaryOp.`+`, Num(pos(8), 3))
+        BinaryOp(pos(6), BinaryOp(pos(2), Num(pos(0), 1), BinaryOp.OP_*, Num(pos(4), 2)), BinaryOp.OP_+, Num(pos(8), 3))
 
       parse("1 + 2 * 3") ==>
-        BinaryOp(pos(2), Num(pos(0), 1), BinaryOp.`+`, BinaryOp(pos(6), Num(pos(4), 2), BinaryOp.`*`, Num(pos(8), 3)))
+        BinaryOp(pos(2), Num(pos(0), 1), BinaryOp.OP_+, BinaryOp(pos(6), Num(pos(4), 2), BinaryOp.OP_*, Num(pos(8), 3)))
     }
     test("duplicateFields") {
       parseErr("{ a: 1, a: 2 }") ==> """Expected no duplicate field: a:1:14, found "}""""
```
lihaoyi-databricks, line 25: Can we add a few tests to `ParserTests` to validate the construction of static `Val.Arr`s and `Val.Obj`s? A lot of the new changes are semantically indistinguishable from the status quo, and thus wouldn't get validated through the normal course of compiling jsonnet. We're also starting to have edge cases that are worth validating in tests, e.g. nested static arrays, nested static objects, alternating nested static arrays and static objects

sjsonnet/src/sjsonnet/Evaluator.scala
```diff
@@ -33,20 +33,24 @@ class Evaluator(parseCache: collection.mutable.HashMap[(Path, String), fastparse
   def visitExpr(expr: Expr)
                (implicit scope: ValScope): Val = try {
     expr match {
+      case Id(pos, value) => visitId(pos, value)
```
lihaoyi-databricks, line 36: I assume you re-ordered these to try and speed up lookup for the most common cases. How did you find the order? e.g. was it just a guesstimate, did you instrument it to see which ones are most common, or something else?

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -26,71 +27,416 @@ object Std {
...
+  val functions: Seq[(String, Val.Func)] = Seq(
+    "assertEqual" -> AssertEqual,
+    "toString" -> ToString,
+    "codepoint" -> Codepoint,
+    "length" -> Length,
+    "objectHas" -> ObjectHas,
+    "objectHasAll" -> ObjectHasAll,
+    "objectFields" -> ObjectFields,
+    "objectFieldsAll" -> ObjectFieldsAll,
+    builtin("objectValues", "o"){ (pos, ev, v1: Val.Obj) =>
```
lihaoyi-databricks, line 429: What's the distinction between the builtins you moved to `object`s, and those you left as-is using the `builtin` helper? Were the ones you rewrote the "hot" builtins, or was there some other distinguishing factor?

bench/src/main/scala/sjsonnet/ParserBenchmark.scala
```diff
@@ -0,0 +1,54 @@
...
+    this.interp = new Interpreter(
+      parseCache,
+      Map.empty[String, ujson.Value],
+      Map.empty[String, ujson.Value],
+      OsPath(wd),
+      importer = SjsonnetMain.resolveImport(config.jpaths.map(os.Path(_, wd)).map(OsPath(_)), None),
+      preserveOrder = config.preserveOrder.value,
+      strict = config.strict.value,
+      storePos = if (config.yamlDebug.value) currentPos = _ else _ => ()
+    )
```
lihaoyi-databricks, line 40: This part of the setup is pretty boilerplatey, could we extract it into a helper function?

sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -40,16 +53,22 @@ object Val{
...
-    private[this] var cached: Val = null
+    protected[this] var cached: Val = null
     def compute(): Val
     final def force: Val = {
       if(cached == null) cached = compute()
       cached
     }
   }
+  final class Strict(v: Val) extends Lazy {
+    this.cached =v
```
lihaoyi-databricks, line 64: whitespace after the `=`

sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -354,6 +455,124 @@ object Val{
...
+              aname,
+              Error.fail(s"Function has no parameter $aname", outerPos)
+            ) else params.indices(i)
+            i += 1
+          }
+        } catch { case e: IndexOutOfBoundsException =>
+          Error.fail("Too many args, function has " + params.names.length + " parameter(s)", outerPos)
+        }
+        arrI
+      } else if(params.indices.length < argsSize) {
```
lihaoyi-databricks, line 496: Would moving this case to the top of the if-else chain allow us to avoid the `try`-`catch` above? That would also allow us to avoid duplicating the `Error.fail` call.

lihaoyi-databricks:
Looks good, left some comments

sjsonnet/src/sjsonnet/Evaluator.scala
szeiger: AFAIR `@switch` is broken. But even if it works, it also accepts a `lookupswitch` so it's not much of help.

sjsonnet/src/sjsonnet/Evaluator.scala
szeiger: I gathered statistics from our universe benchmark. It may not be representative for all use cases but the current version doesn't seem to be ordered for performance anyway.

sjsonnet/src/sjsonnet/Std.scala
szeiger: Yes, I mainly rewrote the ones that are used in the benchmark. It's a lot of work...

sjsonnet/src/sjsonnet/Val.scala
szeiger: I already tested all of these variants and they are all much slower.

The trick with setting `this.cached` lies in avoiding any megamorphic call when the `Lazy` gets evaluated. `force` is monomorphic, `compute` is megamorphic.

szeiger:
Some of those might already be obsolete with the further changes I made. Current benchmark time is 483ms. I'm still exploring a few options for improving it.

sjsonnet/src/sjsonnet/Std.scala
lihaoyi-databricks: Got it, leaving them half-half is probably fine for now then

sjsonnet/src/sjsonnet/Val.scala
lihaoyi-databricks: Got it

lihaoyi-databricks: @szeiger although on further thought: couldn't we keep `.force` final, while still transitioning to the single type hierarchy to implement `.compute`? That would allow us to keep `force` monomorphic, keep `.compute` just as megamorphic as it is now, while still avoiding the overhead of boxing things in `Strict`

szeiger: It's also slower (although I don't fully understand why in this case)

lihaoyi-databricks: Got it. Hotspot moves in mysterious ways

szeiger: I tired it again and it actually works now. The last attempt (where I already used the trick of assigning the cached value in the constructor) was only a week ago. It's not a clear performance win but at least it's not slower and it allocates fewer objects.

szeiger:
Updated with the latest changes. I'm running out of ideas and haven't made progress in a while. We should get this version merged. I tested it against universe. It will require 2 changes there because we relied on incorrect behavior of the old Sjsonnet release.

szeiger:
0.4.0 stands at 751 ms in our benchmark. Here's the progress over the course of this PR:

```
Remove Applyer & optimize function application:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  738.439 ± 6.316  ms/op

Scope-free function application:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  690.672 ± 4.851  ms/op

Optimize Std:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  673.334 ± 5.585  ms/op

Optimize builtinWithDefaults:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  663.788 ± 2.486  ms/op

Optimize Builtin args handling:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  642.562 ± 3.537  ms/op

Shared Val.Strict + Array literals:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  626.566 ± 3.125  ms/op

Static object optimization + Obj.Member without lambdas:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  616.049 ± 4.625  ms/op

Reorder operations by frequency:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  595.176 ± 3.789  ms/op

tableswitch-based operator lookup:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  582.002 ± 5.800  ms/op

Static bindings of Std calls:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  501.764 ± 3.263  ms/op

More static optimization:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  494.850 ± 2.327  ms/op

Fix import caching:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  483.840 ± 3.675  ms/op

Conventional scope handling:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  480.387 ± 2.640  ms/op

Optimize scope handling:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  469.976 ± 2.968  ms/op

Val extends Lazy:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  464.764 ± 1.853  ms/op

Optimize Materializer + Renderer:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  431.154 ± 1.928  ms/op

ApplyN:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  408.612 ± 3.715  ms/op

Add special-case Exprs

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  401.357 ± 1.430  ms/op

Stdlib specialization:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  391.974 ± 2.038  ms/op

% specialization:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  378.203 ± 2.186  ms/op

Static apply + length optimization:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  363.323 ± 2.544  ms/op

MaterializeRenderer:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  337.980 ± 1.187  ms/op

Filter strictness optimization:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  332.285 ± 1.596  ms/op

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  331.977 ± 1.212  ms/op

Builtin with defaults & std.setInter:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  322.575 ± 0.885  ms/op

length(filter) specialization:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  314.919 ± 1.103  ms/op

Reduce phase mismatches in scoped transform:

[info] Benchmark           Mode  Cnt    Score   Error  Units
[info] MainBenchmark.main  avgt  160  310.339 ± 1.590  ms/op
```

build.sbt
```diff
@@ -6,7 +6,7 @@ cancelable in Global := true
 
 lazy val main = (project in file("sjsonnet"))
   .settings(
-    scalacOptions in Compile ++= Seq("-opt:l:inline", "-opt-inline-from:sjsonnet.**"),
+    scalacOptions in Compile ++= Seq("-opt:l:inline", "-opt-inline-from:sjsonnet.*,sjsonnet.**"),
```
szeiger, line 9: The `-opt-inline-from` syntax is confusing. `sjsonnet.**` only inlines from subpackages (which we don't have). We need to add `sjsonnet.*` to inline from the `sjsonnet` package, too. Ultimately it doesn't make a big difference. HotSpot has become so good that it doesn't matter in most cases but the optimizations are not entirely deterministic. Sometimes you end up with a benchmark run that is 10% slower than it should be. Letting scalac do the trivial inlining (and thus creating less work for HotSpot) makes this more reliable.

lihaoyi-databricks:
Looks good, few things before we merge it:

- Flesh out each of the commits with a paragraph (2-3 sentence) description of what you are doing and why? That will help make the blame useful for when we merge/rebase it  without squashing
- Update the readme.md `Architecture` section: we have a new phase (static optimizations) and the data types are different (now `Val` and `Expr` are related, and `Lazy` and `Strict` now exist)

lihaoyi-databricks (approved):

szeiger:
Updated with additional docs.

PR #125, opened 2021-07-12
Error handling improvements

### LINKED ISSUE #122: Unknown variables should be rejected statically
This static verification is required by the Jsonnet specification, but Sjsonnet does not verify references until they are evaluated. Both of these expressions are accepted by Sjsonnet because the illegal references are never evaluated:

```
$ echo "local x = y; 1" | jsonnet -
STATIC ERROR: <stdin>:1:11: Unknown variable: y

$ echo "local x = self.y; 1" | jsonnet -
STATIC ERROR: <stdin>:1:11-15: Can't use self outside of an object.
```

szeiger:
- More consistent error handling:
  - Remove Error.Delegate
  - Skip JVM stack trace handling
  - Consistent formatting of raw delegate errors and regular positioned errors
- Unify error stack trace handling:
  Instead of sprinkling error propagation with or without stack frames all over the Evaluator and other parts of the codebase, we now do it in one place (visitExpr) with additional logic in Error.withStackFrame. Duplicate frames are skipped, an error without any stack frame always gets the current position, and all other frames are added or skipped depending on the Expr at the position.
- Add expression types to stack traces:
  This makes stack traces much more descriptive. We also change some of the reported positions / expr types to better match the semantics of the errors.
- Consistent treatment of parse errors:
  They are reported as sjsonnet.ParseError with normal stack traces instead of the previous one-liners which required customized error messages for parse errors in imported files. Errors are now stored as Error objects instead of Strings in the parse cache. This paves the way for error reporting during static evaluation and it is already required now to translate the paths in parse error stack traces to a different base directory (when reusing a cached parse error from a previous run).
- Static errors for unknown variables:
  Unknown variables and illegal self/super/$ references are now checked statically as they are supposed to be. Error messages are consistent with Google Jsonnet's.

Fixes #122

### DIFF
--- a/sjsonnet/src/sjsonnet/Error.scala
+++ b/sjsonnet/src/sjsonnet/Error.scala
@@ -16,33 +16,21 @@ class Error(val msg: String, stack: List[StackTraceElement] = Nil, underlying: O
 
   override def fillInStackTrace: Throwable = this
 
-  def addFrame(pos: Position, wd: Path, expr: Expr = null)(implicit ev: EvalErrorScope): Error = {
+  def addFrame(pos: Position, expr: Expr = null)(implicit ev: EvalErrorScope): Error = {
     if(stack.isEmpty || alwaysAddPos(expr)) {
-      val cl = ""
-  //    val cl = expr match {
-  //      case null => ""
-  //      case expr => expr.getClass.getName.replaceFirst("^sjsonnet.Expr\\$(.*)", "[$1]")
-  //    }
-      val newFrame = ev.prettyIndex(pos) match {
-        case None =>
-          new StackTraceElement(
-            cl, "",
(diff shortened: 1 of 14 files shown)


sjsonnet/src/sjsonnet/Error.scala
```diff
@@ -9,61 +9,90 @@ import scala.util.control.NonFatal
...
+        case s :: ss if s.pos == pos =>
+          if(s.exprErrorString == null && exprErrorString != null) copy(stack = newFrame :: ss)
+          else this
+        case _ => copy(stack = newFrame :: stack)
+      }
+    } else this
   }
-}
 
+  def asSeenFrom(ev: EvalErrorScope): Error =
```
iuliand-db, line 32: ❤️

iuliand-db (approved):

PR #128, opened 2021-09-01
[Plat 33341]

tanmay-db:


sjsonnet/src-jvm-native/sjsonnet/SjsonnetMain.scala
```diff
@@ -7,8 +7,24 @@ import java.nio.file.NoSuchFileException
 import scala.util.Try
 import scala.util.control.NonFatal
 
+// JsonnetWorker (in universe) extends this trait in the universe so that it can pass the cache based on Caffeine to main0 here
+trait ParseCacheInterface {
```
iuliand-db, line 11: Let's call this `ParseCache` and leave `Interface` out of the name, it's simpler and there's no confusion around what it is.

sjsonnet/src-jvm-native/sjsonnet/SjsonnetMain.scala
```diff
@@ -7,8 +7,24 @@ import java.nio.file.NoSuchFileException
...
+  def getOrElseUpdate(key: (Path, String), defaultValue:  => Either[Error, (Expr, FileScope)]): Either[Error, (Expr, FileScope)]
+}
+
+// As JsonnetWorker (in universe) passes an object to main0 which extends the trait: ParseCacheInterface, class based on HashMap
+// which extends this trait is needed to use the default HashMap
+case class HashMapDefault () extends ParseCacheInterface {
```
iuliand-db, line 17: The comment is a bit too detailed. What you could say is something like `A default implementation based on a mutable HashMap. This implementation is not thread-safe.`

I'd make this a simple `class` (not a `case class`, because it's not "data" and we don't pattern match on it). The name could be `DefaultParseCache` or something like that. The name should reflect what the class does (it's a parse cache), not only how it's implemented.

sjsonnet/src-jvm-native/sjsonnet/SjsonnetMain.scala
```diff
@@ -7,8 +7,24 @@ import java.nio.file.NoSuchFileException
...
+  val cache = new collection.mutable.HashMap[(Path, String), Either[Error, (Expr, FileScope)]]()
+
+  override def getOrElseUpdate(key: (Path, String), defaultValue:  => Either[Error, (Expr, FileScope)]): Either[Error, (Expr, FileScope)] = {
+    cache.getOrElseUpdate((key._1, key._2), defaultValue)
```
szeiger, line 21: You don't want to allocate a new tuple for `(key._1, key._2)` here, just use `key` directly.

sjsonnet/src-jvm-native/sjsonnet/SjsonnetMain.scala
szeiger: Yes, let's make this `ParseCache` and `DefaultParseCache`. Adding `Interface` to a name is an old Java practice that never caught on in Scala.

sjsonnet/src-jvm-native/sjsonnet/SjsonnetMain.scala
```diff
@@ -7,8 +7,24 @@ import java.nio.file.NoSuchFileException
...
+  }
+}
+
 object SjsonnetMain {
-  def createParseCache() = collection.mutable.HashMap[(Path, String), Either[Error, (Expr, FileScope)]]()
+//  def createParseCache() = collection.mutable.HashMap[(Path, String), Either[Error, (Expr, FileScope)]]()
+  def createParseCache() = new HashMapDefault
```
szeiger, line 27: The presence or absence of an argument list at the call site should generally match the definition (and this is often required in Scala). In this case it's better to change the definition (`class DefaultParseCache` instead of `class DefaultParseCache()`). Since it's a constructor, you already know that it always allocates a new object, and there are no further side-effects.

sjsonnet/src-jvm-native/sjsonnet/SjsonnetMain.scala
tanmay-db: Thank you for the review iulian and Stefan, I have updated them.

sjsonnet/src/sjsonnet/Importer.scala
```diff
@@ -43,8 +43,8 @@ class CachedImporter(parent: Importer) extends Importer {
 
 class CachedResolver(
   parentImporter: Importer,
-  val parseCache: mutable.HashMap[(Path, String), Either[Error, (Expr, FileScope)]] = new mutable.HashMap
-) extends CachedImporter(parentImporter) {
+  val parseCache: ParseCache = new DefaultParseCache
```
iuliand-db, line 46: Let's remove this default value so we can make sure we pass the right parse cache from all code paths.

This means you'll have to find all places where this method is called and decide if it should pass a fresh `DefaultParseCache` (if it's on a code path that's NOT coming from Bazel, meaning mostly tests) or make sure we have an external parse cache.

sjsonnet/src/sjsonnet/Interpreter.scala
```diff
@@ -18,8 +18,8 @@ class Interpreter(extVars: Map[String, ujson.Value],
                   preserveOrder: Boolean = false,
                   strict: Boolean = false,
                   storePos: Position => Unit = null,
-                  val parseCache: mutable.HashMap[(Path, String), Either[Error, (Expr, FileScope)]] = new mutable.HashMap,
-                  ) { self =>
+                  val parseCache: ParseCache = new DefaultParseCache
```
iuliand-db, line 21: Same here, let's not have a default argument for the parse cache

build.sc
```diff
@@ -1,5 +1,5 @@
 import mill._, scalalib._, publish._, scalajslib._, scalanativelib._, scalanativelib.api._
-val sjsonnetVersion = "0.4.1"
+val sjsonnetVersion = "0.4.1-SNAPSHOT"
```
szeiger, line 2: Let's leave this at 0.4.1, we're going to cut a 0.4.1 release from it anyway. (We could update to 0.4.2-SNAPSHOT afterwards, but it's probably not worth the effort)

sjsonnet/server/src/sjsonnet/SjsonnetServerMain.scala
```diff
@@ -24,7 +24,11 @@ trait SjsonnetServerMain[T]{
             wd: os.Path): (Boolean, Option[T])
 }
 
-object SjsonnetServerMain extends SjsonnetServerMain[collection.mutable.HashMap[(Path, String), Either[String, (Expr, FileScope)]]]{
+//  [Ask] SjsonnetServerMain has a valid main method (args0: Array[String]): Unit,
+//  [warn]   but sjsonnet.SjsonnetServerMain will not have an entry point on the JVM.
+//  [warn]   Reason: companion is a trait, which means no static forwarder can be generated.
+//  [warn] object SjsonnetServerMain extends SjsonnetServerMain[DefaultParseCache]{
```
szeiger, line 30: What's with this comment? Is this an issue that should be addressed?

sjsonnet/src-js/sjsonnet/SjsonnetMain.scala
```diff
@@ -6,7 +6,7 @@ import scala.scalajs.js.annotation.{JSExport, JSExportTopLevel}
 
 @JSExportTopLevel("SjsonnetMain")
 object SjsonnetMain {
-  def createParseCache() = collection.mutable.HashMap[(Path, String), Either[String, (Expr, FileScope)]]()
+  def createParseCache() = new DefaultParseCache
```
szeiger, line 9: Can we deprecate this method? We have the default encapsulated nicely in `new DefaultParseCache` now and it is used directly in several places, so this method seems superfluous.

sjsonnet/src-jvm-native/sjsonnet/SjsonnetMain.scala
```diff
@@ -8,7 +8,7 @@ import scala.util.Try
 import scala.util.control.NonFatal
 
 object SjsonnetMain {
-  def createParseCache() = collection.mutable.HashMap[(Path, String), Either[Error, (Expr, FileScope)]]()
+  def createParseCache() = new DefaultParseCache
```
szeiger, line 11: same here

sjsonnet/server/src/sjsonnet/SjsonnetServerMain.scala
tanmay-db: This got generated by I ran `./mill __.test`, I am not sure why this came but after some changes this doesn't come however the comment already present - "Disable SIGINT interrupt signal in the Mill server." got more detailed.

build.sc
tanmay-db: Okay updated

sjsonnet/src-js/sjsonnet/SjsonnetMain.scala
tanmay-db: Yes, that makes sense. Updated

PR #132, opened 2021-10-05
Add support for warnings and --no-static-errors

szeiger:
Warnings could prove useful in the long run if we want to add linting. `--no-static-errors` is provided for migrating from the previous version but should be removed again in the long term because the specification requires static error handling. Since warnings are not thrown they do not get a full stack trace. Only the actual position of the warning is reported.

### DIFF
--- a/sjsonnet/test/src-jvm-native/sjsonnet/ErrorTests.scala
+++ b/sjsonnet/test/src-jvm-native/sjsonnet/ErrorTests.scala
@@ -12,7 +12,7 @@ object ErrorTests extends TestSuite{
       OsPath(os.pwd),
       importer = sjsonnet.SjsonnetMain.resolveImport(Array.empty[Path]),
       warnLogger = (msg: String) => out.append(msg).append('\n'),
-      noStaticErrors = noStaticErrors
+      settings = new Settings(noStaticErrors = noStaticErrors),
     )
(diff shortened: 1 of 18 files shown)


iuliand-db (approved):

sjsonnet/test/src-jvm-native/sjsonnet/ErrorTests.scala
```diff
@@ -170,12 +184,24 @@ object ErrorTests extends TestSuite{
...
+        """[warning] sjsonnet.StaticError: Unknown variable: x
+          |    at [Id x].(sjsonnet/test/resources/test_suite/lib/static_check_failure.jsonnet:2:1)
+          |
+          |sjsonnet.Error: Unknown variable: x
+          |    at [Id x].(sjsonnet/test/resources/test_suite/lib/static_check_failure.jsonnet:2:1)
+          |    at [Import].(sjsonnet/test/resources/test_suite/error.import_static-check-failure.jsonnet:1:1)
+          |""".stripMargin,
+        noStaticErrors = true
+      )
+    }
```
tanmay-db, line 204: Hi Stefan, thanks for the PR. Just a question, I didn't understood why the two checks -- the original one with ` """sjsonnet.StaticError: Unknown variable: x` and new one with warning `"""[warning] sjsonnet.StaticError: Unknown variable: x` would be needed. Can you please tell about this?

PR #135, opened 2021-10-30
implements support for safe select operator ?.

jam01:


jam01:
sorry wrong repository!

PR #136, opened 2021-11-03
Make lazy initialization of static Val.Obj thread-safe

szeiger:
Static Val.Obj instances are created by the optimizer and will end up in the parse cache. If the cache is shared by multiple threads, initialization must be sufficiently safe, i.e. computing a value multiple times in race conditions is allowed (and cheaper than ensuring that it doesn't happen), but object must never get into an invalid intermediate state.

### DIFF
--- a/sjsonnet/src/sjsonnet/Val.scala
+++ b/sjsonnet/src/sjsonnet/Val.scala
@@ -155,10 +155,12 @@ object Val{
 
     private[this] def getValue0: util.LinkedHashMap[String, Obj.Member] = {
       if(value0 == null) {
-        value0 = new java.util.LinkedHashMap[String, Val.Obj.Member]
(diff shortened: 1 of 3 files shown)


sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -155,10 +155,11 @@ object Val{
 
     private[this] def getValue0: util.LinkedHashMap[String, Obj.Member] = {
       if(value0 == null) {
-        value0 = new java.util.LinkedHashMap[String, Val.Obj.Member]
```
iuliand-db, line 158: Could you add a comment explaining why we do things in this way, and the multi-threaded guarantee we want to maintain?

apattidb (approved):

PR #143, opened 2022-01-21
Implement quote_keys flag for std functions manifestYamlDoc/manifestY…

jnyi:
…amlStream to match official jsonnet doc

### Purpose
This PR creates a flag documented in jsonnet official site std library: https://jsonnet.org/ref/stdlib.html for function
`manifestYamlDoc` and `manifestYamlStream`, it will allow us to generate yaml without quoting the keys.

### Test
* Tested via unit tests.
* compile sjsonnet using `./mill -i show "sjsonnet[2.13.4]".jvm.assembly` and run again a locally created test.jsonnet to output YAML via flag `--yaml-out`

### DIFF
--- a/sjsonnet/src/sjsonnet/YamlRenderer.scala
+++ b/sjsonnet/src/sjsonnet/YamlRenderer.scala
@@ -8,7 +8,7 @@ import upickle.core.{ArrVisitor, ObjVisitor}
 
 
 class YamlRenderer(_out: StringWriter = new java.io.StringWriter(), indentArrayInObject: Boolean = false,
-                   indent: Int = 2) extends BaseCharRenderer(_out, indent){
+                   quoteKeys: Boolean = true, indent: Int = 2) extends BaseCharRenderer(_out, indent){
   var newlineBuffered = false
   var dashBuffered = false
   var afterKey = false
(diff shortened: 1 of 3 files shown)


lihaoyi-databricks:
Passing the review to @szeiger

sjsonnet/src/sjsonnet/YamlRenderer.scala
```diff
@@ -30,6 +30,15 @@ class YamlRenderer(_out: StringWriter = new java.io.StringWriter(), indentArrayI
     }
   }
 
+  private[this] def removeQuoteKey(): Unit = {
+    // refer to quote_keys parameter in https://jsonnet.org/ref/stdlib.html, only unquote keys
+    if (!quoteKeys && !afterKey) {
```
szeiger, line 35: This looks wrong. quote_keys=false is supposed to omit the quotes *if possible*, so there should be a check that determines which strings are allowed to be unquoted.

Patching the buffer after writing is a bit awkward. Can we also move this code into writeString so we can determine quoting *first* (according to the correct criteria, not just the flag) and then emit a quoted or non-quoted string? The new code for non-quoted writing should only end up in a single path because the other ones won't qualify for quote removal anyway.

jnyi: I've tested the behavior of jsonnet's quote_keys=false which removes quotes of the yaml keys, so my understanding `!afterKeys` indicates we are dealing with key right now, as for patching the buffer, agree it is a bit counter-intuitive, let me see if I can do something to writeString, found it is a bit harder to modify that function interfaces which touches more places.

szeiger: It should only have to touch the last branch in `writeString`. There other 2 should not be applicable for quote removal.

stephenamar-db:
I cloned your PR and finished the work in https://github.com/databricks/sjsonnet/pull/231

PR #147, opened 2022-04-25
*New* Add UUID Function

agrawroh:
## What changes are proposed in this PR?

In this PR we are adding a new standard function to return a random UUID string. This could be used to tag the different k8s resources if they are all being deployed as a list (kind: List). Kubernetes creates a separate UUID for each individual resource even if they are all getting deployed together and we have a use-case where we want to identify everything getting deployed as part of the same list.

### DIFF
--- a/sjsonnet/src-js/sjsonnet/Platform.scala
+++ b/sjsonnet/src-js/sjsonnet/Platform.scala
@@ -12,6 +12,9 @@ object Platform {
   def xzString(s: String): String = {
     throw new Exception("XZ not implemented in Scala.js")
   }
+  def uuid(length: Int): String = {
+    throw new Exception("UUID not implemented in Scala.js")
+  }
   def md5(s: String): String = {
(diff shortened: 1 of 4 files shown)


szeiger:
Jsonnet is a pure language. We cannot add impure functions that break referential transparency. I found related discussions in https://github.com/google/jsonnet/issues/225 and https://github.com/google/jsonnet/issues/478. The second one mentions k8s descriptors, this may be close to your use case. I think the approach from the first ticket should work for us as well: Compute a hash of the content (or maybe a list index when iterating over a list of elements) based on a seed that is passed in as an extVar. We already have hashing functions in the standard library. You could use this to construct a deterministic UUID from a hash (plus an external seed).

Such a deterministic UUID function could be implemented in Scala but it's probably not worth the effort. If we still want to do it, it should go into a separate namespace, maybe `stdext` instead of `std` to make clear that it's not part of the official standard library, together with a flag that allows our Sjsonnet-specific extensions to be disabled.

PR #151, opened 2022-06-10
Fix parse error when `id == ...` is used as an argument

### LINKED ISSUE #127: Parsing error on valid conditionals as local function parameter
This program:

```
local myFunc = function(a)
  if (a) then "a" else "b";

local b = "aaa";

myFunc(b == [] || b == ['e'])
```

works on https://jsonnet.org/ but produces a parsing error `"Parse error: Expected ")":7:8, found "b == [] ||""` on `sjsonnet` 0.4.0.

gilles-duboscq:
Fixes #127


gilles-duboscq:
Thanks!

Issue #138, opened 2021-11-16
sjsonnet allows computed imports unlike go-jsonnet


nicklan:
Sample file:
```
local i = import "a" + import "b";
0
```

go-jsonnet:
```
$ jsonnet foo.jsonnet
STATIC ERROR: /tmp/foo.jsonnet:1:18-34: computed imports are not allowed.
```

sjsonnet:
```
$ sjsonnet foo.jsonnet
0
```

szeiger:
This looks like a precedence problem. `import` should have lower precedence than `+` but Sjsonnet treats it as higher. We see the same for `.` (`local buiApiServerReleasePipeline = import "./release-pipeline.jsonnet.TEMPLATE".buiApiServerReleasePipeline;`)

szeiger:
From https://jsonnet.org/ref/spec.html:
> Everything is left associative. In the case of assert, error, function, if, import, importstr, and local, ambiguity is resolved by consuming as many tokens as possible on the right hand side.

But the grammar contains
> import string

and this is what Sjsonnet implements. The intention is probably to parse as `import expr` and then assert that the `expr` is a `string`.

PR #153, opened 2022-09-01
Parse arbitrary exprs after "import" and "importStr"

szeiger:
Fixes https://github.com/databricks/sjsonnet/issues/138

The spec is not very clear on this but this behavior is consistent with the C++ and Go implementations.


nicklan:
Is this still something you want to merge? Also, will this catch cases like:
```
local default = import "foo.jsonnet".DEFAULT;
```

Issue #154, opened 2022-10-19
+: is erroneously interpreted as : within object comprehension


NicholasHusin:
Given the following:
```
{ ['foo']+: [x] for x in [1]} + { ['foo']+: [x] for x in [2]}
```
The official implementation of `jsonnet` would output:
```
{
  "foo": [
    1,
    2
  ]
}
```
due to the `+:` used.

However, on `sjsonnet`, the `+` in `+:` seems to be ignored, resulting in
```
{ ['foo']+: [x] for x in [1]} + { ['foo']+: [x] for x in [2]}
```
being interpreted as:
```
{ ['foo']: [x] for x in [1]} + { ['foo']: [x] for x in [2]}
```
and resulting in:
```
{
  "foo": [
    2
  ]
}
```
This causes an inconsistency between `jsonnet` and `sjsonnet`.

lihaoyi-databricks:
Should be fixed by https://github.com/databricks/sjsonnet/pull/155

PR #155, opened 2022-10-19
Fix +: being erroneously interpreted as :

NicholasHusin:
Addresses issue #154.

When using +: within an object comprehension, sjsonnet will interpret it as a : instead. This creates an inconsistency with the official implementation of jsonnet. This commit fixes said issue by making sure that + are correctly parsed and evaluated within object comprehension.


AjayMT:
LGTM

PR #156, opened 2022-11-01
[ES-449270] Fix duplicate key error in object comprehension

### LINKED ISSUE #121: Duplicate fields in object comprehensions are not detected
This should fail but instead the second binding wins:
```
{ [t[0]]: t[1] for t in [['foo',1],['foo',2]] }
```

ckolb-db:
Fixes #121 

Please advise on checking for size twice vs calling `contains` once per key, I haven't benchmarked this.

This might also break some existing files which rely on this behavior without realizing it's out of spec, either with duplicate list entries getting reduced or more complex replacements where only the last entry matters. Not sure if we're ok with that or  what other options we have. Breaking these probably beats allowing duplicate entries in the future.

### DIFF
--- a/sjsonnet/src/sjsonnet/Evaluator.scala
+++ b/sjsonnet/src/sjsonnet/Evaluator.scala
@@ -589,12 +589,16 @@ class Evaluator(resolver: CachedResolver,
 
         visitExpr(e.key)(s) match {
           case Val.Str(_, k) =>
+            val prev_length = builder.size()
             builder.put(k, new Val.Obj.Member(false, Visibility.Normal) {
               def invoke(self: Val.Obj, sup: Val.Obj, fs: FileScope, ev: EvalScope): Val =
                 visitExpr(e.value)(
                   s.extend(newBindings, self, null)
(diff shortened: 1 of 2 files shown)


iuliand-db (commented):
I wonder how do these duplicate entries happen? Since the fields are keys in a map, aren't they unique by construction? Even the error is on the map not increasing in size. I wonder if the current tests would fail without the fix, and if yes, then where do we get the two copies from in the materialized json?

ckolb-db:
It's not an object as input, but rather a list of arbitrary values/objects from which the object entry can be constructed yet again arbitrarily. The first uniqueness check in the system is the one that was missing here.

The test would fail so far as it would not throw an error and rather return an object with only the last mention of the key in it.

This surfaced again due to an issue in our `teams.jsonnet` file where a team name was duplicated in two unrelated entries.

I'm checking the map size rather than map.contains before inserting as an optimization, since contains is only armortized O(1) and contains the string length comparison.

iuliand-db (approved):

PR #160, opened 2023-04-08
fix parsing of args with = in them

carl-db:
The CLI args with 'k=v' structure fail because it will split multiple times (if there is an '=' in the value)


carl-db:
It looks like the tests don't pass in this repo at all anymore? I couldn't find what is actually failing from the test output.

szeiger:
CI failures fixed by https://github.com/databricks/sjsonnet/pull/163. This was accidentally pushed to master. The build passed. I added a branch protection rule to prevent direct pushes to master in the future.

PR #162, opened 2023-04-11
Upgrade to Scala.js 1.8.0

szeiger:
The CI image was updated and we're hitting https://github.com/scala-js/scala-js-js-envs/issues/12 on 1.4.0.


lihaoyi-databricks:
Closing for now. Since https://github.com/databricks/sjsonnet/pull/163 fixed CI, so we don't need to do this, and upgrading may be inconvenient to downstream users who also get forced to upgrade

PR #165, opened 2023-04-17
update build to use fastParse 3.0.1

ngbinh:
and other libraries


lihaoyi-databricks:
Might need to hold off on this a bit. We're still on the older versions of libraries internally, and would need to upgrade that before (or together with) upgrading sjsonnet to make sure they don't diverge

He-Pin:
@ngbinh it should be 3.1.0 now:)

PR #166, opened 2023-05-18
Make Jsonnet std lib non-global

lihaoyi-databricks:
This serves two purposes:

1. We think we're seeing race conditions where the lazily-evaluated mutable state in the std lib `Val.Obj` is causing livelocks when evaluating Jsonnet in a multithreaded environment.

2. This will allow whoever instantiates `sjsonnet.Interpreter` to customize the `std.*`, allowing people to inject their own functions. Useful if google/jsonnet has some new std lib functions that haven't been added to sjsonnet yet, or if you have some domain-specific functions you want to expose in your config

### DIFF
--- a/sjsonnet/src/sjsonnet/Interpreter.scala
+++ b/sjsonnet/src/sjsonnet/Interpreter.scala
@@ -19,11 +19,12 @@ class Interpreter(extVars: Map[String, ujson.Value],
                   settings: Settings = Settings.default,
                   storePos: Position => Unit = null,
                   warnLogger: (String => Unit) = null,
+                  std: Val.Obj = new Std().Std
                   ) { self =>
 
   val resolver = new CachedResolver(importer, parseCache) {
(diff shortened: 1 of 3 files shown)


lihaoyi-databricks:
Review by @szeiger

szeiger (approved):
I think in the longer term we should look into getting a clearer separation of official Jsonnet functions and custom extensions instead of putting non-standard functions into `std`.

lihaoyi-databricks:
Yeah this change would open up the possibility of defining custom `std` functions like `gzip` or `xz` in the databricks codebase, and other people can define their own `std` functions easily, rather than having to chuck everything into the shared `Std.scala`

szeiger:
What I meant is that we should put them into a separate object (e.g. `stdext`). I think we already have some non-standard features in `std`. This makes it hard to tell if a source is compatible with Jsonnet or if it requires non-standard Sjsonnet features.

Issue #108, opened 2020-12-15
Folding strings


isamert:
I found that sjsonnet and Google's jsonnet implementation differ in `foldr`/`foldl`. You can fold strings in Google's implementation, that is, strings are treated as `[char]` but `sjsonnet` errors out when string is supplied as an argument to `fold{r,l}`. Example:

test.jsonnet
```
std.foldr(function (acc, it) acc + it, "this is ", "jsonnet")
```

Go implementation outputs:
```
$ jsonnet test.jsonnet
"this is jsonnet"
```

sjsonnet outputs:
```
$ ./sjsonnet.jar test.jsonnet
sjsonnet.Error: Wrong parameter type: expected Array, got string
    at .(test.jsonnet:1:10)
```

Not a big problem, but Google implementation is kinda more convenient. I can just workaround this with the following of course:

```
std.foldr(function (acc, it) acc + it, "this is ", ["j", "s", "o", "n", "n", "e", "t"])
```

isamert:
Related to #73

PR #173, opened 2023-06-02
Allow std.fold{l,r} to take strings

lihaoyi-databricks:
Fixes https://github.com/databricks/sjsonnet/issues/108. This is not documented behavior, but we should try and be compatible anyway


sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -151,24 +151,47 @@ class Std {
...
+            val c = current
+            current = func.apply2(c, item, pos.noOffset)(ev)
+          }
+          current
+
+        case s: Val.Str =>
+          var current = init
+          for (char <- s.value) {
+            val c = current
+            current = func.apply2(c, Val.Str(pos, new String(Array(char))), pos.noOffset)(ev)
```
szeiger, line 167: There is `String.valueOf(char)` but from looking at the implementation it may actually be worse than `new String` (since it supports compact strings)

szeiger (approved):

Issue #112, opened 2021-03-03
Output multiple yaml documents


akizminet:
It seem like  `--yaml-out` doesn't work with `-m --multi`. It always output json files.

Issue #77, opened 2020-06-18
Inconsistency with error keyword output between sjsonnet and go-jsonnet


Qinusty:
With go-jsonnet, the error keyword triggers a failure with an error message and some form of trace.

sjsonnet however, simply errors with status code 1 and nothing sent to stdout or stderr.

### Reproducer

```bash
jsonnet -e "error 'Foo.'"
java -jar ~/Downloads/sjsonnet.jar -e "error 'Foo.'"
```

rhanqtl:
The reason is that sjsonnet does not support the `-e` or `--exec` option as go-jsonnet does. Print the help or read the source code (`sjsonnet/src-jvm/sjsonnet/Cli.scala`) to check this.

Issue #60, opened 2020-04-02
YAML Stream Output doesn't work with `--output-file`


FraBle:
When specifying `-y` for the yaml stream output as well as `--output-file`, sjsonnet writes to stdout instead of the output file.

lihaoyi-databricks:
Not sure when this was fixed, but it appears to work correctly in master

```
$ cat ../sjsonnet/sjsonnet/test/resources/db/stream.jsonnet
[
    1,
    2,
    3,
    4,
    5
]

$ bin/sjsonnet --yaml-stream --output-file output.txt ../sjsonnet/sjsonnet/test/resources/db/stream.jsonnet

$ cat output.txt
--- 1
--- 2
--- 3
--- 4
--- 5
```

PR #174, opened 2023-06-05
Fixes for main method

lihaoyi-databricks:
Bug fixes and unit tests to validate behavior of various command line flags and their combinations: `--exec`, `--yaml-out`, `--yaml-stream`, `--multi`, `--output-file`. I also propagated the `std: Std` argument all the way to `sjsonnet.SjsonnetMain.main0`, so it can easily be passed in by people who use the CLI-arg-based programmatic interface without having to drop down into `sjsonnet.Interpreter`, e.g. in Databricks' `JsonnetWorker`

Fixes https://github.com/databricks/sjsonnet/issues/112, and fixes https://github.com/databricks/sjsonnet/issues/77, and adds tests for https://github.com/databricks/sjsonnet/issues/60 (which was already working before, but without test coverage)

The logic inside `SjsonnetMain.scala` is still pretty messy, but at least it has some rudimentary unit tests now. 

Another thing to note is that trailing newline handling isn't great; depending on what code path you go through, you may get zero, one, or two trailing newlines. That should be benign, since trailing newlines are not semantically meaningful in either JSON or YAML, so for now I just left them in place. At least now we have tests to surface some of the weirdness

I had to move `MainTests.scala` from `src-jvm-native/` to `src-jvm/`, due to weird crashes in the Scala Native test-running infrastructure. Given the experimental nature of Scala-Native, losing a bit of coverage is probably OK for now, and also it's unlikely that this will cause anything to break given the same code receives test coverage on the JVM.


sjsonnet/src-jvm-native/sjsonnet/SjsonnetMain.scala
```diff
@@ -121,26 +131,34 @@ object SjsonnetMain {
...
                      wd: os.Path,
                      allowedInputs: Option[Set[os.Path]] = None,
                      importer: Option[(Path, String) => Option[os.Path]] = None,
-                     warnLogger: String => Unit = null): Either[String, String] = {
-    val path = os.Path(file, wd)
+                     warnLogger: String => Unit = null,
+                     std: Val.Obj = new Std().Std): Either[String, String] = {
+
+    val (jsonnetCode, path) =
+      if (config.exec.value) (file, wd / "<exec>") else (os.read(os.Path(file)), os.Path(file))
```
szeiger, line 160: Abusing `file` to be the source code is rather ugly.

szeiger (approved):
Can we also support `-` input on the command line now that the file reading happens later? It's very useful in the C++ version.

lihaoyi-databricks:
Yes, let me open up another PR

PR #176, opened 2023-06-12
Release 0.4.5

lihaoyi-databricks:
WIP to cut a stable 0.4.5. Just some last minute fixes and tweaks to ensure a smooth rollout within Databricks

### DIFF
--- a/sjsonnet/src-jvm-native/sjsonnet/Config.scala
+++ b/sjsonnet/src-jvm-native/sjsonnet/Config.scala
@@ -123,4 +123,9 @@ case class Config(
     doc = "Evaluate the given string as Jsonnet rather than treating it as a file name"
   )
   exec: Flag = Flag(),
+  @arg(
+    name = "no-duplicate-keys-in-comprehension",
+    doc = "Raise an error if an object comprehension contains duplicate keys"
+  )
+  noDuplicateKeysInComprehension: Flag = Flag(),
 )
--- a/sjsonnet/src-jvm-native/sjsonnet/Config.scala
+++ b/sjsonnet/src-jvm-native/sjsonnet/Config.scala
@@ -133,4 +133,9 @@ case class Config(
     doc = """Raise an error if import expressions are used without proper parentheses, e.g. import "foo".bar rather than (import "foo").bar"""
   )
   strictImportSyntax: Flag = Flag(),
+  @arg(
+    name = "strict-inherited-assertions",
+    doc = """Properly handle assertions defined in a Jsonnet dictionary that is extended more than once"""
+  )
+  strictInheritedAssertions: Flag = Flag(),
 )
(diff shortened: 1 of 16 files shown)


sjsonnet/src-jvm-native/sjsonnet/Config.scala
```diff
@@ -123,4 +123,9 @@ case class Config(
     doc = "Evaluate the given string as Jsonnet rather than treating it as a file name"
   )
   exec: Flag = Flag(),
+  @arg(
+    name = "no-duplicate-keys-in-comprehension",
+    doc = "Raise an error if an object comprehension contains duplicate keys"
+  )
+  noDuplicateKeysInComprehension: Flag = Flag(),
```
szeiger, line 130: Can we make the correct behavior the default? We did the same when we added static checking.

sjsonnet/src/sjsonnet/Interpreter.scala
```diff
@@ -84,19 +84,21 @@ class Interpreter(extVars: Map[String, String],
...
+              defaults2(i) = parsed
+              tlaExpressions.add(parsed)
             }
             i += 1
           }
           new Val.Func(f.pos, f.defSiteValScope, Params(f.params.names, defaults2)) {
             def evalRhs(vs: ValScope, es: EvalScope, fs: FileScope, pos: Position) = f.evalRhs(vs, es, fs, pos)
             override def evalDefault(expr: Expr, vs: ValScope, es: EvalScope) = {
-              evaluator.visitExpr(expr)(ValScope.empty)
+              evaluator.visitExpr(expr)(if (tlaExpressions.exists(_ eq expr)) ValScope.empty else vs)
```
szeiger, line 101: When would this be called for an expr that is not in `tlaExpressions`?

sjsonnet/src-jvm-native/sjsonnet/Config.scala
lihaoyi-databricks: This is mostly to de-risk the version bump on the Databricks side of things. I'd rather make this opt-in and then go and turn it on piece-meal in a controlled fashion, rather than making it opt-out and then worrying about whether there's some place where we forgot to opt out that may blow up during the upgrade

sjsonnet/src/sjsonnet/Interpreter.scala
lihaoyi-databricks: It is not in `tlaExpressions` if it is a default value (as in `function(a=b) ...`) used when the tla expression is not present. 

For the TLA expressions, they should run in an empty `ValScope` since they kind of live in their own snippets and can't directly reference anything else, but the default values for the top-level functions should evaluate as part of the file they were defined in and be able to reference other things defined in that file

PR #193, opened 2023-11-17
allow custom trace callback

carl-db:
This adds the option to provide the interpreter with a custom callback for `std.trace`, instead of only logging to stderr. This will allow for capturing structured trace output for linting.


lihaoyi-databricks:
Looks good to me I think, just get CI passing and I can merge it

PR #194, opened 2023-12-20
Avoid Storing Large Strings in Memory

ahirreddy:
- This PR adds a new `ResolvedFile` interface that encapsulates all file reads (import, importstr, etc). Behind this a file can be backed by a string in memory (old behavior) or by a file on disk
- We implement `ParserInputs` for each - the in-memory one is trivial and the on-disk one is backed by a RandomAccessFile (with buffered reads around it for performance)
- Testing against our internal workloads - this greatly reduced memory presure as there are cases where we would previously store single files ranging between 100MB - 1GB in memory, in addition to its parsed AST.

# Misc
- Add support for customizing compression levels in std.xz. The default level (6) is very expensive (both in time and space) - so individual call sites can opt for lower levels
- Uses (path, CRC32) for the ParseCache key. CRC32 should be sufficient to detect changes in a given file - and is significantly more performant that easily available alternatives (~10x faster than MD5)


bench/src/main/scala/sjsonnet/MainBenchmark.scala
```diff
@@ -8,9 +8,12 @@ import org.openjdk.jmh.infra._
...
 object MainBenchmark {
   val mainArgs = Array[String](
-    "../../universe2/rulemanager/deploy/rulemanager.jsonnet",
-    "-J", "../../universe2",
-    "-J", "../../universe2/mt-shards/dev/az-westus-c2",
+    "../../universe/rulemanager/deploy/rulemanager.jsonnet",
+    // "../../universe/kubernetes/admission-controller/gatekeeper/deploy/gatekeeper.jsonnet",
+    "-J", "../../universe",
+    "-J", "../../universe/mt-shards/dev/az-westus-c2",
+    "-J", "../../universe/bazel-bin",
```
ahirreddy, line 15: Added these extra imports as they're now required to compile the given target. I also added the gatekeeper as another benchmark case.

lihaoyi-databricks:
Can we use a [fastparse.ReaderParserInput](https://github.com/com-lihaoyi/fastparse/blob/49ba3cf46705a5068541537991ce1c2167760a2b/fastparse/src/fastparse/ParserInput.scala#L222-L235) or `IteratorParserInput` here? This kind of use case is exactly what those guys are for (see [Streaming Parsing](https://com-lihaoyi.github.io/fastparse/#StreamingParsing) docs), and it would avoid us needing to maintain our own bespoke implementation

ahirreddy:
I tried using that one, but it doesn't support backtracking - which is used at some point when dealing with syntax errors

lihaoyi-databricks:
`ReaderParserInput` is meant to support backtracking, in that it is supposed to buffer up the input text until it hits a cut that allows it to call `dropBuffer`. Is that not working for some reason, is it or it failing with some error? Or is it simply keeping a larger buffer than is possible than with the random access interface that you are using?

On further thought, I wonder if using a `ParserInput` is giving us any benefit here: do we actually need to avoid reading individual files into memory for parsing, or is the change of `ParseCache` key to `(path, CRC)` sufficient? I would expect that the number of files being parsed at any point in time (O(10s)) would be dwarfed by the number of files that were previously being kept in the parse cache (O(10000s)), and so maybe reading the individual files into memory during parsing is sufficient as long as we don't keep the file contents around after

ahirreddy:
`ReaderParserInput` definitely doesn't support backtracking. It extends `BufferedParserInput` - for which [checkTraceable‎](https://github.com/com-lihaoyi/fastparse/blob/49ba3cf46705a5068541537991ce1c2167760a2b/fastparse/src/fastparse/ParserInput.scala#L185C15-L185C15) throws

```
  def checkTraceable() = throw new RuntimeException(
    s"Cannot perform `.traced` on an `${getClass.getName}`, as it needs to parse " +
      "the input a second time to collect traces, which is impossible after an " +
      "`IteratorParserInput` is used once and the underlying Iterator exhausted."
  )
```

I don't see any code that would make it work for Reader even if we changed checkTraceable. It think we'd need to call `mark(position)` on the underlying Reader and then reset() whenever we need to go backwards. That's basically what I did for our use case - just with a RandomAccessFile directly.

> On further thought, I wonder if using a custom ParserInput is giving us any benefit here: do we actually need to avoid reading individual files into memory for parsing,

On this point, we actually have 100MB-1GB input JSONs that take up a huge amount of memory. Depending on the order in which you process - you could easily OOM if you concurrently buffer and parse multiple of these (depending on the order in which Bazel requests a particular compile). I figured the easiest thing here was just to avoid the in memory buffering. There are further issues with 100+MB arrays - as they'll result in heap fragmentation and large object pinning - so even with plenty of space the GC may still declare OOM as it can't move the huge objects and can lose the ability to allocate new ones.

lihaoyi-databricks:
Got it. The first parse should be fine, but the call to `.traced` that we do on failure does a second parse that doesn't work with `java.io.Reader`. I think the current approach is fine then

lihaoyi-databricks:
I think this looks fine. @szeiger leaving this open for a bit for you to take a look before we merge it

sjsonnet/src/sjsonnet/Importer.scala
```diff
@@ -1,37 +1,172 @@
...
+  private val buffer = new Array[Byte](bufferSize)
+
+  private var bufferStart: Long = -1
+
+  private var bufferEnd: Long = -1
+
+  private val fileLength: Long = file.length()
+
+  private def fillBuffer(position: Long): Unit = {
+    file.seek(position)
```
lihaoyi-databricks, line 101: Do we always need to do a `file.seek` here? I imagine in the common case that `fillBuffer` gets called, `position` would simply be equivalent to `bufferEnd`, in which case the `seek` is unnecessary. Leaving it out would halve the number of filesystem calls.

sjsonnet/src/sjsonnet/Importer.scala
```diff
@@ -1,37 +1,172 @@
...
+  def readString(from: Long, until: Long): String = {
+    if (!(from < fileLength && until <= fileLength && from <= until)) {
+      throw new IndexOutOfBoundsException(s"Invalid range: $from-$until for file of length $fileLength")
+    }
+    val length = (until - from).toInt
+    val stringBytes = new Array[Byte](length)
+
+    if (from >= bufferStart && until <= bufferEnd) {
+      // Range is within the buffer
+      System.arraycopy(buffer, (from - bufferStart).toInt, stringBytes, 0, length)
```
lihaoyi-databricks, line 126: Could we directly return `new String(buffer, (from - bufferStart).toInt, stringBytes)` here? That would save a copy

sjsonnet/src/sjsonnet/Importer.scala
```diff
@@ -1,37 +1,172 @@
...
+    } else {
+      // Range is outside the buffer
+      file.seek(from)
+      file.readFully(stringBytes, 0, length)
+    }
+
+    new String(stringBytes)
```
lihaoyi-databricks, line 133: The current implementation uses `StandardCharsets.UTF_8`, so we should probably follow suit

sjsonnet/src-jvm-native/sjsonnet/SjsonnetMain.scala
```diff
@@ -291,4 +293,18 @@ object SjsonnetMain {
...
+
+  /**
+   * Read a path into a [[ResolvedFile]] if it exists and is a file. A resolved file acts as a layer
+   * of caching on top of the underlying file system. Small files are read into memory, while large
+   * files are read from disk.
+   */
+  private[this] def readPath(path: Path): Option[ResolvedFile] = {
+    val osPath = path.asInstanceOf[OsPath].p
+    if (os.exists(osPath) && os.isFile(osPath)) {
+      Some(new CachedResolvedFile(path.asInstanceOf[OsPath], memoryLimitBytes = 2048L * 1024L * 1024L))
```
lihaoyi-databricks, line 305: 2GB seems a bit large for a threshold for reading files from disk. You mentioned being worried about 100mb-1gb JSON files hogging memory during parse, but given this threshold none of those would trigger the streaming parsing code path at all. 

Are there really that many >2gb jsonnet files in universe that streaming parsing just for those makes the difference? Or should this threshold be lower, e.g. 10mb? Or were all our tests done accidentally with streaming parsing mostly disabled, and actually we don't need it and reading each file fully into memory during the parse is fine?

lihaoyi-databricks, line 305: What's the significance of this number? Should it just be `Int.MaxValue.toLong` or `Int.MaxValue.toLong + 1`?

szeiger:
Why are we keeping the parsed source files at all? AFAIR it's just for cache invalidation. Maybe we could replace them with hashes. Parsing could be done with a standard streaming parse or even load the entire file into memory temporarily.

ahirreddy:
@szeiger This PR does replace the cache entry with a CRC hash (was much faster than MD5- which made a huge difference for ~1GB files). Right now the PR will keep up to 1MB files in memory, but I can change that so that we always streaming parse.

sjsonnet/src/sjsonnet/Importer.scala
ahirreddy: Good point, added a check. If we're already at the right position in the file, we skip the seek.

sjsonnet/src/sjsonnet/Importer.scala
ahirreddy: Done

sjsonnet/src/sjsonnet/Importer.scala
ahirreddy: Done

sjsonnet/src-jvm-native/sjsonnet/SjsonnetMain.scala
ahirreddy: Changed to Int.MaxValue.toLong and added scaladoc:

```
 * @param memoryLimitBytes The maximum size of a file that we will resolve. This is not the size of
 * the buffer, but a mechanism to fail when being asked to resolve (and downstream parse) a file
 * that is beyond this limit.
```.

Basically, we have some pathological imports (1GB+) which I eventually want to ban (all the ones I found could be trivially modified upstream to not produce such huge files). In a followup we can make this param configurable.


cc @carl-db

File: sjsonnet/src/sjsonnet/Importer.scala [hunk 6]
```diff
@@ -46,9 +182,9 @@ class CachedResolver(
   val parseCache: ParseCache,
   strictImportSyntax: Boolean) extends CachedImporter(parentImporter) {
 
-  def parse(path: Path, txt: String)(implicit ev: EvalErrorScope): Either[Error, (Expr, FileScope)] = {
-    parseCache.getOrElseUpdate((path, txt), {
-      val parsed = fastparse.parse(txt, new Parser(path, strictImportSyntax).document(_)) match {
+  def parse(path: Path, content: ResolvedFile)(implicit ev: EvalErrorScope): Either[Error, (Expr, FileScope)] = {
+    parseCache.getOrElseUpdate((path, content.contentHash.toString), {
```
szeiger, line 186: Oh, I see, we're storing the hash as a string which explains why the ParseCache is unchanged.

jam01:
Sorry to haunt an old issue. Am I right in concluding that this PR would break scala.js compatability given these 

```
import java.io.{BufferedInputStream, BufferedReader, ByteArrayInputStream, File, FileInputStream, FileReader, InputStream, RandomAccessFile, Reader, StringReader}
```

in Importer. As in, not all of these are supported in scala.js javalib.

PR #196, opened 2024-01-03
Compact HashMaps

ahirreddy:
In this PR we compact [[LinkedHashMap]]s to be as small as possible. This greatly reduces the overhead for small Jsonnet objects (< 16 fields) - as the default size for a HashMap is 16.

There are 3 different cases for compaction:
1. If the map is empty, return [[Collections.emptyMap]]. This is an immutable singleton - and thus is space efficient.
2. If the map has 1 element, so we return an immutable single element map.
3. Otherwise, we return a new [[LinkedHashMap]] with the same contents as the original map that's been compacted to the smallest size possible - subject to the desired load factor: (num elements) / load-factor + 1. The default load factor is 0.75.

Return types from (1) & (2) are [[java.util.Map]] - where as 3 is a [[LinkedHashMap]]. So we need to pick the common type - which is [[java.util.Map]].

All returned maps preserve the insertion order of the original map. No map returned from this method should be mutated.

This change reduces the memory pressure in large concurrent builds. Benchmarking the exact improvement is difficult (allocation tracing is very expensive) - however, there is a net change in that a previously OOMing workload - no longer OOMs with this change.

We could further reduce memory usage by using `fastutil` linked hashmaps - which offer specialized maps for primitive types (Object -> Boolean in particular is useful for us) and generally have a smaller footprint - in addition to having built-in methods for size trimming. I've prototyped with fastutil and indeed observed an improvement, however, due to packaging and distribution complexities - I've avoided going down that route for now.


bench/src/main/scala/sjsonnet/MainBenchmark.scala
```diff
@@ -40,26 +43,32 @@ object MainBenchmark {
...
+        MainBenchmark.mainArgs,
+        cache,
+        System.in,
+        dummyOut,
+        System.err,
+        os.pwd,
+        None
+      )
+      println("hi")
+      Thread.sleep(100000000)
```
szeiger, line 71: Was this file committed by accident? The changes should probably be excluded (this one in particular). If we're really changing what the benchmark runs, the README should be updated to reflect this.

sjsonnet/src-jvm/sjsonnet/Platform.scala
```diff
@@ -24,13 +32,16 @@ object Platform {
...
+    ""
+    /*
     val outputStream: ByteArrayOutputStream = new ByteArrayOutputStream(b.length)
     val xz: XZOutputStream = new XZOutputStream(outputStream, new LZMA2Options())
     xz.write(b)
     xz.close()
     val xzedBase64: String = Base64.getEncoder.encodeToString(outputStream.toByteArray)
     outputStream.close()
     xzedBase64
+    */
```
szeiger, line 44: another accidental change?

sjsonnet/src-jvm/sjsonnet/Platform.scala
```diff
@@ -45,4 +56,78 @@ object Platform {
...
+   *    (num elements) / load-factor + 1. The default load factor is 0.75.
+   *
+   * Return types from (1) & (2) are [[java.util.Map]] - where as 3 is a [[LinkedHashMap]]. So we
+   * need to pick the common type - which is [[java.util.Map]].
+   *
+   * All returned maps preserve the insertion order of the original map. No map returned from this
+   * method should be mutated.
+   */
+  /*
+  def compactHashMap[K, V: TypeTag](map: LinkedHashMap[K, V]): JMap[K, V] = {
```
szeiger, line 76: Don't use Scala reflection for simple stuff like this. (It is almost always the wrong choice.) All you need is a `ClassTag` for Java reflection.

sjsonnet/src-jvm/sjsonnet/Platform.scala
```diff
@@ -45,4 +56,78 @@ object Platform {
...
+    val res = typeOf[V] match {
+      case t if t =:= typeOf[java.lang.Boolean] => "It's an Int"
```
szeiger, line 78: "It's an Int"?

sjsonnet/src-jvm/sjsonnet/Platform.scala
```diff
@@ -45,4 +56,78 @@ object Platform {
...
+    val size = map.size
+    if (size == 0) {
+      java.util.Collections.emptyMap[K, V]()
+    } else {
+      val builder = ImmutableMap.builder[K, V]()
+      map.forEach { case (key, value) => builder.put(key, value) }
+      builder.build()
+    }
+  }
+  */
```
szeiger, line 132: Some unused code down here

sjsonnet/src/sjsonnet/Evaluator.scala
```diff
@@ -598,7 +598,7 @@ class Evaluator(resolver: CachedResolver,
           builder.put(k, v)
         }
     }
-    cachedObj = new Val.Obj(objPos, builder, false, if(asserts != null) assertions else null, sup)
+    cachedObj = new Val.Obj(objPos, Platform.compactHashMap(builder), false, if(asserts != null) assertions else null, sup)
```
szeiger, line 601: I don't remember all the different call sites and didn't look into each one of them for this review. Can we know the size up-front and do it more efficiently (i.e. construct the right type directly instead of compacting later)? At least the type should be known in all cases.

ahirreddy:
Closing in favor of https://github.com/databricks/sjsonnet/pull/197

sjsonnet/src/sjsonnet/Evaluator.scala
ahirreddy: Makes sense! I"m doing that in the followup - https://github.com/databricks/sjsonnet/pull/197 that replaces/subsumes the work from this PR.

sjsonnet/src-jvm/sjsonnet/Platform.scala
ahirreddy: chat gpt copypasta

sjsonnet/src-jvm/sjsonnet/Platform.scala
ahirreddy: Yeah, I mostly did this for expediency of testing. I was less concerned with performance and instead wanted to quickly iterate on the heap-usage of the fully processed + optimized AST. The reflection was clearly slowing down the benchmark by 2x.

sjsonnet/src-jvm/sjsonnet/Platform.scala
ahirreddy: This was to speed up my interation during the benchmark. The file in question spends ~50% of its time in xz - so I quickly just hacked this out.

bench/src/main/scala/sjsonnet/MainBenchmark.scala
ahirreddy: Yeah sorry, this branch turned into an experiement branch after I submitted it to. you guys for review. Instead: https://github.com/databricks/sjsonnet/pull/197 is the final PR

PR #197, opened 2024-01-04
Static Object Optimizations

ahirreddy:
- This PR uses the Parser & StaticOptimizer thread local string interner for keys in static objects
- Similarly, we deduplicate the String -> Boolean map used to determine if a field is static.
- For static objects we also use immutable.VectorMap (with a JavaWrapper) for the field set.
- Lastly for the value cache, we size it according to the number of keys in the object this reduces unnecessary up-sizing for large objects, and more importantly removes the large number of sparse maps we previously had for small objects (the default was 16 elements)

Before: 855MB for the parsed file
image (9)

After: 425MB
image (10)

### DIFF
--- a/bench/src/main/scala/sjsonnet/MainBenchmark.scala
+++ b/bench/src/main/scala/sjsonnet/MainBenchmark.scala
@@ -43,10 +43,10 @@ object MainBenchmark {
 }
 
 @BenchmarkMode(Array(Mode.AverageTime))
-@Fork(1)
+@Fork(4)
 @Threads(1)
-@Warmup(iterations = 5)
-@Measurement(iterations = 10)
+@Warmup(iterations = 30)
+@Measurement(iterations = 40)
 @OutputTimeUnit(TimeUnit.MILLISECONDS)
 @State(Scope.Benchmark)
 class MainBenchmark {
@@ -74,6 +74,10 @@ class MainBenchmark {
 @Measurement(iterations = 10)
 @OutputTimeUnit(TimeUnit.MILLISECONDS)
 @State(Scope.Benchmark)
+// This is a dummy benchmark to see how much memory is used by the interpreter.
+// You're meant to execute it, and once it prints "sleeping" you can attach yourkit and take a heap
+// dump. Because we store the cache, the parsed objects will have strong references - and thus will
+// be in the heap dump.
 class MemoryBenchmark {
 
   val dummyOut = MainBenchmark.createDummyOut
--- a/bench/src/main/scala/sjsonnet/MainBenchmark.scala
+++ b/bench/src/main/scala/sjsonnet/MainBenchmark.scala
@@ -66,3 +66,34 @@ class MainBenchmark {
     ))
   }
 }
+
+@BenchmarkMode(Array(Mode.AverageTime))
+@Fork(1)
+@Threads(1)
+@Warmup(iterations = 5)
+@Measurement(iterations = 10)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@State(Scope.Benchmark)
+class MemoryBenchmark {
+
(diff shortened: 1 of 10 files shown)


bench/src/main/scala/sjsonnet/MainBenchmark.scala
```diff
@@ -66,3 +66,38 @@ class MainBenchmark {
...
+@Fork(1)
+@Threads(1)
+@Warmup(iterations = 5)
+@Measurement(iterations = 10)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@State(Scope.Benchmark)
+// This is a dummy benchmark to see how much memory is used by the interpreter.
+// You're meant to execute it, and once it prints "sleeping" you can attach yourkit and take a heap
+// dump. Because we store the cache, the parsed objects will have strong references - and thus will
+// be in the heap dump.
```
szeiger, line 80: Could we turn this into a simple command line program that does a gc before and after parsing and then prints the memory usage diff to the console so you can easily retest this for future changes without having to attach a profiler?

sjsonnet/src/sjsonnet/StaticOptimizer.scala
```diff
@@ -16,6 +18,11 @@ class StaticOptimizer(ev: EvalScope, std: Val.Obj) extends ScopedExprTransform {
     } else throw e
   }
 
+  // HashMap to deduplicate strings.
+  private[this] val strings = new mutable.HashMap[String, String]
+
+  private[this] val fieldSet = new mutable.HashMap[Val.StaticObjectFieldSet, java.util.LinkedHashMap[String, java.lang.Boolean]]
+
```
szeiger, line 25: Why is this cache separate from the `Parser`'s? Should it be a single cache at the `Interpreter` level?

sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -297,15 +298,45 @@ object Val{
     }
   }
 
-  def staticObject(pos: Position, fields: Array[Expr.Member.Field]): Obj = {
+  final case class StaticObjectFieldSet(keys: Array[String]) {
```
szeiger, line 301: This shouldn't be a case class. `Array` has no useful equality or `toString` and you're overriding `equals` and `hashCode`.

szeiger (commented):
What's the performance impact of these changes? Are we trading memory for speed?

sjsonnet/src/sjsonnet/Val.scala
ahirreddy: Fixed

sjsonnet/src/sjsonnet/StaticOptimizer.scala
ahirreddy: Changed, there's now a single cache at the interpreter level.

bench/src/main/scala/sjsonnet/MainBenchmark.scala
ahirreddy: Done and updated readme.

ahirreddy:
Sorry I missed your most important question. The performance impact here was undetectable in the benchmark. I think the fact that the object interning is thread-local and unsynchronized makes it pretty fast.

bench/src/main/scala/sjsonnet/OptimizerBenchmark.scala
```diff
@@ -3,6 +3,8 @@ package sjsonnet
 import java.io.StringWriter
 import java.util.concurrent.TimeUnit
 
+import scala.collection.mutable.HashMap
```
szeiger, line 6: I'd prefer to keep mutable types qualified (i.e. only import `scala.collection.mutable`) everywhere for consistency.

szeiger (approved):

PR #209, opened 2024-10-21
Scala 3 Migration

LaurelineSwissBorg:
Migrates the code to Scala 3 and enables cross-compilation with Scala 2.13.

Libraries were updated to their latest available version.


stephenamar-db:
Thanks for contributing, in the end, we are going with a more recent attempt to migrate to scala3. The timing was just not right in Oct 2024 :/

Issue #185, opened 2023-09-11
Formatting (std.mod, %) does not perform type check


zapster:
See the following reproducer:
```
$ cat test.jsonnet
[ "test%d" % "3" ]
$ sjsonnet test.jsonnet
[
   "test3"
]
$ jsonnet test.jsonnet
RUNTIME ERROR: Format required number at 0, got string
        <std>:(588:11)-(589:47) function <format_code>
        <std>:715:15-60 thunk <s> from <function <format_codes_arr>>
        <std>:720:24-25 thunk from <thunk <s_padded> from <function <format_codes_arr>>>
        <std>:480:30-33 thunk from <thunk from <function <pad_left>>>
        <std>:480:19-34 thunk from <function <pad_left>>
        <std>:476:11-12 thunk from <function <padding>>
        <std>:472:12-13 function <aux>
        <std>:476:7-17  function <padding>
        <std>:480:7-38  function <pad_left>
        <std>:720:15-39 thunk <s_padded> from <function <format_codes_arr>>
        <std>:726:55-63 thunk from <function <format_codes_arr>>
        <std>:726:11-64 function <format_codes_arr>
        <std>:774:7-48  function <anonymous>
        <std>:237:7-23  function <anonymous>
        test.jsonnet:1:3-17
        During manifestation
```

So the reference implementation issues an error while `sjsonnet` accepts the input.

According to the [reference](https://jsonnet.org/ref/stdlib.html#math):
> The function std.mod(a, b) is what the % operator is desugared to. It performs modulo arithmetic if the left hand side is a number, or if the left hand side is a string, it does Python-style string formatting with std.format(). 

Python issues an error as well:
```
python -c "'test%d' % '3'"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
TypeError: %d format: a real number is required, not str
```

Issue #139, opened 2021-11-17
%d in format strings should only accept numbers


nicklan:
Example file:
```jsonnet
"%(thing)d" % {
  thing: "NOT A NUMBER"
}
```

go-jsonnet:
```
$ jsonnet num.jsonnet
RUNTIME ERROR: Format required number at thing, got string
	<std>:(588:11)-(589:47)	function <format_code>
	<std>:761:15-50	thunk <s> from <function <format_codes_obj>>
	<std>:766:24-25	thunk from <thunk <s_padded> from <function <format_codes_obj>>>
	<std>:480:30-33	thunk from <thunk from <function <pad_left>>>
	<std>:480:19-34	thunk from <function <pad_left>>
	<std>:476:11-12	thunk from <function <padding>>
	<std>:472:12-13	function <aux>
	<std>:476:7-17	function <padding>
	<std>:480:7-38	function <pad_left>
	<std>:766:15-35	thunk <s_padded> from <function <format_codes_obj>>
	<std>:767:51-59	thunk from <function <format_codes_obj>>
	<std>:767:11-60	function <format_codes_obj>
	<std>:772:7-43	function <anonymous>
	<std>:237:7-23	function <anonymous>
	/tmp/num.jsonnet:(1:1)-(3:2)	
	During evaluation
```

sjsonnet:
```
$ sjsonnet num.jsonnet
"NOT A NUMBER"
```

PR #212, opened 2024-11-05
Add better error handling for format

stephenamar-db:
Add a check to make sure the format array is of the right size and type

- fix https://github.com/databricks/sjsonnet/issues/139
- fix https://github.com/databricks/sjsonnet/issues/185

### DIFF
--- a/sjsonnet/src/sjsonnet/Format.scala
+++ b/sjsonnet/src/sjsonnet/Format.scala
@@ -101,7 +101,9 @@ object Format{
       val cooked0 = formatted.conversion match{
         case '%' => widenRaw(formatted, "%")
         case _ =>
-
+          if  (values.isInstanceOf[Val.Arr] && i >= values.cast[Val.Arr].length) {
+            Error.fail("Too few values to format: %d, expected at least %d".format(values.cast[Val.Arr].length, i + 1))
+          }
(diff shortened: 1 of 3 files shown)


sjsonnet/src/sjsonnet/Format.scala
```diff
@@ -101,7 +101,9 @@ object Format{
       val cooked0 = formatted.conversion match{
         case '%' => widenRaw(formatted, "%")
         case _ =>
-
+          if  (values.isInstanceOf[Val.Arr] && i >= values.cast[Val.Arr].length) {
```
jordanmoxon-db, line 104: Does this `i` print work correctly? For instance, if I give "%s %s %s %s %s" % ["one"] I would expect "Too few values to format: 5, expected 1" where it seems from the iteration this will fail at i = 1, giving a confusing error message.

stephenamar-db: yeah good catch. may be I can rephrase to "at least i values"?

jordanmoxon-db: Yeah, that sounds good. I don't think it's worth it to do extra parsing just to make the error report the real size of the collection.

PR #213, opened 2024-11-07
Implement strict set operations

stephenamar-db:
Implement strict operations, enabled by `--strict-set-operations`.
This PR also adds `--throw-error-for-invalid-sets`, which should help find people not passing sets to the set methods.

Finally, it fixes an inefficiency in std.member()

### DIFF
--- a/sjsonnet/src/sjsonnet/Std.scala
+++ b/sjsonnet/src/sjsonnet/Std.scala
@@ -6,9 +6,8 @@ import java.util.Base64
 import java.util
 import java.util.regex.Pattern
 import sjsonnet.Expr.Member.Visibility
-import sjsonnet.Expr.BinaryOp
 
-import scala.collection.Searching.Found
+import scala.collection.Searching._
 import scala.collection.mutable
 import scala.util.matching.Regex
 
@@ -17,7 +16,7 @@ import scala.util.matching.Regex
   * in Scala code. Uses `builtin` and other helpers to handle the common wrapper
   * logic automatically
   */
-class Std(settings: Settings) {
+class Std {
   private val dummyPos: Position = new Position(null, 0)
   private val emptyLazyArray = new Array[Lazy](0)
 
@@ -138,15 +137,7 @@ class Std(settings: Settings) {
   }
 
   private object Type extends Val.Builtin1("x") {
-    def evalRhs(x: Val, ev: EvalScope, pos: Position): Val = Val.Str(pos, x match {
-      case _: Val.Bool => "boolean"
-      case _: Val.Null => "null"
-      case _: Val.Obj => "object"
-      case _: Val.Arr => "array"
-      case _: Val.Func => "function"
-      case _: Val.Num => "number"
-      case _: Val.Str => "string"
-    })
+    def evalRhs(x: Val, ev: EvalScope, pos: Position): Val = Val.Str(pos, x.prettyName)
   }
 
   private object Format_ extends Val.Builtin2("str", "vals") {
@@ -178,6 +169,8 @@ class Std(settings: Settings) {
             current = func.apply2(c, Val.Str(pos, new String(Array(char))), pos.noOffset)(ev)
           }
           current
+
+        case _ => Error.fail("Cannot call foldl on " + arr.prettyName)
       }
 
 
@@ -202,6 +195,7 @@ class Std(settings: Settings) {
             current = func.apply2(Val.Str(pos, new String(Array(char))), c, pos.noOffset)(ev)
           }
           current
+        case _ => Error.fail("Cannot call foldr on " + arr.prettyName)
       }
     }
   }
@@ -454,10 +448,8 @@ class Std(settings: Settings) {
           }
           str.value.contains(secondArg)
         case a: Val.Arr =>
-          var c = 0
-          a.foreach(v => if(ev.equal(v, x)) c += 1)
-          c > 0
-        case x => Error.fail("std.member first argument must be an array or a string, got " + arr.prettyName)
+          a.asLazyArray.indexWhere(v => ev.equal(v.force, x)) >= 0
+        case _ => Error.fail("std.member first argument must be an array or a string, got " + arr.prettyName)
       })
     }
   }
@@ -671,6 +663,7 @@ class Std(settings: Settings) {
             case (Some(lChild), None) => k -> createMember{lChild}
             case (Some(lChild: Val.Obj), Some(rChild: Val.Obj)) => k -> createMember{recPair(lChild, rChild)}
             case (_, Some(rChild)) => k -> createMember{recSingle(rChild)}
+            case (None, None) => Error.fail("std.mergePatch: This should never happen")
           }
 
           Val.Obj.mk(mergePosition, kvs:_*)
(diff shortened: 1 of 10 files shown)


sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -1141,68 +1056,64 @@ class Std {
...
+
+      for (v <- a) {
+        if (!existsInSet(ev, pos, keyF, b, v.force) && !existsInSet(ev, pos, keyF, out.toArray, v.force)) {
+          out.append(v)
+        }
+      }
 
-      if (keyF.isInstanceOf[Val.False]) {
-        val ujson.Arr(mArr) = Materializer(args(1))(ev)
-        val mx = Materializer(args(0))(ev)
```
JoshRosen, line 1194: FWIW, I've found that removing the materializer here is an _enormous_ performance improver for some inputs. In one real-world case I saw a whole _third_ of all memory allocations occurring from this materializer call.

I noticed that a few other built-in functions have similar use of `Materializer` and I'm wondering whether there's a principled reason for that or whether it's safe for us to simply remove all non-output / non `toString` uses of `Materializer`.

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -1130,78 +1039,91 @@ class Std {
     builtinWithDefaults("setUnion", "a" -> null, "b" -> null, "keyF" -> Val.False(dummyPos)) { (args, pos, ev) =>
       val a = args(0) match {
         case arr: Val.Arr => arr.asLazyArray
-        case str: Val.Str => stringChars(pos, str.value).asLazyArray
+        case str: Val.Str if !ev.settings.strictSetOperations => stringChars(pos, str.value).asLazyArray
         case _ => Error.fail("Arguments must be either arrays or strings")
```
JoshRosen, line 1043: I think this error message, as written, is only valid for the old behavior. With strict set operations enabled, I think it should read `"Arguments must be arrays"`.

I see that this

```scala
match {
        case arr: Val.Arr => arr.asLazyArray
        case str: Val.Str if !ev.settings.strictSetOperations => stringChars(pos, str.value).asLazyArray
        case _ => Error.fail("Arguments must be either arrays or strings")
      }
```

style of block is pre-existingly copy-paste duplicated across all of the set operation implementations.

WDYT about factoring it out into a helper function and then cleaning up the error message to respect the new flag?

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -1281,6 +1203,49 @@ class Std {
...
   )
+
+  private def validateSet(ev: EvalScope, pos: Position, keyF: Val, arr: Val): Unit = {
+    if (ev.settings.throwErrorForInvalidSets) {
+      val sorted = uniqArr(pos.noOffset, ev, sortArr(pos.noOffset, ev, arr, keyF), keyF)
+      if (!ev.equal(arr, sorted)) {
+        if (ev.settings.strictSetOperations) {
+          Error.fail(arr.prettyName + " is not a set")
+        } else {
+          System.err.println(arr.prettyName + " is not a set")
```
JoshRosen, line 1214: Perhaps this should use use `ev.warn` instead of println`? `ev.warn` calls the `warnLogger` which is configured at higher layers.

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -1130,78 +1039,91 @@ class Std {
...
         case _ => Error.fail("Arguments must be either arrays or strings")
       }
 
+      val out = new mutable.ArrayBuffer[Lazy]
+
+      // The intersection will always be, at most, the size of the smallest set.
+      val sets = if (b.length < a.length) (b, a) else (a, b)
+      for (v <- sets._1) {
+        if (existsInSet(ev, pos, keyF, sets._2, v.force) &&
+          (ev.settings.strictSetOperations || !existsInSet(ev, pos, keyF, out.toArray, v.force))) {
```
JoshRosen, line 1081: In non-strict mode, this is going to call `out.toArray` on every element added and that's going to perform a lot of garbage allocation.

Given that we're using an `ArrayBuffer` here, I think we might be able to avoid those repeated array copies by simply passing `out` here and then in `existsInSet` changing it to `arr: mutable.IndexedSeq[Lazy]`.

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -1383,18 +1347,9 @@ class Std {
...
-        val o1KeyExpr = Materializer.toExpr(Materializer.apply(o1Key)(ev))(ev)
-        val o2KeyExpr = Materializer.toExpr(Materializer.apply(o2Key)(ev))(ev)
-
-        val comparisonExpr = Expr.BinaryOp(dummyPos, o1KeyExpr, BinaryOp.OP_!=, o2KeyExpr)
-        val exprResult = ev.visitExpr(comparisonExpr)(ValScope.empty)
-
-        val res = Materializer.apply(exprResult)(ev).asInstanceOf[ujson.Bool]
-
-        if (res.value) {
+        if (!ev.equal(o1Key, o2Key)) {
```
JoshRosen, line 1352: For my own intuition:

Why isn't this

```
if (!ev.equal(o1Key.force, o2Key.force))
```

?

I don't fully understand why we call `.force` in the other equality comparisons in other contexts so I'm unsure if we should be doing it here.

stephenamar-db: My understanding is that there are 2 ways of doing it in sjsonnet:

- You can materialize it to json with Materializer and then evaluate it
- You can for the evaluation of a Val and then compare it.

For consistency, I'm picking the latter.

JoshRosen: Yep, 100% on the same page about avoiding `Materializer` usage.

But what I'm asking here is about whether we force the evaluation of the key expression vs. comparing lazy, e.g. why this is

```
if (!ev.equal(o1Key, o2Key))
```

and not

```
if (!ev.equal(o1Key.force, o2Key.force))
```

My understanding is that `ev.equal` has some fastpaths to allow for comparison of values without actually computing them, e.g. if we have a lazy not-yet-evaluated `x` then we can sometimes determine that x == x` via reference equality.

I noticed that some of the `ev.equal` call sites call `.force` in order to guarantee that the values are computed and I'm wondering why we do that. Does it have something to do with formal semantics of when evaluation occurs? I don't think it's necessary for correctness w.r.t. the final answer, but it might affect the side-effecting behavior of whether evaluation is performed and whether assertions in the to-be-computed subtree are evaluated?

stephenamar-db: ah ok.
unclear as to why ev.equal does not take Lazy.
in the end, it does not matter, because right now, we force the evaluation of all parameters (not inside the arrays though) in Var.Builtin()

```
    final def evalRhs(scope: ValScope, ev: EvalScope, fs: FileScope, pos: Position): Val = {
      val args = new Array[Val](params.names.length)
      var i = 0
      var j = scope.length - args.length
      while(i < args.length) {
        args(i) = scope.bindings(i).force
        i += 1
      }
      evalRhs(args, ev, pos)
    }
```

JoshRosen (approved):
LGTM!

I was unsure about whether our test harness tests all of the combinations of flags, so I also manually ran the unit tests with strict set operations disabled and confirmed that they all pass 👍

Issue #200, opened 2024-02-06
std.extVar are resolved as FilePath in windows


lmartella1:
Hello,
When attempting to load jsonnet/json values via std.extVar(), a stacktrace appear with sjsonnet when running in Windows. The same input files and sjsonnet jar works as expected in other OS as Unix or MacOS.

**Reproducer**
```
# example.jsonnet
local x = std.extVar('x');
x
```

```
# reproduce.bash
./sjsonnet-0.4.9.jar --ext-code x="6057" example.jsonnet

#gives the following stacktrace in Windows
sjsonnet.Error: Internal Error
    at [ApplyBuiltin1].(example.jsonnet:1:21)
    at [ValidId x].(example.jsonnet:2:1)
Caused by: java.nio.file.InvalidPathException: Illegal char <<> at index 0: <ext-var x>
    at java.base/sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
    at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
    at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
    at java.base/sun.nio.fs.WindowsPath.parse(WindowsPath.java:92)
    at java.base/sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:232)
    at java.base/java.nio.file.Path.resolve(Path.java:515)
    at os.Path.$div(Path.scala:440)
    at sjsonnet.OsPath.$div(OsPath.scala:11)
    at sjsonnet.Interpreter.parseVar(Interpreter.scala:43)
    at sjsonnet.Interpreter.$anonfun$evaluator$2(Interpreter.scala:49)
    at scala.Option.map(Option.scala:242)
    at sjsonnet.Interpreter.$anonfun$evaluator$1(Interpreter.scala:49)
    at sjsonnet.Std$ExtVar$.evalRhs(Std.scala:544)
    at sjsonnet.Evaluator.visitApplyBuiltin1(Evaluator.scala:223)
    at sjsonnet.Evaluator.visitExpr(Evaluator.scala:37)
    at sjsonnet.Evaluator$$anonfun$visitAsLazy$2.compute(Evaluator.scala:83)
    at sjsonnet.Lazy.force(Val.scala:23)
    at sjsonnet.Evaluator.visitValidId(Evaluator.scala:88)
[...]

PR #208, opened 2024-08-22
fix: Use different chars for synthetic paths.

cfraenkel:
this fixes --exec, --tla-str and --ext-str on windows as < and > are illegal file name characters.

fixes #200

PR #215, opened 2024-11-22
Flag use of Unicode less-than and greater-than characters in FilePath to be a Windows-only change

JoshRosen:
This PR fixes a regression introduced by https://github.com/databricks/sjsonnet/issues/208

That PR aimed to fix https://github.com/databricks/sjsonnet/issues/200, an issue where `std.extVar` failed on Windows because it tried to resolve mock paths which contained `<` and `>` characters which are forbidden in Windows filenames.

That PR's solution was to use Unicode less-than and greater-than characters in place of ASCII `<` and `>`, but that broke things for Unix/Linux platforms with a non-UTF8 `LANG` (see https://github.com/databricks/sjsonnet/pull/208#pullrequestreview-2452961190).

In this PR, I aim to fix this by flagging the other PR's change to only occur for Windows, while continuing to use ASCII `<` and `>` as before on other platforms.

I also added a regression test by pinning `LANG=C` in our GitHub Actions test setup. This successfully reproduced the bug fixed in this PR, and also uncovered a minor test-only issue related to the use of default character encodings (which I've fixed by pinning the test code to UTF-8).

Note that these file paths aren't actually materialized onto disk: rather, these are "placeholder / mock" paths used to indicate when jsonnet is reading from synthetic paths. In theory, we might be able to avoid the need to ever perform resolution in the first place by changing uses of these paths to be an `Either` of either an actual `os.Path` or a placeholder, but that's a much larger and more invasive change than I want to make right now. Here, I've chosen to pursue a narrowly-targeted tactical fix-forward.

### DIFF
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -25,4 +25,4 @@ jobs:
       - name: Compile with sbt
         run: sbt compile bench/compile bench/jmh:compile
       - name: Run sbt tests
-        run: sbt test
+        run: echo "Lang is $LANG" && sbt test
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -12,6 +12,9 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
+    env:
+      # Set LANG=C to simulate least-common-denominator target deployment environments:
+      LANG: C
 
     steps:
       - uses: actions/checkout@v2
@@ -25,4 +28,4 @@ jobs:
       - name: Compile with sbt
         run: sbt compile bench/compile bench/jmh:compile
       - name: Run sbt tests
-        run: echo "Lang is $LANG" && sbt test
+        run: sbt test
--- a/sjsonnet/src/sjsonnet/Util.scala
+++ b/sjsonnet/src/sjsonnet/Util.scala
@@ -34,7 +34,10 @@ object Util{
     }
   }
 
-  val isWindows: Boolean = System.getProperty("os.name").toLowerCase.startsWith("windows")
+  val isWindows: Boolean = {
+    // This is normally non-null on the JVM, but it might be null in ScalaJS hence the Option:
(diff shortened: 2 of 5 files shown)


sjsonnet/src/sjsonnet/Util.scala
```diff
@@ -33,4 +33,20 @@ object Util{
         new String(range.dropWhile(_ < 0).takeWhile(_ < s.length).map(s).toArray)
     }
   }
+
+  val isWindows: Boolean = System.getProperty("os.name").toLowerCase.startsWith("windows")
```
JoshRosen, line 37: This matches https://github.com/databricks/sjsonnet/blob/759cea713bc8e39f45b844a75358fd780f75d480/sjsonnet/client/src/sjsonnet/client/Util.java#L11, but we can't directly that other implementation here.

.github/workflows/pr-build.yaml
```diff
@@ -12,6 +12,9 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
+    env:
+      # Set LANG=C to simulate least-common-denominator target deployment environments:
+      LANG: C
```
JoshRosen, line 17: Tested this over in my fork at https://github.com/JoshRosen/sjsonnet/pull/1 and confirmed that it reproduced the bug.

sjsonnet/src/sjsonnet/Util.scala
JoshRosen: Come to think of it, though, I'm wondering whether this should be lifted to `Platform` to ensure it won't blow up on Scala JS.

JoshRosen: Yep, we need to wrap in options to avoid an NPE under ScalaJS: https://github.com/databricks/sjsonnet/pull/215/commits/d745f75eeb855f9d57216387ab7c32c177776054

See https://scastie.scala-lang.org/VuigbI2gS5G1WOnTYCdTdg for seeing some default properties in ScalaJS (look at Javascript console output after running).

stephenamar-db (approved):
Thanks for fixing this bug!

PR #216, opened 2024-11-22
Optimize std.findSubstr and std.flattenArrays

JoshRosen:
This PR implements small performance optimizations for two built-in functions:

- **std.findSubstr**:
  - Skip array allocation in case of no matches. This is a common scenario because `findSubstr` is often used to implement "string contains" checks.
  - Use `ArrayBuilder` instead of `ArrayBuffer`: in this context we only need to append and don't need to update, remove, or inspect the elements that we've written, and for those requirements ArrayBuilder is cheaper to instantiate and cheaper to invoke.
  - Store `Val.Num` into the array builder rather than storing indices then mapping the result into an output array: this saves an array allocation and iteration.
- **std.flattenArrays**:
  - Apply a similar ArrayBuffer -> ArrayBuilder replacement.

### DIFF
--- a/sjsonnet/src/sjsonnet/Std.scala
+++ b/sjsonnet/src/sjsonnet/Std.scala
@@ -463,15 +463,15 @@ class Std {
 
   private object FlattenArrays extends Val.Builtin1("arrs") {
     def evalRhs(arrs: Val, ev: EvalScope, pos: Position): Val = {
-      val out = new mutable.ArrayBuilder.ofRef[Lazy]
+      val out = new mutable.ArrayBuffer[Lazy]
       for(x <- arrs.asArr) {
         x match{
           case Val.Null(_) => // do nothing
-          case v: Val.Arr => out.addAll(v.asLazyArray)
+          case v: Val.Arr => out.appendAll(v.asLazyArray)
           case x => Error.fail("Cannot call flattenArrays on " + x)
         }
       }
-      new Val.Arr(pos, out.result())
+      new Val.Arr(pos, out.toArray)
     }
   }
   private object Reverse extends Val.Builtin1("arrs") {
--- a/sjsonnet/src/sjsonnet/Std.scala
+++ b/sjsonnet/src/sjsonnet/Std.scala
@@ -922,13 +922,16 @@ class Std {
     builtin("findSubstr", "pat", "str") { (pos, ev, pat: String, str: String) =>
       if (pat.length == 0) new Val.Arr(pos, emptyLazyArray)
       else {
-        val indices = mutable.ArrayBuffer[Int]()
         var matchIndex = str.indexOf(pat)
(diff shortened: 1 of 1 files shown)


stephenamar-db:
Do you want to merge this?

JoshRosen:
Hmm, it looks like we're getting a compilation error in the scalajs 2.12 code because we're missing `ArrayBuilder.addAll`:

```
[info] compiling 31 Scala sources to /home/runner/work/sjsonnet/sjsonnet/out/sjsonnet/2.12.18/js/compile.dest/classes ...
Error:  /home/runner/work/sjsonnet/sjsonnet/sjsonnet/src/sjsonnet/Std.scala:470:34: value addAll is not a member of scala.collection.mutable.ArrayBuilder.ofRef[sjsonnet.Lazy]
Error:            case v: Val.Arr => out.addAll(v.asLazyArray)
Error:                                   ^
Error:  one error found
```

I've gone ahead and backed out the `flattenArrays` optimization and have updated this PR title and description to only reflect the `findSubstr` optimizations (which are probably the more impactful of the two, to be honest).

I still think we should merge that optimization because it's a cheap win and actually was a hotspot (relatively speaking) in some files I profiled.

stephenamar-db:
> Hmm, it looks like we're getting a compilation error in the scalajs 2.12 code because we're missing `ArrayBuilder.addAll`:
> 
> ```
> [info] compiling 31 Scala sources to /home/runner/work/sjsonnet/sjsonnet/out/sjsonnet/2.12.18/js/compile.dest/classes ...
> Error:  /home/runner/work/sjsonnet/sjsonnet/sjsonnet/src/sjsonnet/Std.scala:470:34: value addAll is not a member of scala.collection.mutable.ArrayBuilder.ofRef[sjsonnet.Lazy]
> Error:            case v: Val.Arr => out.addAll(v.asLazyArray)
> Error:                                   ^
> Error:  one error found
> ```
> 
> I've gone ahead and backed out the `flattenArrays` optimization and have updated this PR title and description to only reflect the `findSubstr` optimizations (which are probably the more impactful of the two, to be honest).
> 
> I still think we should merge that optimization because it's a cheap win and actually was a hotspot (relatively speaking) in some files I profiled.

I believe you can use the ++= operator in-lieu of addAll

JoshRosen:
> I believe you can use the ++= operator in-lieu of addAll

Ah, good call: it looks like that's a synonym for an `addAll` in 2.13 and in both 2.12 and 2.13 it looks like it has fast-path optimizations for the case where you're adding from an array. Those optimizations are implemented slightly differently in 2.12 vs. 2.13 but I believe they'll both have the same net effect of bottoming out in a System.arrayCopy.

I've gone ahead and incorporated this suggestion, adding `++=` and reverting back to the old PR title and description.

stephenamar-db (approved):

PR #217, opened 2024-11-23
Fix thread-safety bug in Obj.getAllKeys

JoshRosen:
This PR fixes a thread-safety bug in Val.obj.getAllKeys, similar to the previous fix in https://github.com/databricks/sjsonnet/pull/136. It looks like this particular site was overlooked in that earlier PR.

I spotted this potential issue while reading through the code, not via actual use.

### DIFF
--- a/sjsonnet/src/sjsonnet/Val.scala
+++ b/sjsonnet/src/sjsonnet/Val.scala
@@ -194,8 +194,10 @@ object Val{
 
     private def getAllKeys = {
       if(allKeys == null) {
-        allKeys = new util.LinkedHashMap[String, java.lang.Boolean]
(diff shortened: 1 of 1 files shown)


JoshRosen:
On reflection, I'm actually unsure about whether this race condition can occur in practice:

If the parse cache is the means by which objects can be shared across threads and that can only cache static objects, then:

- Static objects have `allKeys` pre-assigned,
- thus we'll never attempt to initialize `allKeys`,
- thus we'll not hit a race condition here.

In contrast, I think that `getValue0` could potentially be invoked for a cached static object if `addSuper` is called, which could happen if we're using the binary `+` operation for object-object concatenation.

PR #223, opened 2024-12-05
Improve std.get and error messages

stephenamar-db:
This PR does 2 things:

It improves std.get() to only compute default when it's needed.

and it makes builtin error more readable
Before

```
bazel-bin/bazel/jsonnet/sjsonnet.jar -e 'std.get(["a", "b"], "a", null)'
sjsonnet.Error: Wrong parameter type: expected Object, got array
    at [ApplyBuiltin].(<exec>:1:8)
```

after
```
bazel-bin/bazel/jsonnet/sjsonnet.jar -e 'std.get(["a", "b"], "a", null)'
sjsonnet.Error: Wrong parameter type: expected Object, got array
    at [std.get].(<exec>:1:8)
````

### DIFF
--- a/sjsonnet/src/sjsonnet/Std.scala
+++ b/sjsonnet/src/sjsonnet/Std.scala
@@ -20,7 +20,7 @@ class Std {
   private val dummyPos: Position = new Position(null, 0)
   private val emptyLazyArray = new Array[Lazy](0)
 
-  private object AssertEqual extends Val.Builtin2("a", "b") {
+  private object AssertEqual extends Val.Builtin2("assertEqual", "a", "b") {
     def evalRhs(v1: Val, v2: Val, ev: EvalScope, pos: Position): Val = {
       val x1 = Materializer(v1)(ev)
       val x2 = Materializer(v2)(ev)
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: sjsonnet/src/sjsonnet/Std.scala [hunk 1]
```diff
@@ -79,76 +79,84 @@ class Std {
...
     }
-    class SpecF(f: String) extends Val.Builtin1("o") {
+    class SpecF(f: String) extends Val.Builtin1("objectHasAll", "o") {
       def evalRhs(o: Val, ev: EvalScope, pos: Position): Val =
         Val.bool(pos, o.asObj.containsKey(f))
     }
   }
 
-  private object ObjectFields extends Val.Builtin1("o") {
+  private object ObjectFields extends Val.Builtin1("objectHasAll", "o") {
```
JoshRosen, line 113: Typo: this should be 

```suggestion
  private object ObjectFields extends Val.Builtin1("objectFields", "o") {
```

sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -469,32 +470,25 @@ object Val{
   }
 
   /** Superclass for standard library functions */
-  abstract class Builtin(paramNames: Array[String], defaults: Array[Expr] = null)
+  abstract class Builtin(val functionName: String, paramNames: Array[String], defaults: Array[Expr] = null)
```
JoshRosen, line 473: A nice consequence of this new `functionName` thing is that we can now straightforwardly improve the `RunProfiler` output for built-in functions.

Today this outputs confusing anonymous class names like

```
[info] - 1029ms 762549  1350ns  sjsonnet.Std$$anon$8
[info] - 153ms  499763  307ns   sjsonnet.Std$$anon$5
```

but if we change

https://github.com/databricks/sjsonnet/blob/89a83b1dcf9ee5206be04525f25477ae50ad5f92/bench/src/main/scala/sjsonnet/ProfilingEvaluator.scala#L160-L161

to use `.functionName` then we get nice output like

```
[info] - 914ms  762032  1200ns  setMember
[info] - 133ms  233603  572ns   stripChars
```

PR #226, opened 2024-12-06
Fix parseInt() not supporting Longs

### LINKED ISSUE #191: std.parseInt only supports 32-bit integers
The stdlib function std.parseInt uses a Java Int (32-bit) instead of arbitrary length or at least a 64-bit Long.

stephenamar-db:
Fixes #191

### DIFF
--- a/sjsonnet/test/resources/test_suite/stdlib.jsonnet
+++ b/sjsonnet/test/resources/test_suite/stdlib.jsonnet
@@ -846,6 +846,9 @@ std.assertEqual(
 ) &&
 
 std.assertEqual(std.parseInt('01234567890'), 1234567890) &&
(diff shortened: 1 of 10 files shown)


sjsonnet/test/resources/test_suite/stdlib.jsonnet
```diff
@@ -846,6 +846,9 @@ std.assertEqual(
 ) &&
 
 std.assertEqual(std.parseInt('01234567890'), 1234567890) &&
+std.assertEqual(std.parseInt('01234567890'), 1234567890) &&
```
JoshRosen, line 849: Duplicated with the line above?

JoshRosen (approved):

PR #227, opened 2024-12-06
More missing methods implemented in sjsonnet

stephenamar-db:
- Basically almost every missing methods
- Also imported the most recent version of the c-jsonnet stdlib test file


sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -491,29 +529,22 @@ private object Get extends Val.Builtin("get", Array("o", "f", "default", "inc_hi
...
-        if(str.charAt(i) == c) {
-          val finalStr = Val.Str(pos, str.substring(start, i))
-          b.+=(finalStr)
-          start = i+1
-        }
-        i += 1
-      }
-      b.+=(Val.Str(pos, str.substring(start, math.min(i, str.length))))
-      new Val.Arr(pos, b.result())
+      new Val.Arr(pos, _str.asString.split(java.util.regex.Pattern.quote(_c.asString), -1).map(s => Val.Str(pos, s)))
```
JoshRosen, line 532: This implementation will be slower than the original because it will need to internally allocate a fresh `Pattern` invocation on every iteration, whereas the old version avoided that.

I was actually considering trying to refactor `SplitLimit` to work closer to `Split`.

Alternatively, we might consider using specialization to cache the Pattern instances: I have a some work-in-progress draft code to do this for a couple of other built-ins.

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -1248,7 +1280,7 @@ private object Get extends Val.Builtin("get", Array("o", "f", "default", "inc_hi
       str.isEmpty
     },
     builtin("trim", "str") { (_, _, str: String) =>
-      str.trim
+      str.replaceAll("^[ \t\n\f\r\u0085\u00A0']+", "").replaceAll("[ \t\n\f\r\u0085\u00A0']+$", "")
```
JoshRosen, line 1283: Because the regex is a constant here, I think we can factor these out into a static Pattern instance in a companion and then use it's `replaceAll`: that will save a ton of allocations.

JoshRosen (approved):
LGTM!

PR #231, opened 2024-12-10
Implement quote_keys and c_document_end for yaml functions

stephenamar-db:
Forked from PR https://github.com/databricks/sjsonnet/pull/143


sjsonnet/src/sjsonnet/BaseCharRenderer.scala
```diff
@@ -136,20 +145,20 @@ class BaseCharRenderer[T <: upickle.core.CharOps.Output]
...
 
     if (s eq null) visitNull(index)
     else visitNonNullString(s, index)
   }
 
-  def visitNonNullString(s: CharSequence, index: Int) = {
+  private def visitNonNullString(s: CharSequence, index: Int) = {
     flushBuffer()
-    upickle.core.RenderUtils.escapeChar(unicodeCharBuilder, elemBuilder, s, escapeUnicode)
+    upickle.core.RenderUtils.escapeChar(null, elemBuilder, s, escapeUnicode)
```
JoshRosen, line 156: It looks like this `unicodeCharBuilder` parameter is accepted by `escapeChar` but isn't actually used by that code. Checking the history of that class in upickle, it looks like it's _never_ been used by that method.

I think it's probably fine to pass `null` here.

Is the intent here just to save a small amount of unnecessary allocation per `BaseCharRenderer` construction?

sjsonnet/src/sjsonnet/YamlRenderer.scala
```diff
@@ -155,9 +174,23 @@ class YamlRenderer(_out: StringWriter = new java.io.StringWriter(), indentArrayI
...
+  private val yamlTimestampPattern = Pattern.compile("^(?:[0-9]*-){2}[0-9]*$", Pattern.CASE_INSENSITIVE)
+  private val yamlBinaryPattern = Pattern.compile("^[-+]?0b[0-1_]+$")
+  private val yamlHexPattern = Pattern.compile("[-+]?0x[0-9a-fA-F_]+")
+
+  private def isSafeBareKey(k: String) = {
+    val l = k.toLowerCase
+    !yamlReserved.contains(l) && safeYamlKeyPattern.matcher(k).matches() &&
+      !yamlTimestampPattern.matcher(k).matches() && !yamlBinaryPattern.matcher(k).matches() &&
+      !yamlHexPattern.matcher(k).matches() && (Try(l.replace("_", "").toLong).isFailure
+      && Try(l.replace("_", "").toDouble).isFailure)
```
JoshRosen, line 190: Since most keys are valid, this will on average result in throwing and catching two exceptions per key, which is going to be disproportionately expensive compared to everything else in this method:

image

I think it might be possible to use an additional regex to check if something is an integer or float.

JoshRosen: In `jsonnet`, this is implemented at https://github.com/google/jsonnet/blob/913281d203578bb394995bacc792f2576371e06c/stdlib/std.jsonnet#L1119-L1134 . 

---

Because their stdlib is implemented in jsonnet itself and they lack regex, some of their implementation is a bit weird. For example:

```
      /* Check for integers.  Keys that meet all of the following:
           - all characters match [0-9_\-]
           - has at most 1 dash
         are considered integers.
      */
      else if onlyChars(intChars, keySetLc)
              && std.length(std.findSubstr('-', key)) < 2 then
```

I imagine that we don't intend to be _that_ jsonnet-compliant, though, since it doesn't look like even they have test cases for corner cases like `1-2-3`, etc.

---

I think we can pass all of the tests with

```scala
  val newlinePattern: Pattern = Pattern.compile("\n")
  private val safeYamlKeyPattern = Pattern.compile("^[a-zA-Z0-9/._-]+$")
  private val yamlReserved = Set("true", "false", "null", "yes", "no", "on", "off", "y", "n", ".nan",
    "+.inf", "-.inf", ".inf", "null", "-", "---", "''")
  private val yamlTimestampPattern = Pattern.compile("^(?:[0-9]*-){2}[0-9]*$")
  private val yamlBinaryPattern = Pattern.compile("^[-+]?0b[0-1_]+$")
  private val yamlHexPattern = Pattern.compile("[-+]?0x[0-9a-fA-F_]+")
  private val yamlFloatPattern = Pattern.compile(
    "^-?(?>[0-9_]*)*(?>\\.[0-9_]*)?(?>e[-+][0-9_]+)?$"
  )
  private val yamlIntPattern = Pattern.compile("^[-+]?[0-9_]+$")

  private def isSafeBareKey(k: String) = {
    val l = k.toLowerCase
    !yamlReserved.contains(l) &&
      safeYamlKeyPattern.matcher(k).matches() &&
      !yamlTimestampPattern.matcher(l).matches() &&
      !yamlBinaryPattern.matcher(k).matches() &&
      !yamlHexPattern.matcher(k).matches() &&
      !yamlFloatPattern.matcher(l).matches() &&
      !yamlIntPattern.matcher(l).matches()
  }
```

where I've gotten a bit fancy with the float pattern with the use of atomic groups to avoid backtracking. I've also chosen to handle the underscores in the regex rather than using `str.replace()` to strip them.

stephenamar-db: I kind of followed what go-sjsonnet is doing, but working around java's inconsistencies with _ separator (they are invalid in parseInt() functions.)
Let's use your version. Seems to work

sjsonnet/src/sjsonnet/BaseCharRenderer.scala
stephenamar-db: I think? It sometimes feel like a lot of microoptimizations are used and I'm always a bit wary of changing things

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -549,6 +549,7 @@ class Std {
         q.removeFirst().force match {
           case v: Val.Arr => v.asLazyArray.reverseIterator.foreach(q.push)
           case s: Val.Str => out.write(s.value)
+          case _ =>
```
JoshRosen, line 552: For compatibility with jsonnet behavior, I think this should throw an error rather than silently ignoring elements of unexpected types:

https://github.com/google/jsonnet/blob/913281d203578bb394995bacc792f2576371e06c/stdlib/std.jsonnet#L321-L327

Maybe we can change this in a followup / separate PR?

JoshRosen (approved):
LGTM!

PR #233, opened 2024-12-11
Stop manifesting functions with no arguments in objects by default

### LINKED ISSUE #168: Sjsonnet is more lenient than google/jsonnet in calling 0-argument function values during materialization
Input:

```jsonnet
{
  foo(): 123
}
```

google/jsonnet output:

```
Error: RUNTIME ERROR: couldn't manifest function as JSON
	Field "foo"	
	During manifestation	
```

Sjsonnet output:

```
{
  "foo": 123
}
```

It appears that google/jsonnet will only auto-evaluate a function during materialization if it's at the top level, whereas sjsonnet will auto-evaluate it anywhere in the jsonnet document

stephenamar-db:
This is inline with go-jsonnet's behavior.

This behavior difference made TLA vars working properly in sjsonnet - I moved this part to the interpreter.
This fixes #168

### DIFF
--- a/sjsonnet/src/sjsonnet/Interpreter.scala
+++ b/sjsonnet/src/sjsonnet/Interpreter.scala
@@ -107,7 +107,7 @@ class Interpreter(extVars: Map[String, String],
             override def evalDefault(expr: Expr, vs: ValScope, es: EvalScope) = {
               evaluator.visitExpr(expr)(if (tlaExpressions.exists(_ eq expr)) ValScope.empty else vs)
             }
-          }
+          }.apply0(f.pos)(evaluator)
         case x => x
       }
     } yield res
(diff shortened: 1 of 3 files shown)


sjsonnet/src/sjsonnet/Interpreter.scala
```diff
@@ -107,7 +107,7 @@ class Interpreter(extVars: Map[String, String],
             override def evalDefault(expr: Expr, vs: ValScope, es: EvalScope) = {
               evaluator.visitExpr(expr)(if (tlaExpressions.exists(_ eq expr)) ValScope.empty else vs)
             }
-          }
+          }.apply0(f.pos)(evaluator)
```
JoshRosen, line 110: The old code was calling

```
apply(Evaluator.emptyLazyArray, null, outerPos)
```

but this new code is calling

```scala
    def apply0(outerPos: Position)(implicit ev: EvalScope): Val = {
      if(params.names.length != 0) apply(Evaluator.emptyLazyArray, null, outerPos)
      else {
        val funDefFileScope: FileScope = pos match { case null => outerPos.fileScope case p => p.fileScope }
        evalRhs(defSiteValScope, ev, funDefFileScope, outerPos)
      }
    }
```

and I think these are evidently equivalent in case the function has parameters.

In the case where the function _doesn't_ have parameters, it looks like it goes into an `evalRhs` path but that just looks like an optimization to short-circuit parts of `apply()` (i.e. it corresponds to the `if(simple)` branch in `apply` but without having to generate garbage to handle args (which aren't present)).

Therefore I think these are behaviorally equivalent 👍

JoshRosen (approved):
LGTM!

PR #234, opened 2024-12-11
Fix "Duplicate Local Variables in Object Scope"

### LINKED ISSUE #178: `sjsonnet` accepts duplicate local variables whereas `jsonnet` doesn't
See the following jsonnet file:
```
$ cat duplicate_local.jsonnet
{
  local var = ['foo'],
  local var = var + ['bar'],
  var: var,
}
```
The `jsonnet` reference implementation complains:
```
$ jsonnet duplicate_local.jsonnet
duplicate_local.jsonnet:3:9-12 Duplicate local var: var

  local var = var + ['bar'],
```
However, `sjsonnet` happily accepts it:
```
$ sjsonnet duplicate_local.jsonnet
{
   "var": [
      "foo",
      "bar"
   ]
}
```
Independent of what makes more sense, I think `sjsonnet` should behave the same as `jsonnet`.

Jsonnet commandline interpreter v0.17.0
Sjsonnet 0.4.3

stephenamar-db:
Fixes #178

### DIFF
--- a/sjsonnet/src/sjsonnet/Expr.scala
+++ b/sjsonnet/src/sjsonnet/Expr.scala
@@ -18,6 +18,8 @@ trait Expr{
     val n = getClass.getName
     if(n.startsWith("sjsonnet.Expr$")) n.substring(14) else n
   }
+
+  override def toString: String = s"$exprErrorString@$pos"
 }
 object Expr{
   private final def arrStr(a: Array[_]): String = {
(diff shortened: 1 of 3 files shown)


JoshRosen (approved):
LGTM! Looks like the implementation here is consistent with the pre-existing logic for detecting duplicate function parameters or duplicate object members.

PR #236, opened 2024-12-12
Optimize lstripChars, rstripChars, and stripChars via specialization

JoshRosen:
This PR optimizes `lstripChars` / `rstripChars` / `stripChars` built-in functions by using the specialization framework from #119 / 0bd255a7486f4889b06581afeac8058548995cf3 to pre-compile and re-use `Pattern` instances when the replacement / strip arguments are constants.

In Java, `String.replaceAll()` compiles and uses a `Pattern` under the hood and this is relatively expensive; specialization lets us save this cost when the `chars`-to-be-stripped are constant.

## Testing

**Correctness**: ran all tests with a manual change to disable static function application optimizations (a prerequisite required to achieve full test coverage, as the static application prevents specialization from kicking in during most tests). In a followup, I think we should explore adding flags for optimizer features and running all tests with/without optimization.

**Performance**: ran benchmarks on a large file and with this change I save ~11% of allocated bytes and ~8% of wall time. I ran that same file through `RunProfiler` and saw a large speedup in `stripChars`, costing ~619ns/call before and ~154ns/call after.

### DIFF
--- a/sjsonnet/src/sjsonnet/Std.scala
+++ b/sjsonnet/src/sjsonnet/Std.scala
@@ -446,14 +446,14 @@ class Std {
         try {
           val pattern = Pattern.compile(from.value, Pattern.LITERAL)
           val quotedTo = java.util.regex.Matcher.quoteReplacement(to.value)
-          (new SpecStringReplace(pattern, quotedTo), Array(str))
+          (new SpecStrReplace(pattern, quotedTo), Array(str))
         } catch {
           case _: Exception => null
         }
       case _ => null
     }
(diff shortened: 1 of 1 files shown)


JoshRosen:
I considered specializing `strReplace` but doing so doesn't help (and is actually slower) on JDK 9+ because https://bugs.openjdk.org/browse/JDK-8058779 optimized `java.lang.String.replace(CharSequence, CharSequence)`.

stephenamar-db (approved):

PR #244, opened 2024-12-19
Add RegEx supports using RE2 to sjsonnet

stephenamar-db:
With this PR, I'm adding a handful of methods to expose regular expressions in jsonnet, through std.native()
I'm modeling them after [this open PR from jsonnet](https://github.com/google/jsonnet/pull/1039), which was ported to jrsonnet.

For now, they are in std.native() as they are not part of the default std package and use RE2 instead of the native regexp package (for performance and compatibility reasons with a future go-jsonnet implementation).

- regexFullMatch(pattern, str) -- Full match regex
- regexPartialMatch(pattern, str) -- Partial match regex
- regexReplace(str, pattern, to) -- Replace single occurance using regex
- regexGlobalReplace(str, pattern, to) -- Replace globally using regex

and the utility function:
- regexQuoteMeta(str) -- Escape regex metachararacters

Those functions return a object:
```
std.native("regexFullMatch")("h(?P<mid>.*)o", "hello")

{
   "captures": [
      "ell"
   ],
   "string": "hello"
}
```

This PR does not add support for the "namedCaptures" return field due to some complications with scalajs and scalanative. Those language both use the JDK Pattern class (js being powered by ECMA regex and Native being powered by RE2(!)), but JDK<20 Pattern class does not have a straightforward way to list the names of groups without some additional hacks. This will be dealt with in a follow up PR.

This PR also adds the ability to cache patterns, and refactors all users of regexes to use it.

### DIFF
--- a/sjsonnet/src-js/sjsonnet/Platform.scala
+++ b/sjsonnet/src-js/sjsonnet/Platform.scala
@@ -40,9 +40,7 @@ object Platform {
     throw new Exception("hashFile not implemented in Scala.js")
   }
 
-  private val regexCache = new util.LinkedHashMap[String, Pattern](100, 0.75f, true) {
-    override def removeEldestEntry(eldest: util.Map.Entry[String, Pattern]): Boolean = size() > 100
-  }
+  private val regexCache = new util.concurrent.ConcurrentHashMap[String, Pattern]
   def getPatternFromCache(pat: String) : Pattern = regexCache.computeIfAbsent(pat, _ => Pattern.compile(pat))
 
   def regexQuote(s: String): String = Pattern.quote(s)
--- a/sjsonnet/src-js/sjsonnet/Platform.scala
+++ b/sjsonnet/src-js/sjsonnet/Platform.scala
@@ -1,5 +1,10 @@
 package sjsonnet
+
 import java.io.File
+import java.util
+import java.util.regex.Pattern
+
+
 object Platform {
   def gzipBytes(s: Array[Byte]): String = {
     throw new Exception("GZip not implemented in Scala.js")
@@ -34,4 +39,11 @@ object Platform {
   def hashFile(file: File): String = {
     throw new Exception("hashFile not implemented in Scala.js")
   }
+
+  private val regexCache = new util.LinkedHashMap[String, Pattern](100, 0.75f, true) {
+    override def removeEldestEntry(eldest: util.Map.Entry[String, Pattern]): Boolean = size() > 100
+  }
+  def getPatternFromCache(pat: String) : Pattern = regexCache.computeIfAbsent(pat, _ => Pattern.compile(pat))
+
+  def regexQuote(s: String): String = Pattern.quote(s)
 }
--- a/sjsonnet/src-js/sjsonnet/Platform.scala
+++ b/sjsonnet/src-js/sjsonnet/Platform.scala
@@ -41,6 +41,9 @@ object Platform {
   }
 
   private val regexCache = new util.concurrent.ConcurrentHashMap[String, Pattern]
+
+  // scala.js does not rely on re2. Per https://www.scala-js.org/doc/regular-expressions.html.
+  // Expect to see some differences in behavior.
   def getPatternFromCache(pat: String) : Pattern = regexCache.computeIfAbsent(pat, _ => Pattern.compile(pat))
 
   def regexQuote(s: String): String = Pattern.quote(s)
--- a/sjsonnet/src-jvm/sjsonnet/Platform.scala
+++ b/sjsonnet/src-jvm/sjsonnet/Platform.scala
@@ -1,14 +1,17 @@
 package sjsonnet
 
 import java.io.{BufferedInputStream, ByteArrayOutputStream, File, FileInputStream}
+import java.util
 import java.util.Base64
 import java.util.zip.GZIPOutputStream
+import com.google.re2j.Pattern
 import net.jpountz.xxhash.{StreamingXXHash64, XXHashFactory}
 import org.json.{JSONArray, JSONObject}
 import org.tukaani.xz.LZMA2Options
 import org.tukaani.xz.XZOutputStream
 import org.yaml.snakeyaml.{LoaderOptions, Yaml}
 import org.yaml.snakeyaml.constructor.SafeConstructor
+
 import scala.jdk.CollectionConverters._
 
 object Platform {
@@ -107,4 +110,11 @@ object Platform {
 
     hash.getValue.toString
   }
+
+  private val regexCache = new util.LinkedHashMap[String, Pattern](100, 0.75f, true) {
+    override def removeEldestEntry(eldest: util.Map.Entry[String, Pattern]): Boolean = size() > 100
+  }
+  def getPatternFromCache(pat: String) : Pattern = regexCache.computeIfAbsent(pat, _ => Pattern.compile(pat))
+
+  def regexQuote(s: String): String = Pattern.quote(s)
 }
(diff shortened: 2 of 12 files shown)


sjsonnet/src-jvm/sjsonnet/Platform.scala
```diff
@@ -107,4 +110,11 @@ object Platform {
 
     hash.getValue.toString
   }
+
+  private val regexCache = new util.LinkedHashMap[String, Pattern](100, 0.75f, true) {
```
He-Pin, line 114: should be a concurrenthashmap i think

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -474,26 +472,24 @@ class Std(private val additionalNativeFunctions: Map[String, Val.Builtin] = Map.
...
-    private def getTrailingPattern(chars: String): Pattern =
-      Pattern.compile("[" + Regex.quote(chars) + "]+$")
+    private def getTrailingPattern(chars: String): String = "[" + Platform.regexQuote(chars) + "]+$"
 
     def unspecializedStrip(str: String, chars: String, left: Boolean, right: Boolean): String = {
       var s = str
-      if (right) s = getTrailingPattern(chars).matcher(s).replaceAll("")
-      if (left) s = getLeadingPattern(chars).matcher(s).replaceAll("")
+      if (right) s = Platform.getPatternFromCache(getTrailingPattern(chars)).matcher(s).replaceAll("")
+      if (left) s = Platform.getPatternFromCache(getLeadingPattern(chars)).matcher(s).replaceAll("")
```
He-Pin, line 492: now it performent better.

He-Pin:
the reg2j's performance is better?

stephenamar-db:
> the reg2j's performance is better?

```
RE2 is a regular expression engine that runs in time linear in the size of the input. RE2/J is a port of C++ library [RE2](https://github.com/google/re2) to pure Java.
```

It's also what's used in scala native.
https://scala-native.org/en/latest/lib/javalib.html#regular-expressions-java-util-regex

He-Pin:
Seems like we could have a scala port from it ,t hen the scala-js and scala-native can use it too.

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -508,8 +504,8 @@ class Std(private val additionalNativeFunctions: Map[String, Val.Builtin] = Map.
 
       def evalRhs(str: Val, ev: EvalScope, pos: Position): Val = {
         var s = str.asString
-        if (right) s = rightPattern.matcher(s).replaceAll("")
-        if (left) s = leftPattern.matcher(s).replaceAll("")
+        if (right) s = Platform.getPatternFromCache(rightPattern).matcher(s).replaceAll("")
+        if (left) s = Platform.getPatternFromCache(leftPattern).matcher(s).replaceAll("")
```
JoshRosen, line 508: Note that this is losing some of the benefits of the old specialization: I think it would be preferable to do a `Pattern.compile` on `leftPattern` and `rightPattern` above so that we can avoid the cache lookup cost.

stephenamar-db: Good point

He-Pin:
Is there a test suit for regex?, maybe this will change some behavior or result, eg with default behavior changed.

stephenamar-db:
> Is there a test suit for regex?, maybe this will change some behavior or result, eg with default behavior changed.

I added a test suite. Is there something else you are thinking about?

sjsonnet/src-js/sjsonnet/Platform.scala
```diff
@@ -1,5 +1,10 @@
 package sjsonnet
+
 import java.io.File
+import java.util
+import java.util.regex.Pattern
```
JoshRosen, line 5: I suppose this means that the `scala-js` implementation won't exactly match RE2 syntax and instead will work according to https://www.scala-js.org/doc/regular-expressions.html . Possibly a corner-case worthy of mention in case we end up writing docs for these functions?

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -474,26 +472,24 @@ class Std(private val additionalNativeFunctions: Map[String, Val.Builtin] = Map.
...
       case Array(str, from: Val.Str, to) =>
-        try { (new SpecFrom(Pattern.compile(from.value)), Array(str, to)) } catch { case _: Exception => null }
+        try { (new SpecFrom(from.value), Array(str, to)) } catch { case _: Exception => null }
       case _ => null
     }
-    private class SpecFrom(from: Pattern) extends Val.Builtin2("strReplaceAll", "str", "to") {
+    private class SpecFrom(from: String) extends Val.Builtin2("strReplaceAll", "str", "to") {
       def evalRhs(str: Val, to: Val, ev: EvalScope, pos: Position): Val =
-        Val.Str(pos, from.matcher(str.asString).replaceAll(to.asString))
+        Val.Str(pos, Platform.getPatternFromCache(from).matcher(str.asString).replaceAll(to.asString))
```
JoshRosen, line 480: This is still defeating some of the point of the specialization: in these specialized ones we should be doing the pattern retrieval at the `def specialize` call and be using it in the body.

sjsonnet/src/sjsonnet/StdRegex.scala
```diff
@@ -0,0 +1,89 @@
...
+      }
+    },
+    "regexGlobalReplace" -> new Val.Builtin3("regexGlobalReplace", "str", "pattern", "to") {
+      override def evalRhs(str: Val, pattern: Val, to: Val, ev: EvalScope, pos: Position): Val = {
+        val compiledPattern = Platform.getPatternFromCache(pattern.asString)
+        val matcher = compiledPattern.matcher(str.asString)
+        Val.Str(pos.noOffset, matcher.replaceAll(to.asString))
+      }
+    },
+    "regexReplace" -> new Val.Builtin3("regexGlobalReplace", "str", "pattern", "to") {
```
JoshRosen, line 76: Typo:

```suggestion
    "regexReplace" -> new Val.Builtin3("regexReplace", "str", "pattern", "to") {
```

JoshRosen (approved):
Looks good overall, just minor comments on a typo and some suggestions to save on cache lookups in a couple of hot paths.

sjsonnet/src/sjsonnet/Std.scala
stephenamar-db: Not sure I fully understand what the benefit of building the pattern in the constructor is.
in all cases, I'm trying to avoid explicitly importing Pattern in Std (due to the different implementations - which sadly don't even share an explicit common interface)

What if I do:
```
    private class SpecFrom(from: String) extends Val.Builtin2("strReplaceAll", "str", "to") {
      private val fromPattern = Platform.getPatternFromCache(from)
      def evalRhs(str: Val, to: Val, ev: EvalScope, pos: Position): Val =
        Val.Str(pos, fromPattern.matcher(str.asString).replaceAll(to.asString))
    }
```

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -48,7 +46,7 @@ class Std(private val additionalNativeFunctions: Map[String, Val.Builtin] = Map.
     },
   )
   require(oldNativeFunctions.forall(k => !additionalNativeFunctions.contains(k._1)), "Conflicting native functions")
-  private val nativeFunctions = oldNativeFunctions ++ additionalNativeFunctions
+  private val nativeFunctions = oldNativeFunctions ++ additionalNativeFunctions ++ StdRegex.functions
```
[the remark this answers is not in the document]
stephenamar-db: yes. It's just the standard interface for people with custom mains

sjsonnet/src/sjsonnet/Std.scala
JoshRosen: Ah, I think that works. You could also make it `private[this]` so it's just a regular private field, too.

PR #251, opened 2025-01-01
Add importbin to sjsonnet

stephenamar-db:
Per spec https://jsonnet.org/ref/spec.html#:~:text=UTF%2D8.%20For-,importbin,-%2C%20the%20file%20is

importbin imports the file as an array of ints from 0-255


JoshRosen (approved):

sjsonnet/src-jvm-native/sjsonnet/CachedResolvedFile.scala
```diff
@@ -39,15 +40,17 @@ class CachedResolvedFile(val resolvedImportPath: OsPath, memoryLimitBytes: Long,
     new String(Files.readAllBytes(jFile.toPath), StandardCharsets.UTF_8);
   }
 
+  private[this] def readRawBytes(jFile: File): Array[Byte] = Files.readAllBytes(jFile.toPath)
```
He-Pin, line 43: Should be safer to check the file length first. Eg , we will not have a 2 gb sjsonnet file

sjsonnet/src-jvm-native/sjsonnet/CachedResolvedFile.scala
```diff
@@ -39,15 +40,17 @@ class CachedResolvedFile(val resolvedImportPath: OsPath, memoryLimitBytes: Long,
...
+
   /**
    * A method that will return a reader for the resolved import. If the import is too large, then this will return
    * a reader that will read the file from disk. Otherwise, it will return a reader that reads from memory.
    */
-  def getParserInput(): ParserInput = {
+  def getParserInput: ParserInput = {
```
He-Pin, line 49: This should keep the (), as it reads from a file

sjsonnet/src-jvm-native/sjsonnet/CachedResolvedFile.scala
stephenamar-db: We do check for file size in `resolvedImportContent`

PR #256, opened 2025-01-03
Multiple bug fixes and enhancements for std regex

stephenamar-db:
Implement namedGroups (which was missing from the first implementation)
- JS is tricky because you need to convert Python-style named groups (the only ones supported by RE2) to perl ones
- You need to manually find them due to the lack of clear API in the JDK

Align our implementation with jrsonnet implementations:
- return null if no match
- make fullmatch just partialmatch with anchors - it also allows us to share more code.

Fix the - problem with regexQuote in RE2 across the board


stephenamar-db:
@JoshRosen  at your convenience

sjsonnet/src/sjsonnet/StdRegex.scala
```diff
@@ -4,66 +4,58 @@ import sjsonnet.Expr.Member.Visibility
...
 object StdRegex {
+  def regexPartialMatch(pos: Position, pattern: String, str: String): Val = {
+    val compiledPattern = Platform.getPatternFromCache(pattern)
+    val matcher = compiledPattern.matcher(str)
+    var returnStr: Val = null
+    val groupCount = matcher.groupCount()
+    val captures = Array.newBuilder[Val]
+    captures.sizeHint(groupCount)
+
+    while (matcher.find()) {
```
JoshRosen, line 15: On closer look, I'm not actually sure that we want a `while()` loop here: I think we only intend to find the _first_ match, not all? Let me try adding a test case to see if I can demonstrate a bug from the loop behavior.

JoshRosen: I think our unit test cases are wrong:

```
{
    res: std.regexPartialMatch("a(b)c", "cabc")
}
```

in jrsonnet master built with the experimental regex support enabled yields:

```
{
   "res": {
      "captures": [
         "b"
      ],
      "namedCaptures": { },
      "string": "abc"
   }
}
```

sjsonnet/src/sjsonnet/StdRegex.scala
```diff
@@ -4,66 +4,58 @@ import sjsonnet.Expr.Member.Visibility
...
+      if (returnStr == null) {
+        val m = matcher.group(0)
+        if (m != null) {
+          returnStr = Val.Str(pos.noOffset, matcher.group(0))
+        } else {
+          returnStr = Val.Str(pos.noOffset, "")
+        }
+      }
```
JoshRosen, line 23: Once we change the `while` to an `if`, we can add an `else` and greatly simplify this entire block. For example, we can move the allocation of the captures builder into the `if` and avoid the allocation in case of no match.

sjsonnet/src/sjsonnet/StdRegex.scala
JoshRosen: We might also want to pull in all of the test cases from `google/jsonnet` PR, too, or at least make sure that we have exact parity for their edge cases.

PR #257, opened 2025-01-04
Add support for tailstrict

### LINKED ISSUE #189: sjsonnet does not support recursive functions/tailstrict
Recursive function test is commented out: https://github.com/databricks/sjsonnet/blob/master/sjsonnet/test/src-jvm-native/sjsonnet/FileTests.scala#L56

When attempting to compile a jsonnet file with `tailstrict`:
```
sjsonnet.ParseError: Expected ";":30:23, found "tailstrict"
    at .(../sjsonnet/sjsonnet/test/resources/test_suite/recursive_function.jsonnet:30:23)
```

stephenamar-db:
With this PR, I'm implementing tailstrict in sjsonnet.
I'm following the guidance from https://github.com/google/jsonnet/issues/343#issuecomment-325987638:

Quote:
1. If you call a function with tailstrict annotation on the apply AST, e.g. foo(42) tailstrict then the evaluation of the function body happens in "tail strict mode". The annotated AST need not actually be a tail call. Also, the arguments of this function are forced.
2. When another function call (can be a completely different function to the one that was originally called) is made when we're evaluating in "tail strict mode" and it is a tail call, then the current stack frame is re-used for the next call, rather than being pushed on top of.
3. Note that in order to preserve the tail strict mode into the new function, the new call AST has to be tailstrict as well.

Resolves #189.

### DIFF
--- a/sjsonnet/src/sjsonnet/Evaluator.scala
+++ b/sjsonnet/src/sjsonnet/Evaluator.scala
@@ -4,6 +4,7 @@ import Expr.{Error => _, _}
 import sjsonnet.Expr.Member.Visibility
 import ujson.Value
 
+import scala.annotation.tailrec
 import scala.collection.mutable
 
 /**
@@ -27,11 +28,7 @@ class Evaluator(resolver: CachedResolver,
 
   def materialize(v: Val): Value = Materializer.apply(v)
   val cachedImports = collection.mutable.HashMap.empty[Path, Val]
-
-  var isInTailstrictMode = false
-
-  override def tailstrict: Boolean = isInTailstrictMode
-
+  var tailstrict: Boolean = false
 
   override def visitExpr(e: Expr)(implicit scope: ValScope): Val = try {
     e match {
@@ -190,12 +187,12 @@ class Evaluator(resolver: CachedResolver,
   private def visitApply(e: Apply)(implicit scope: ValScope) = {
     val lhs = visitExpr(e.value)
 
-    if (isInTailstrictMode) {
+    if (tailstrict) {
       lhs.cast[Val.Func].apply(e.args.map(visitExpr(_)), e.namedNames, e.pos)
     } else if (e.tailstrict) {
-      isInTailstrictMode = true
+      tailstrict = true
       val res = lhs.cast[Val.Func].apply(e.args.map(visitExpr(_)), e.namedNames, e.pos)
-      isInTailstrictMode = false
+      tailstrict = false
       res
     } else {
       val args = e.args
@@ -211,17 +208,24 @@ class Evaluator(resolver: CachedResolver,
 
   private def visitApply0(e: Apply0)(implicit scope: ValScope): Val = {
     val lhs = visitExpr(e.value)
-    lhs.cast[Val.Func].apply0(e.pos)
+    if (e.tailstrict) {
+      tailstrict = true
+      val res = lhs.cast[Val.Func].apply0(e.pos)
+      tailstrict = false
+      res
+    } else {
+      lhs.cast[Val.Func].apply0(e.pos)
+    }
   }
 
   private def visitApply1(e: Apply1)(implicit scope: ValScope): Val = {
     val lhs = visitExpr(e.value)
-    if (isInTailstrictMode) {
+    if (tailstrict) {
       lhs.cast[Val.Func].apply1(visitExpr(e.a1), e.pos)
     } else if (e.tailstrict) {
-      isInTailstrictMode = true
+      tailstrict = true
       val res = lhs.cast[Val.Func].apply1(visitExpr(e.a1), e.pos)
-      isInTailstrictMode = false
+      tailstrict = false
       res
     } else {
       val l1 = visitAsLazy(e.a1)
@@ -231,12 +235,12 @@ class Evaluator(resolver: CachedResolver,
 
   private def visitApply2(e: Apply2)(implicit scope: ValScope): Val = {
     val lhs = visitExpr(e.value)
-    if (isInTailstrictMode) {
+    if (tailstrict) {
       lhs.cast[Val.Func].apply2(visitExpr(e.a1), visitExpr(e.a2), e.pos)
     } else if (e.tailstrict) {
-      isInTailstrictMode = true
+      tailstrict = true
       val res = lhs.cast[Val.Func].apply2(visitExpr(e.a1), visitExpr(e.a2), e.pos)
-      isInTailstrictMode = false
+      tailstrict = false
       res
     } else {
       val l1 = visitAsLazy(e.a1)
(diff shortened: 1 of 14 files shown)


sjsonnet/src/sjsonnet/Evaluator.scala
```diff
@@ -28,7 +28,12 @@ class Evaluator(resolver: CachedResolver,
   def materialize(v: Val): Value = Materializer.apply(v)
   val cachedImports = collection.mutable.HashMap.empty[Path, Val]
 
-  def visitExpr(e: Expr)(implicit scope: ValScope): Val = try {
+  var isInTailstrictMode = false
```
JoshRosen, line 31: What is the thread-safety model here?

By having this be a mutable property of the evaluator itself I think we are foregoing safe multi-threaded usage of `Evaluator`.

A few lines up, I see that we have a non-thread-safe mutable HashMap for `cachedImports`, so it's possible that this class is de-facto pre-existingly not threadsafe.

Does this match your understanding? And is our current usage in line with this?

I raise this because I've been considering implementing certain performance optimizations which strongly rely on single-threaded evaluation (such as same-thread recycling of certain temporary structures used in string materialization) and it would be convenient to be able to make strong single-threadedness assumptions about evaluator.

sjsonnet/src/sjsonnet/Evaluator.scala
```diff
@@ -184,14 +189,24 @@ class Evaluator(resolver: CachedResolver,
...
+      while (idx < args.length) {
+        argsL(idx) = visitAsLazy(args(idx))
+        idx += 1
+      }
+      lhs.cast[Val.Func].apply(argsL, e.namedNames, e.pos)
     }
-    lhs.cast[Val.Func].apply(argsL, e.namedNames, e.pos)
   }
 
   private def visitApply0(e: Apply0)(implicit scope: ValScope): Val = {
```
JoshRosen, line 212: I think we might need handling for tailstrict here too? Despite the fact that a zero-argument function cannot itself be tail recursive, don't we still need to propagate the `isInTailstrictMode` flag for use in child evaluations?

sjsonnet/src/sjsonnet/Expr.scala
```diff
@@ -128,11 +128,11 @@ object Expr{
...
-  case class Apply0(pos: Position, value: Expr) extends Expr
-  case class Apply1(pos: Position, value: Expr, a1: Expr) extends Expr
-  case class Apply2(pos: Position, value: Expr, a1: Expr, a2: Expr) extends Expr
-  case class Apply3(pos: Position, value: Expr, a1: Expr, a2: Expr, a3: Expr) extends Expr
+  case class Apply(pos: Position, value: Expr, args: Array[Expr], namedNames: Array[String], tailstrict: Boolean) extends Expr
+  case class Apply0(pos: Position, value: Expr, tailstrict: Boolean) extends Expr
+  case class Apply1(pos: Position, value: Expr, a1: Expr, tailstrict: Boolean) extends Expr
+  case class Apply2(pos: Position, value: Expr, a1: Expr, a2: Expr, tailstrict: Boolean) extends Expr
+  case class Apply3(pos: Position, value: Expr, a1: Expr, a2: Expr, a3: Expr, tailstrict: Boolean) extends Expr
   case class ApplyBuiltin(pos: Position, func: Val.Builtin, argExprs: Array[Expr]) extends Expr {
```
JoshRosen, line 136: I'm curious about the interaction between tail strict mode, built-in function application, and potential lazy evaluation in built-ins.

For example, it is possible to have a `tailstrict` user-defined function which invokes a built-in function which returns lazy values which in turn reference user defined functions that could potentially be affected by tail strict mode? i.e. do we need "memory" of whether tail strict mode was in effect at the time of creation of a lazy value?

sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -377,11 +377,14 @@ object Val{
 
     override def asFunc: Func = this
 
-    def apply(argsL: Array[_ <: Lazy], namedNames: Array[String], outerPos: Position)(implicit ev: EvalScope): Val = {
+    def apply(argsL: Array[_ <: Lazy], namedNames: Array[String], outerPos: Position)(implicit ev: EvalScope, vs: ValScope = defSiteValScope): Val = {
```
JoshRosen, line 380: I'm a little confused about the semantics of having an implicit parameter which has a default value. Where do we override this?

JoshRosen: Courtesy of ChatGPT, an example demonstrating that implicits in the caller scope take precedence over the default parameter value:

```scala
object DefaultVsImplicitExample extends App {

  // A method with an implicit parameter that has a default value
  def greet(implicit name: String = "DefaultName"): Unit =
    println(s"Hello, $name!")

  // No implicit in scope, so it uses the default value
  greet // prints: "Hello, DefaultName!"

  // Introduce an implicit value in this scope
  {
    implicit val localName: String = "Alice"
    // Now greet() finds the implicit "Alice" instead of using the default
    greet // prints: "Hello, Alice!"
  }

  // Outside that block, the implicit is no longer visible, so greet() falls back to the default
  greet // prints: "Hello, DefaultName!"
}
```

---

Do we ever actually use the default value of `vs` here? i.e. do we ever mutate `defSiteValScope`?

I tried dropping the defaults and then ran into compilation problems in built-in functions, e.g. in `uniqArr` where we have calls like

```scala
val o1Key = keyFFunc.apply1(v, pos.noOffset)(ev)
```

which won't apply without a `vs`. But if this is indeed tail rec mode, don't we want to be using the caller's scope?

I find this pretty confusing overall and I'd like to get a better handle on how this works, esp. to make sure that we're not somehow having unexpected interactions between tail recursive functions and built in functions.

sjsonnet/src/sjsonnet/Evaluator.scala
stephenamar-db: Evaluator is definitely not thread safe - it has never been. It walks the AST step by step.

sjsonnet/src/sjsonnet/Expr.scala
stephenamar-db: Maybe? I'm not sure to be honest. I'm trying to find an example where this would cause an issue.
As far as I can see, arguments to builtins are always forced (see Builtin classes in Val)

sjsonnet/src/sjsonnet/Val.scala
stephenamar-db: I removed this whole thing

sjsonnet/src/sjsonnet/Expr.scala
JoshRosen: I think my concern here has substantially been addressed by the latest round of changes: before, I had some open concerns about how stack frames were being handled in a seemingly-different manner in tailstrict evaluation mode related to the questions at https://github.com/databricks/sjsonnet/pull/257#discussion_r1907734066, but it looks like that's now been updated and that distinction doesn't exist in the same form AFAICT.

PR #265, opened 2025-01-24
Make std.get handle default lazily

stephenamar-db:
Basically, 

```
std.get({a: 1}, "a", default=error "a")
```

should not execute default unless needed.
In the current implementation, builtins evaluates all arguments by default. This change changes that. Now Builtins have to force arguments when they need it.

### DIFF
--- a/sjsonnet/src/sjsonnet/Std.scala
+++ b/sjsonnet/src/sjsonnet/Std.scala
@@ -49,42 +49,42 @@ class Std(private val additionalNativeFunctions: Map[String, Val.Builtin] = Map.
   private val nativeFunctions = builtinNativeFunctions ++ additionalNativeFunctions
 
   private object AssertEqual extends Val.Builtin2("assertEqual", "a", "b") {
-    def evalRhs(v1: Val, v2: Val, ev: EvalScope, pos: Position): Val = {
-      val x1 = Materializer(v1)(ev)
-      val x2 = Materializer(v2)(ev)
+    def evalRhs(v1: Lazy, v2: Lazy, ev: EvalScope, pos: Position): Val = {
+      val x1 = Materializer(v1.force)(ev)
+      val x2 = Materializer(v2.force)(ev)
       if (x1 == x2) Val.True(pos)
       else Error.fail("assertEqual failed: " + x1 + " != " + x2)
     }
   }
 
   private object ToString extends Val.Builtin1("toString", "a") {
-    def evalRhs(v1: Val, ev: EvalScope, pos: Position): Val = Val.Str(pos, v1 match {
+    def evalRhs(v1: Lazy, ev: EvalScope, pos: Position): Val = Val.Str(pos, v1 match {
       case Val.Str(_, s) => s
-      case v => Materializer.stringify(v)(ev)
+      case v => Materializer.stringify(v.force)(ev)
     })
   }
 
   private object Length extends Val.Builtin1("length", "x") {
-    def evalRhs(x: Val, ev: EvalScope, pos: Position): Val =
-      Val.Num(pos, x match {
+    def evalRhs(x: Lazy, ev: EvalScope, pos: Position): Val =
+      Val.Num(pos, x.force match {
         case Val.Str(_, s) => s.length
         case a: Val.Arr => a.length
         case o: Val.Obj => o.visibleKeyNames.length
         case o: Val.Func => o.params.names.length
         case x => Error.fail("Cannot get length of " + x.prettyName)
       })
-    override def specialize(args: Array[Expr]) = args match {
-      case Array(Expr.ApplyBuiltin2(_, Filter, f, a)) => (CountF, Array(f, a))
+    override def specialize(args: Array[Expr], tailstrict: Boolean) = args match {
+      case Array(Expr.ApplyBuiltin2(_, Filter, f, a, tailstrict)) => (CountF, Array(f, a))
       case _ => null
     }
   }
 
   private object CountF extends Val.Builtin2("length", "func", "arr") {
-    def evalRhs(_func: Val, arr: Val, ev: EvalScope, pos: Position): Val = {
+    def evalRhs(_func: Lazy, arr: Lazy, ev: EvalScope, pos: Position): Val = {
       val p = pos.noOffset
-      val a = arr.asArr.asLazyArray
+      val a = arr.force.asArr.asLazyArray
       var i = 0
-      val func = _func.asFunc
+      val func = _func.force.asFunc
       var res = 0
       if(func.isInstanceOf[Val.Builtin] || func.params.names.length != 1) {
         while(i < a.length) {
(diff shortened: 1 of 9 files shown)


jordanmoxon-db (approved):

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -213,113 +213,113 @@ class Std(private val additionalNativeFunctions: Map[String, Val.Builtin] = Map.
...
 
     }
   }
 
   private object Foldr extends Val.Builtin3("foldr", "func", "arr", "init") {
-    def evalRhs(_func: Val, arr: Val, init: Val, ev: EvalScope, pos: Position): Val = {
-      val func = _func.asFunc
+    def evalRhs(_func: Lazy, arr: Lazy, init: Lazy, ev: EvalScope, pos: Position): Val = {
+      val func = _func.force.asFunc
       arr match {
```
JoshRosen, line 265: Should this be `arr.force` here? I see that you have the `arr.force` below in the log line instead, but I think that should be swapped for symmetry with other usage.

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -1372,7 +1372,7 @@ class Std(private val additionalNativeFunctions: Map[String, Val.Builtin] = Map.
       v match{
```
JoshRosen, line 1372: I think we need `v.force` here?

JoshRosen: Oh, hmm, I guess that this shorthand `builtin` syntax is still continuing to force the input arguments.

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -213,113 +213,113 @@ class Std(private val additionalNativeFunctions: Map[String, Val.Builtin] = Map.
...
+    def evalRhs(_func: Lazy, arr: Lazy, ev: EvalScope, pos: Position): Val = {
       val p = pos.noOffset
-      val a = arr.asArr.asLazyArray
+      val a = arr.force.asArr.asLazyArray
       var i = 0
-      val func = _func.asFunc
+      val func = _func.force.asFunc
       if(func.isInstanceOf[Val.Builtin] || func.params.names.length != 1) {
         while(i < a.length) {
           if(!func.apply1(a(i), p)(ev).isInstanceOf[Val.True]) {
```
JoshRosen, line 325: This is possibly a pre-existing bug, but I'm wondering whether we should be calling `.force` on the result of function application throughout this method.

I spotted this while doing a systematic review of places where we do `isInstanceOf` or `match` checks.

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -788,39 +788,39 @@ class Std(private val additionalNativeFunctions: Map[String, Val.Builtin] = Map.
...
 
   private object ParseYaml extends Val.Builtin1("parseYaml", "str") {
-    def evalRhs(str: Val, ev: EvalScope, pos: Position): Val = {
+    def evalRhs(str: Lazy, ev: EvalScope, pos: Position): Val = {
       try {
-        ujson.StringParser.transform(Platform.yamlToJson(str.asString), new ValVisitor(pos))
+        ujson.StringParser.transform(Platform.yamlToJson(str.force.asString), new ValVisitor(pos))
       } catch {
         case _: Exception => null
       }
```
JoshRosen, line 826: Pre-existing bug in nearby code (below the fold):

```
private object ManifestTomlEx extends Val.Builtin2("manifestTomlEx", "value", "indent") {
    private def isTableArray(v: Val) = v match {
      case s: Val.Arr => s.length > 0 && s.asLazyArray.forall(_.isInstanceOf[Val.Obj])
      case _ => false
    }
```

I think we need a `_.force.isInstanceOf` instead of what we have now, i.e. add the force.

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -879,16 +879,16 @@ class Std(private val additionalNativeFunctions: Map[String, Val.Builtin] = Map.
...
     }
   }
 
   private object Set_ extends Val.Builtin2("set", "arr", "keyF", Array(null, Val.False(dummyPos))) {
-    def evalRhs(arr: Val, keyF: Val, ev: EvalScope, pos: Position): Val = {
-      uniqArr(pos, ev, sortArr(pos, ev, arr, keyF), keyF)
+    def evalRhs(arr: Lazy, keyF: Lazy, ev: EvalScope, pos: Position): Val = {
+      uniqArr(pos, ev, sortArr(pos, ev, arr.force, keyF.force), keyF.force)
     }
   }
```
JoshRosen, line 894: Bug in nearby code below the fold:

https://github.com/databricks/sjsonnet/blob/877f5aa9659dc51e277e6e9c62e94cfd137dd38e/sjsonnet/src/sjsonnet/Std.scala#L927-L959

In mergePatch I think we want to do `force` on the members before the isInstanceOf checks, e.g. force after we call `.valueRaw`? Unless there's some invariant whereby we can never get lazy values there.

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -49,42 +49,42 @@ class Std(private val additionalNativeFunctions: Map[String, Val.Builtin] = Map.
...
+    def evalRhs(_func: Lazy, arr: Lazy, ev: EvalScope, pos: Position): Val = {
       val p = pos.noOffset
-      val a = arr.asArr.asLazyArray
+      val a = arr.force.asArr.asLazyArray
       var i = 0
-      val func = _func.asFunc
+      val func = _func.force.asFunc
       var res = 0
       if(func.isInstanceOf[Val.Builtin] || func.params.names.length != 1) {
         while(i < a.length) {
```
JoshRosen, line 90: Similar question here as in other code: do we need to do a `force` on function application results?

sjsonnet/src/sjsonnet/Std.scala
JoshRosen: I think we also need this in `avg` and `sum`:

https://github.com/databricks/sjsonnet/blob/877f5aa9659dc51e277e6e9c62e94cfd137dd38e/sjsonnet/src/sjsonnet/Std.scala#L1530-L1544

JoshRosen: Probably not required in https://github.com/databricks/sjsonnet/blob/877f5aa9659dc51e277e6e9c62e94cfd137dd38e/sjsonnet/src/sjsonnet/Std.scala#L1588-L1590 in `primitiveEquals` as long as the shorthand `builtin` continues to force its args.

JoshRosen: Perhaps required in `uniqArr` after function application? It looks like `ev.equal` doesn't force its arguments.

https://github.com/databricks/sjsonnet/blob/877f5aa9659dc51e277e6e9c62e94cfd137dd38e/sjsonnet/src/sjsonnet/Std.scala#L1742-L1767

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -49,42 +49,42 @@ class Std(private val additionalNativeFunctions: Map[String, Val.Builtin] = Map.
...
+      val x1 = Materializer(v1.force)(ev)
+      val x2 = Materializer(v2.force)(ev)
       if (x1 == x2) Val.True(pos)
       else Error.fail("assertEqual failed: " + x1 + " != " + x2)
     }
   }
 
   private object ToString extends Val.Builtin1("toString", "a") {
-    def evalRhs(v1: Val, ev: EvalScope, pos: Position): Val = Val.Str(pos, v1 match {
+    def evalRhs(v1: Lazy, ev: EvalScope, pos: Position): Val = Val.Str(pos, v1 match {
```
JoshRosen, line 61: I think this should be `v1.force`, not `v.force` below.

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -597,25 +597,25 @@ class Std(private val additionalNativeFunctions: Map[String, Val.Builtin] = Map.
   }
 
   private object Member extends Val.Builtin2("member", "arr", "x") {
-    def evalRhs(arr: Val, x: Val, ev: EvalScope, pos: Position): Val = {
-      Val.bool(pos, arr match {
+    def evalRhs(arr: Lazy, x: Lazy, ev: EvalScope, pos: Position): Val = {
+      Val.bool(pos, arr.force match {
         case str: Val.Str =>
           val secondArg = x match {
```
JoshRosen, line 603: `x.force` here?

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -1372,7 +1372,7 @@ class Std(private val additionalNativeFunctions: Map[String, Val.Builtin] = Map.
       v match{
         case Val.Str(_, value) => Base64.getEncoder().encodeToString(value.getBytes)
         case arr: Val.Arr => Base64.getEncoder().encodeToString(arr.iterator.map(_.cast[Val.Num].value.toByte).toArray)
-        case x => Error.fail("Cannot base64 encode " + x.prettyName)
+        case x => Error.fail("Cannot base64 encode " + x.force.prettyName)
       }
     },
```
JoshRosen, line 1378: In nearby code below the fold in `prune`, I think we should be forcing the value `x` at the start of `rec` before we do the matching:

https://github.com/databricks/sjsonnet/blob/877f5aa9659dc51e277e6e9c62e94cfd137dd38e/sjsonnet/src/sjsonnet/Std.scala#L1478-L1491

We can remove the `.asStrictArray` because the forcing would then be implicitly done at the very start of `rec`, saving an unnecessary array copy.

We might also potentially be able to clean up the `.map(identity)` there, which I think was vestigial / unneeded (spotted while profiling related code recently).

JoshRosen (commented):
I took a careful look at the stdlib updates, paying special attention to whether we have `.force` calls in the right places. I specifically focused on places where we do `.isInstanceOf` or `match` on `Val`s. 

I spotted a couple of potential bugs, some of which were pre-existing issues (e.g. inconsistency in forcing the results of function application or object field retrieval prior to comparison) and others which are new in this PR.

This looks good overall pending fixes to the places noted above.

sjsonnet/src/sjsonnet/Std.scala
stephenamar-db: It does seem to be work though. I think it's because the returned values are bools, so force does nothing.

stephenamar-db: if you look at Evaluator:
```
def visitMethod(rhs: Expr, params: Params, outerPos: Position)(implicit scope: ValScope) =
    new Val.Func(outerPos, scope, params) {
      def evalRhs(vs: ValScope, es: EvalScope, fs: FileScope, pos: Position): Val = visitExpr(rhs)(vs)
      override def evalDefault(expr: Expr, vs: ValScope, es: EvalScope) = visitExpr(expr)(vs)
    }
```

you can see that return values are not lazy.

The only case where we call visitAsLazy are objects defs, array defs, comprehensions and apply arguments.

It could be an area of improvement!

sjsonnet/src/sjsonnet/Std.scala
stephenamar-db: > Bug in nearby code below the fold:
> 
> https://github.com/databricks/sjsonnet/blob/877f5aa9659dc51e277e6e9c62e94cfd137dd38e/sjsonnet/src/sjsonnet/Std.scala#L927-L959
> 
> In mergePatch I think we want to do `force` on the members before the isInstanceOf checks, e.g. force after we call `.valueRaw`? Unless there's some invariant whereby we can never get lazy values there.

Isn't there a bug here:
```
            val lValue = if (l.containsVisibleKey(key)) l.valueRaw(key, l, pos)(ev) else null
            val rValue = if (r.containsVisibleKey(key)) r.valueRaw(key, r, pos)(ev) else null
            if (rValue.isInstanceOf[Val.Null]) { // if we are not removing the key
              if (lValue != null && rValue == null) {
```

rValue is never null?

stephenamar-db: > I think we also need this in `avg` and `sum`:
> 
> https://github.com/databricks/sjsonnet/blob/877f5aa9659dc51e277e6e9c62e94cfd137dd38e/sjsonnet/src/sjsonnet/Std.scala#L1530-L1544

forall calls .force

stephenamar-db: > Perhaps required in `uniqArr` after function application? It looks like `ev.equal` doesn't force its arguments.
> 
> https://github.com/databricks/sjsonnet/blob/877f5aa9659dc51e277e6e9c62e94cfd137dd38e/sjsonnet/src/sjsonnet/Std.scala#L1742-L1767

all callers of uniqArr and sortArr call force

sjsonnet/src/sjsonnet/Std.scala
stephenamar-db: same answer

sjsonnet/src/sjsonnet/Std.scala
JoshRosen: > Isn't there a bug here:
> 
> ```
>             val lValue = if (l.containsVisibleKey(key)) l.valueRaw(key, l, pos)(ev) else null
>             val rValue = if (r.containsVisibleKey(key)) r.valueRaw(key, r, pos)(ev) else null
>             if (rValue.isInstanceOf[Val.Null]) { // if we are not removing the key
>               if (lValue != null && rValue == null) {
> ```
> 
> rValue is never null?

I think you omitted a `!` in the quoted code?

From https://github.com/databricks/sjsonnet/blob/877f5aa9659dc51e277e6e9c62e94cfd137dd38e/sjsonnet/src/sjsonnet/Std.scala#L925-L928

`rValue` could be null. An `null.isInstanceOf[Val.Null]` will be `false`, so its negation would be `true` hence a need to check for `rValue` nullness in that condition.

stephenamar-db: yeah a typo

sjsonnet/src/sjsonnet/Std.scala
JoshRosen: Could we run into trouble if we are applying a function which returns a field from an object where the field is lazy? e.g. a function which is akin to a lambda to extract a field or fields from some object?

JoshRosen: On inspection, I think this is fine as-is and now have a better understanding of why:

The specific concern in these `asInstanceOf` checks is matching on a `LazyWithComputeFunc`.

I was wondering if it was possible for a LazyWithComputeFunc returning boolean to wind up as the return value of a method, but I now think that it can't:

- We're already forcing the inputs to the user-supplied method.
- Therefore, the only way that we could return a `LazyWithComputeFunc` would be if it came from a field or method of the object or array supplied to the function.
- `visitAsLazy` won't construct a `LazyWithComputeFunc` if the input is already `True` or `False`.
- No boolean expression returns `LazyWithComputeFunc`.
- Even though `visitBindings` in object compehensions calls `new LazyWithComputeFunc(() => visitExpr(b.rhs)(scope(self, sup)))` and yields a lazy that could wrap a boolean, retrieval of bindings is done via `visitValidId` which calls `.force`.

I couldn't come with any test input to demonstrate bugs for this user-defined predicate case.

JoshRosen (approved):

PR #275, opened 2025-04-10
Make lazy not a protected class

micheltuu:
Otherwise people can't extend `sjsonnet.Val.Builtin2` since they can't implement
```
  def evalRhs(
      arg1: sjsonnet.Lazy,
      arg2: sjsonnet.Lazy,
      ev: sjsonnet.EvalScope,
      pos: sjsonnet.Position): sjsonnet.Val = {
```

See [example](https://github.com/databricks-eng/universe/pull/1015923)


micheltuu:
@stephenamar-db can you help take a look? looks like I can't assign a reviewer

PR #278, opened 2025-04-17
chore: support query extVar and tlaVar with a function.

He-Pin:
Motivation:
Currently, I'm using sjsonnet from Java, which will require me to convert to Scala's immutable map.

Modification:
Instead of converting, add a query function to look up the value.

Result:
Less overhead when using in Java


He-Pin:
Another way is using an `asImmutableMap` helper method

stephenamar-db:
This is causing a regression. I'm investigating and I may revert.

Issue #279, opened 2025-04-17
seperate the parseCache for parse jsonnet and extVars/tlaVars


He-Pin:
I think it would be better to separate, otherwise, if a user's pass in a static shared parse cache, which can then cause problem when parse the same key.

PR #280, opened 2025-04-17
chore: Avoid using the same parse cache for both jsonnet and variables

He-Pin:
Motivation:
fix: https://github.com/databricks/sjsonnet/issues/279

Otherwise, if a static parse cache is used, then the same extVar/tlaVar with a different value will be cached.

Modification:
Separate the `varParseCache`


He-Pin:
@stephenamar-db Would you like give this a look, thanks.

PR #285, opened 2025-04-18
chore: port https://github.com/google/jsonnet/pull/1218

He-Pin:
port part of https://github.com/google/jsonnet/releases/tag/v0.21.0-rc2

PR #287, opened 2025-04-18
Fix std.min/maxArray

stephenamar-db:
https://github.com/databricks/sjsonnet/pull/285 exposed a bug in the implementation


stephenamar-db:
@He-Pin ^^

PR #291, opened 2025-04-18
chore: Extract buildin* to FunctionBuilder

He-Pin:
Motivation:

Make the `buildin`* helper function accessible without extending the `Std()` class.

Then I can just extend the `FunctionBuilder`, build the `additionalNativeFunctions`, and then pass it to the `Std()`'s constructor.

otherwise: 
```scala
object JsonnetStdExt extends Std(Map.empty) {
  val extFunctions: Map[String, Val.Func] = Map(
    builtin("objectReplaceKey", "obj", "key", "newKey") {
      (pos, ev, o: Val.Obj, key: String, newKey: String) =>
        val bindings: Array[(String, Val.Obj.Member)] = for {
          k <- o.visibleKeyNames
          v = o.value(k, pos.fileScope.noOffsetPos)(ev)
        } yield {
          val newKeyName = if (k == key) newKey else k
          (newKeyName, new Val.Obj.ConstMember(false, Visibility.Normal, v))
        }
        Val.Obj.mk(pos, bindings)
    }
  )
}
```

### DIFF
--- a/sjsonnet/src/sjsonnet/FunctionBuilder.scala
+++ b/sjsonnet/src/sjsonnet/FunctionBuilder.scala
@@ -1,0 +1,73 @@
+package sjsonnet
+
+/**
+ * Function building helpers for building functions of the Jsonnet language.
+ * */
+trait FunctionBuilder {
+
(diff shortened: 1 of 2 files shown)


He-Pin:
@stephenamar-db I have rebased on the current master.

PR #294, opened 2025-04-20
chore: add some static optimization

He-Pin:
Motivation:
Add some static optimization for long run jsonnet script.


stephenamar-db:
Does it actually make a difference?

He-Pin:
our use cases is long running jsonnet script ,so It should help in some scenarios.

PR #301, opened 2025-04-22
chore: Add `evaluateVar` to interpreter.

He-Pin:
Motivation:
When using Java, I query a Java hashmap, I can then emit the `Expr` directly, instead of converting it to a string and then parsing it.

### DIFF
--- a/sjsonnet/src/sjsonnet/ExternalVariable.scala
+++ b/sjsonnet/src/sjsonnet/ExternalVariable.scala
@@ -6,20 +6,38 @@ package sjsonnet
 sealed trait ExternalVariableKind[T]
 
 object ExternalVariableKind {
+  /**
+   * Get an instance of ExternalVariableKind for a code snippet.
+   * */
   def code: ExternalVariableKind[String] = Code
 
+  /**
+   * Get an instance of ExternalVariableKind for an expr.
+   * */
   def expr: ExternalVariableKind[sjsonnet.Expr] = Expr
 
+  /**
+   * Get an instance of ExternalVariableKind for a variable.
+   * */
   def variable: ExternalVariableKind[String] = Variable
 
+  /**
+   * Indicates that the external variable is a code snippet.
+   * */
   case object Variable extends ExternalVariableKind[String] {
     override def toString: String = "variable"
   }
 
+  /**
+   * Indicates that the external variable is a string literal.
+   * */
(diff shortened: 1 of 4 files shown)


sjsonnet/src/sjsonnet/Interpreter.scala
```diff
@@ -66,14 +66,26 @@ class Interpreter(queryExtVar: String => Option[String],
                       settings: Settings, warn: Error => Unit): Evaluator =
     new Evaluator(resolver, extVars, wd, settings, warn)
 
+  /**
+   * Evaluate a variable to an `Expr`.
+   * */
+  def evaluateVar(k: String, v: Any): Expr = v match {
+    case expr: Expr => expr
+    case s: String => parseVar(k, s)
```
He-Pin, line 74: we will need to parse it if it's a `String`

He-Pin:
```go
func prepareExtVars(i *interpreter, ext vmExtMap, kind string) map[string]*cachedThunk {
	result := make(map[string]*cachedThunk)
	for name, content := range ext {
		diagnosticFile := "<" + kind + ":" + name + ">"
		switch content.kind {
		case extKindCode:
			result[name] = codeToPV(i, diagnosticFile, content.value)
		case extKindNode:
			result[name] = nodeToPV(i, diagnosticFile, content.node)
		default:
			result[name] = readyThunk(makeValueString(content.value))
		}
	}
	return result
}
```

I think it would be nice to have a kind here to avoid some parsing, will update

stephenamar-db:
please resolve

sjsonnet/src/sjsonnet/ExternalVariable.scala
```diff
@@ -0,0 +1,48 @@
...
+ * An external variable kind indicates the type of external variable.
+ * */
+sealed trait ExternalVariableKind[T]
+
+object ExternalVariableKind {
+  def code: ExternalVariableKind[String] = Code
+
+  def expr: ExternalVariableKind[sjsonnet.Expr] = Expr
+
+  def variable: ExternalVariableKind[String] = Variable
```
He-Pin, line 13: Java api

stephenamar-db (approved):

He-Pin:
fixed https://github.com/databricks/sjsonnet/issues/303 too

PR #312, opened 2025-04-25
chore: Make visit* method private.

He-Pin:
Motivation:
Make these methods private will help jvm to inline.


HollandDM:
Our company currently extends Evaluator to do some customize, so this will disable that

He-Pin:
@HollandDM oops, would you like to PR some of the optimization here? 
I want to do more optimization on this, I'm ok to close it.

HollandDM:
We're doing some customize logic, it's for our business only, it's not about optimization.

PR #321, opened 2025-04-26
Fix overly-eager evaluation of unmerged or removed target fields in std.mergePatch

### LINKED ISSUE #307: bug:  std.mergePatch evaluates object
```
std.mergePatch({ val: error 'should not error' }, {}) + { val+:: {} }
```
gives:
```
{}
```

bug sjsonnet is error.
refs: https://github.com/CertainLach/jrsonnet/issues/188

Image

```scala
java -jar sjsonnet.jar -e "std.mergePatch({ val: error 'should not error' }, {}) + { val+:: {} }"
sjsonnet.Error: should not error
    at [Error].(<exec>:1:23)
    at [std.mergePatch].(<exec>:1:15)
    at [BinaryOp +].(<exec>:1:55)
```

JoshRosen:
This PR fixes #307, an issue where sjsonnet's `std.mergePatch` is overly eager in its evaluation of target object fields in certain circumstances.

## Example reproductions of the bug

```jsonnet
(std.mergePatch({ boom: error "should not error" }, {}) + { flag: 42 }).flag
```

and

```jsonnet
(std.mergePatch({ boom: error "should not error" }, { boom: null }) + { flag: 42 }).flag
```

and

```jsonnet
(std.mergePatch({ boom: error "should error" }, { boom: 1 }) + { flag: 42 }).flag
```

all print `42` in both google/jsonnet and go-jsonnet. In all cases, the target object contains a field, `boom`, which throws an error if evaluated. 

In the first case, the `boom` field doesn't merge with any patch field and isn't required in the final output materialization and thus should not be evaluated.

In the second case, the `boom `field is removed by an explicitly `null` patch field. In this case, the original target value should not be evaluated.

In the third case, the `boom` field is overwritten by a non-erroring field. In this case, the original target value isn't needed and should not be evaluated.

In sjsonnet, all three of these inputs trigger errors.

## Root cause

This behavior difference stems from a difference in the eagerness of value computation: the google/jsonnet reference [implementation](https://github.com/google/jsonnet/blob/5b507462fe4179bbf7acf1808f9502636eac2306/stdlib/std.jsonnet#L1531C2-L1553C13)  avoids the computation of unmerged or removed target fields:

```jsonnet
 mergePatch(target, patch)::
    if std.isObject(patch) then
      local target_object =
        if std.isObject(target) then target else {};

      local target_fields =
        if std.isObject(target_object) then std.objectFields(target_object) else [];

      local null_fields = [k for k in std.objectFields(patch) if patch[k] == null];
      local both_fields = std.setUnion(target_fields, std.objectFields(patch));

      {
        [k]:
          if !std.objectHas(patch, k) then
            target_object[k]
          else if !std.objectHas(target_object, k) then
            std.mergePatch(null, patch[k]) tailstrict
          else
            std.mergePatch(target_object[k], patch[k]) tailstrict
        for k in std.setDiff(both_fields, null_fields)
      }
    else
      patch,
```

In this reference implementation, unmodified target fields are passed through via `target_object[k]` (which is lazy) and removed fields are never retrieved. In contrast, sjsonnet's implementation eagerly evaluates both the target and patch values in all cases.

Note that performing any comparison or type check of an `error` will throw the error. For example, `std.objectFields(error 'err')` errors, as does `(error 'err') == null`. Due to this, the `local null_fields = [k for k in std.objectFields(patch) if patch[k] == null];` means that error patch fields will always throw.

## This PR's fix

In this PR, I have modified sjonnet's implementation to match the reference implementation:

- When a target field is preserved as-is (i.e. without merging with a patch), use `createLazyMember` to preserve laziness.
- When a target field is deleted or is overwritten without merging (e.g. if the patch is supplying a non-Object value), simply skip evaluation of the removed or replaced target field.

I've added new test cases in `sjsonnet/test/src/sjsonnet/StdMergePatchTests.scala`

### DIFF
--- a/sjsonnet/src/sjsonnet/Std.scala
+++ b/sjsonnet/src/sjsonnet/Std.scala
@@ -927,20 +927,25 @@ class Std(private val additionalNativeFunctions: Map[String, Val.Func] = Map.emp
           var i = 0
           while (i < keys.length) {
             val key = keys(i)
-            val lValue = if (l.containsVisibleKey(key)) l.valueRaw(key, l, pos)(ev) else null
             val rValue = if (r.containsVisibleKey(key)) r.valueRaw(key, r, pos)(ev) else null
             if (!rValue.isInstanceOf[Val.Null]) { // if we are not removing the key
-              if (lValue != null && rValue == null) {
-                // Preserve the LHS/target value:
-                kvs(kvsIdx) = (key, new Val.Obj.ConstMember(false, Visibility.Normal, lValue))
-              } else if (lValue.isInstanceOf[Val.Obj] && rValue.isInstanceOf[Val.Obj]) {
-                // Recursively merge objects:
-                kvs(kvsIdx) = (key, createLazyMember(recPair(lValue, rValue)))
-              } else if (rValue != null) {
+              if (l.containsVisibleKey(key)) {
+                if (rValue == null) {
+                  // Preserve the LHS/target value:
+                  kvs(kvsIdx) = (key, createLazyMember(l.valueRaw(key, l, pos)(ev)))
+                } else {
+                  val lValue = l.valueRaw(key, l, pos)(ev)
+                  if (lValue.isInstanceOf[Val.Obj] && rValue.isInstanceOf[Val.Obj]) {
+                    // Recursively merge objects:
+                    kvs(kvsIdx) = (key, createLazyMember(recPair(lValue, rValue)))
+                  } else {
+                    // Use the RHS/patch value and recursively remove Null or hidden fields:
+                    kvs(kvsIdx) = (key, createLazyMember(recSingle(rValue)))
+                  }
+                }
+              } else {
                 // Use the RHS/patch value and recursively remove Null or hidden fields:
                 kvs(kvsIdx) = (key, createLazyMember(recSingle(rValue)))
-              } else {
-                Error.fail("std.mergePatch: This should never happen")
               }
               kvsIdx += 1
             }
--- a/sjsonnet/src/sjsonnet/Std.scala
+++ b/sjsonnet/src/sjsonnet/Std.scala
@@ -934,6 +934,8 @@ class Std(private val additionalNativeFunctions: Map[String, Val.Func] = Map.emp
                   // Preserve the LHS/target value:
                   kvs(kvsIdx) = (key, createLazyMember(l.valueRaw(key, l, pos)(ev)))
                 } else {
+                  // Below, lValue is lazy so that we can short circuit and skip its
+                  // evaluation when rValue is not an object:
                   lazy val lValue = l.valueRaw(key, l, pos)(ev)
                   if (rValue.isInstanceOf[Val.Obj] && lValue.isInstanceOf[Val.Obj]) {
                     // Recursively merge objects:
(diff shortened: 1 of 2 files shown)


JoshRosen:
Converted to draft pending one more test case that I want to add.

JoshRosen:
I've gone ahead and fixed an additional case, updating the PR description and code accordingly. It's quite likely that there's opportunities to simplify the code structure a bit, so I'm open to suggestions or feedback here.

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -927,20 +927,25 @@ class Std(private val additionalNativeFunctions: Map[String, Val.Func] = Map.emp
...
-              } else if (lValue.isInstanceOf[Val.Obj] && rValue.isInstanceOf[Val.Obj]) {
-                // Recursively merge objects:
-                kvs(kvsIdx) = (key, createLazyMember(recPair(lValue, rValue)))
-              } else if (rValue != null) {
+              if (l.containsVisibleKey(key)) {
+                if (rValue == null) {
+                  // Preserve the LHS/target value:
+                  kvs(kvsIdx) = (key, createLazyMember(l.valueRaw(key, l, pos)(ev)))
+                } else {
+                  lazy val lValue = l.valueRaw(key, l, pos)(ev)
```
He-Pin, line 937: replace the `lazy val` with a `val`

He-Pin, line 937:
```suggestion
                  val lValue = l.valueRaw(key, l, pos)(ev)
```

He-Pin:
There is a `merge.jsonnet` in the jsonnet repo, we can include that.

sjsonnet/src/sjsonnet/Std.scala
JoshRosen: This is intentionally `lazy` in order to skip evaluation in when `rValue` is not an object, avoiding duplication between the "rValue is not an object` and "rValue is an object but lValue is not an object" cases.

JoshRosen:
> There is a `merge.jsonnet` in the jsonnet repo, we can include that.

We already do: it's included at https://github.com/databricks/sjsonnet/blob/master/sjsonnet/test/resources/test_suite/merge.jsonnet, copied from the original, but doesn't contain sufficient tests for all merge behaviors.

sjsonnet/src/sjsonnet/Std.scala
He-Pin: Thanks, better with an inline comments, I was thinking that the createLazyMember is already some kind of lazy.

sjsonnet/test/src/sjsonnet/StdMergePatchTests.scala
```diff
@@ -108,5 +108,22 @@ object StdMergePatchTests extends TestSuite {
       eval("""{a:: 0} + {[a]: 1 for a in ["a"]}""") ==> ujson.Obj()
       eval("""({a:: 0} + {[a]: 1 for a in ["a"]}).a""") ==> ujson.Num(1)
     }
+
+    test("lazy evaluation of LHS members") {
+      // An erroring target field which doesn't merge with any patch fields.
+      // This shouldn't error because we should avoid eager evaluation of unmerged target fields:
+      eval("(std.mergePatch({ boom: error \"should not error\" }, {}) + { flag: 42 }).flag") ==> ujson.Num(42)
```
He-Pin, line 115: How about using `""" """` text block to avoid the `\"`?

stephenamar-db (approved):

PR #324, opened 2025-04-27
chore: avoid look up hashmap when query precedence

He-Pin:
Motivation:
Avoid the hashmap query when climb

### DIFF
--- a/sjsonnet/src/sjsonnet/Parser.scala
+++ b/sjsonnet/src/sjsonnet/Parser.scala
@@ -284,7 +284,7 @@ class Parser(val currentFile: Path,
 
   def unaryOpExpr[$: P](pos: Position, op: Char): P[Expr.UnaryOp] = P(
     expr1.map{ e =>
-      def k2 = op match{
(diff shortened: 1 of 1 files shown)


sjsonnet/src/sjsonnet/Parser.scala
```diff
@@ -298,7 +320,7 @@ class Parser(val currentFile: Path,
           case '$' => Pass(Expr.$(pos))
           case '0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9' =>
             P.current.index = pos.offset; number
-          case x if idStartChar(x) => CharsWhileIn("_a-zA-Z0-9", 0).!.flatMapX { y =>
+          case x => if(idStartChar(x)) CharsWhileIn("_a-zA-Z0-9", 0).!.flatMapX { y =>
```
He-Pin, line 323: rewrite as this, then it can generate switch*

stephenamar-db (approved):

PR #326, opened 2025-04-27
chore: optimize more cases with `===`

He-Pin:
Motivation:
Optimize even more cases, not ready, need implement the `===` method.

refs: https://jsonnet.org/ref/language.html#equivalence-and-equality


stephenamar-db:
Is this ready?

He-Pin:
@stephenamar-db no, let me close it for now, because I will need more time to finish this.

What I want to do is optimize `if cond exprA then exprA` to be an `exprA`.

PR #332, opened 2025-04-28
feat: Add ReadWriter instance for Val.Literal

He-Pin:
Motivation:

Add ReadWriter instance for Val.Literal, we need this when returns a `Val.Literal` from function.


He-Pin:
will add a test for this

stephenamar-db (approved):

PR #305, opened 2025-04-23
feat: Add array compare.

He-Pin:
Motivation:
Add array compare
refs:  https://github.com/databricks/sjsonnet/issues/304

PR #333, opened 2025-04-28
Fix formatter for Array and Obj and add ErrorTests to JS

stephenamar-db:
Per comments in https://github.com/databricks/sjsonnet/pull/305
This fixes formatting of Arr/Obj.

Also add ErrorTests to JS


stephenamar-db:
@He-Pin ^

He-Pin:
@stephenamar-db Can we turn this behind a setting? Maybe this will affect the performance too.

Issue #331, opened 2025-04-28
bug: defining functions in object comprehensions


He-Pin:
refs: https://github.com/CertainLach/jrsonnet/issues/88
https://jsonnet.org/learning/tutorial.html#computed_field_names
https://jsonnet.org/learning/tutorial.html#comprehension

```jsonnet
local funcs = {
  [a](x): x * 2
  for a in ["f1", "f2", "f3"]
};
funcs.f1(10)
```

should gives:
```
20
```
, bug sjsonnet gives:

```
scala.MatchError: Field(Position((memory), 18),Dyn(Id a@Position((memory), 19)),false,Params([x], [null]),Normal,BinaryOp *@Position((memory), 28)) (of class sjsonnet.Expr$Member$Field)
	at sjsonnet.Parser.objinside(Parser.scala:385)
	at sjsonnet.Parser.expr2(Parser.scala:308)
	at sjsonnet.Parser.expr1(Parser.scala:248)
	at sjsonnet.Parser.expr(Parser.scala:185)
	at sjsonnet.Parser.bind(Parser.scala:432)
	at sjsonnet.Parser.parse0$3$1(Parser.scala:182)
	at sjsonnet.Parser.rec$9(Parser.scala:182)
	at sjsonnet.Parser.localExpr(Parser.scala:182)
	at sjsonnet.Parser.local(Parser.scala:271)
	at sjsonnet.Parser.expr2(Parser.scala:337)
	at sjsonnet.Parser.expr1(Parser.scala:248)
	at sjsonnet.Parser.expr(Parser.scala:185)
	at sjsonnet.Parser.document(Parser.scala:468)
	at sjsonnet.CachedResolver.$anonfun$2(Importer.scala:202)
	at fastparse.SharedPackageDefs.parseInputRaw(SharedPackageDefs.scala:69)
	at fastparse.SharedPackageDefs.parseInputRaw$(SharedPackageDefs.scala:6)
	at fastparse.package$.parseInputRaw(package.scala:5)
	at fastparse.SharedPackageDefs.parse$$anonfun$1(SharedPackageDefs.scala:35)
	at fastparse.ParserInputSource$fromParserInput.parseThrough(ParserInput.scala:25)
	at fastparse.SharedPackageDefs.parse(SharedPackageDefs.scala:42)
	at fastparse.SharedPackageDefs.parse$(SharedPackageDefs.scala:6)
	at fastparse.package$.parse(package.scala:5)
[...]

PR #337, opened 2025-04-28
fix: Fix parsing obj comp

He-Pin:
Motivation: 
refs: https://github.com/databricks/sjsonnet/issues/331

It just fixes the MatchError, not fixed yet, feel free to take it from here.

### DIFF
--- a/sjsonnet/src/sjsonnet/Parser.scala
+++ b/sjsonnet/src/sjsonnet/Parser.scala
@@ -382,7 +382,7 @@ class Parser(val currentFile: Path,
       val preLocals = exprs
         .takeWhile(_.isInstanceOf[Expr.Bind])
         .map(_.asInstanceOf[Expr.Bind])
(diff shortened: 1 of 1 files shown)


sjsonnet/src/sjsonnet/Parser.scala
```diff
@@ -382,7 +382,7 @@ class Parser(val currentFile: Path,
       val preLocals = exprs
         .takeWhile(_.isInstanceOf[Expr.Bind])
         .map(_.asInstanceOf[Expr.Bind])
-      val Expr.Member.Field(offset, Expr.FieldName.Dyn(lhs), plus, null, Visibility.Normal, rhs) =
+      val Expr.Member.Field(offset, Expr.FieldName.Dyn(lhs), plus, args, Visibility.Normal, rhs) =
```
He-Pin, line 385: The args information I think should be passed to the `Expr.ObjBody.ObjComp`

Issue #330, opened 2025-04-28
bug: Allow null end parameter in std.slice / array indexing


He-Pin:
refs: https://github.com/CertainLach/jrsonnet/issues/176

```jsonnet
jsonnet -e "std.slice(std.range(1,4), 0, null, 2)"
[
   1,
   3
]
```

In sjsonnet got :

```scala
sjsonnet.Error: Wrong parameter type: expected Int, got null
    at [std.slice].(:1:10)

```

, but it works fine in official jsonnet

```json
[
  1,
  3
]
```

He-Pin:
Seems like we need something like `asOptionInt`?

He-Pin:
I will leave this to @JoshRosen @stephenamar-db

He-Pin:
@adammw @CertainLach Thank you.

PR #338, opened 2025-04-28
fix: Allow null end parameter in std.slice / array indexing,fix (#330)

He-Pin:
Motivation:

fix https://github.com/databricks/sjsonnet/issues/330


He-Pin:
```std.range(1,4)[0:null:2]``` is valid , too :(

PR #339, opened 2025-04-28
chore: Add missing visitApplyBuiltin0 for Function0.

He-Pin:
Motivation:
Add missing `ApplyBuiltin0` when the function0 is not `staticSafe`.

In the test cases, the function0 method is annotated with static safe be false to properly test the rewriting.


sjsonnet/test/src/sjsonnet/ExtFunction0NonStaticTest.scala
```diff
@@ -0,0 +1,55 @@
...
+
+object ExtFunction0NonStaticTest extends TestSuite with FunctionModule {
+  override final val name: String = "ext"
+  override final lazy val module: Val.Obj = moduleFromFunctions(extFunctions: _*)
+
+  private object SayHello extends Val.Builtin0("sayHello") {
+    override def evalRhs(ev: EvalScope, pos: Position): Val = {
+      Val.Str(pos, "Hello, world Every time!")
+    }
+    override def staticSafe: Boolean = false
```
He-Pin, line 14: this is need to actually test it.

PR #340, opened 2025-04-28
chore: Fix matching warning when visit slice.

He-Pin:
Motivation:
Add missing case in https://github.com/databricks/sjsonnet/pull/338

Otherwise it will fails to a MatchError, where it has a parameter type mismatch error message. 

The current spec will not do any implicit conversion anyway.

### DIFF
--- a/sjsonnet/src/sjsonnet/Evaluator.scala
+++ b/sjsonnet/src/sjsonnet/Evaluator.scala
@@ -371,7 +371,8 @@ class Evaluator(resolver: CachedResolver,
     def extractParam(e: Option[Expr], default:Int): Int = e match {
       case Some(expr) => visitExpr(expr) match {
         case _:Val.Null => default
-        case v: Val.Num => v.cast[Val.Num].value.toInt
+        case v: Val.Num => v.value.toInt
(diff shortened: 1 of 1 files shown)


sjsonnet/src/sjsonnet/Evaluator.scala
```diff
@@ -371,7 +371,8 @@ class Evaluator(resolver: CachedResolver,
     def extractParam(e: Option[Expr], default:Int): Int = e match {
       case Some(expr) => visitExpr(expr) match {
         case _:Val.Null => default
-        case v: Val.Num => v.cast[Val.Num].value.toInt
+        case v: Val.Num => v.value.toInt
+        case v: Val => v.cast[Val.Num].value.toInt
```
He-Pin, line 375: `std.range(1,4)[0:null:"2"]` is not valid in jsonnet too, which gives:

```jsonnet
Error: RUNTIME ERROR: Unexpected type number, expected string
	<std>:231:12-27	function <anonymous>
	example1.jsonnet:1:1-27	
	During evaluation	

```

He-Pin: Sorry for missed this.

PR #344, opened 2025-04-29
Revert "chore: add some static optimization (#294)"

stephenamar-db:
This partially reverts commit 60db9d0256b042ce70dc4eb1c50d269ed53fd771.

This is causing performance regressions

### DIFF
--- a/sjsonnet/src/sjsonnet/StaticOptimizer.scala
+++ b/sjsonnet/src/sjsonnet/StaticOptimizer.scala
@@ -40,12 +40,9 @@ class StaticOptimizer(
 
     case BinaryOp(pos, lhs, BinaryOp.OP_in, ValidSuper(_, selfIdx)) =>
       InSuper(pos, lhs, selfIdx)
-    case BinaryOp(pos, Val.Str(_, key), BinaryOp.OP_in, obj: Val.Obj) if obj.staticSafe =>
-      Val.bool(pos, obj.containsKey(key))
     case b2 @ BinaryOp(pos, lhs: Val.Str, BinaryOp.OP_%, rhs) =>
       try ApplyBuiltin1(pos, new Format.PartialApplyFmt(lhs.value), rhs, tailstrict = false)
       catch { case _: Exception => b2 }
-    //optimize booleans
     case Or(_, Val.False(_), rhs) => transform(rhs)
     case Or(pos, Val.True(_), _) => Val.True(pos)
     case Or(pos, _, Val.True(_)) => Val.True(pos)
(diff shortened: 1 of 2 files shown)


stephenamar-db:
@He-Pin it's going to be a partial revert, because your optimizations fixed a bug we had with short-circuiting boolean ops.

He-Pin:
@stephenamar-db Thank you. I think another weird bug left is https://github.com/databricks/sjsonnet/issues/331, anyway, we are unlikely to write that code :)

stephenamar-db:
My gut feeling is that the optimizations around Val.Obj caused them to be evaluated early.

stephenamar-db:
> @stephenamar-db Thank you. I think another weird bug left is #331, anyway, we are unlikely to write that code :)

I'm looking at this too. trying to figure what what needs to be changed in the optimizer to make them work.

He-Pin:
Cool, I think the bug is if there is an `a[x]` , where `[x]` is a params, and then, we should not visit the `x` which is `Id(x)` to resolve the parameter, because before evaluation, the `x` is not binding to the actually value.

That's my guess, and another thing is, we may just pass the `Field` to the `Obj.Comp`, because it now contains two cases.
it turn out the `slice` one is simpler for me to handle :)

sjsonnet/src/sjsonnet/StaticOptimizer.scala
```diff
@@ -55,61 +52,6 @@ class StaticOptimizer(
...
-    case BinaryOp(pos, Val.Num(_, l), BinaryOp.OP_!=, Val.Num(_, r)) => Val.bool(pos, l != r)
-    case BinaryOp(pos, Val.Str(_, l), BinaryOp.OP_!=, Val.Str(_, r)) => Val.bool(pos, l != r)
-    case BinaryOp(pos, Val.Num(_, l), BinaryOp.OP_<, Val.Num(_, r)) => Val.bool(pos, l < r)
-    case BinaryOp(pos, Val.Str(_, l), BinaryOp.OP_<, Val.Str(_, r)) => Val.bool(pos, l < r)
-    case BinaryOp(pos, Val.Num(_, l), BinaryOp.OP_>, Val.Num(_, r)) => Val.bool(pos, l > r)
-    case BinaryOp(pos, Val.Str(_, l), BinaryOp.OP_>, Val.Str(_, r)) => Val.bool(pos, l > r)
-    case BinaryOp(pos, Val.Num(_, l), BinaryOp.OP_<=, Val.Num(_, r)) => Val.bool(pos, l <= r)
-    case BinaryOp(pos, Val.Str(_, l), BinaryOp.OP_<=, Val.Str(_, r)) => Val.bool(pos, l <= r)
-    case BinaryOp(pos, Val.Num(_, l), BinaryOp.OP_>=, Val.Num(_, r)) => Val.bool(pos, l >= r)
-    case BinaryOp(pos, Val.Str(_, l), BinaryOp.OP_>=, Val.Str(_, r)) => Val.bool(pos, l >= r)
```
He-Pin, line 92: Maybe we can extract all `BinaryOp` to a single method with `transformBinaryOp`, which may help too.
@stephenamar-db

PR #346, opened 2025-04-29
chore: Add static optimizer with seperated method.

He-Pin:
Motivation: 
Try to re-add https://github.com/databricks/sjsonnet/pull/344 back.

@stephenamar-db , How about this one? I separate the code into dedicated methods.


stephenamar-db:
I need more time to look.
Removing all remaining static optimizations for booleans actually improves performance. There's something going on that I don't get.

He-Pin:
@stephenamar-db I'm fine, our use cases it the scripts are been parsed and we will use them several times, as the current parse caches are separated for `script` and `extVar` and `tlaVar`, so it would be better to cache a parsed and optimized scripts' AST.

Another thing is if the `tableSwitch` did not affect the performance, we can make use of it to do a quick O(1) jump, my current work is not been blocked by any of this, let's take it easy, thank you.

PR #350, opened 2025-04-29
chore: Make variable resolver allocation less.

He-Pin:
Motivation:
The current way I PRed using many allocations, better to make it less allocations.

But the code is not elegant now


sjsonnet/test/src/sjsonnet/ExtFunctionNamespaceTest.scala
```diff
@@ -21,22 +24,25 @@ object ExtFunctionNamespaceTest extends TestSuite with FunctionModule {
...
+      new StaticOptimizer(ev, std, internedStrings, internedStaticFieldSets) {
+        override def variableResolver(name: String): Expr = {
+          if (name == "$ext" || name == "ext") {
+            module
+          } else {
+            null
+          }
+        }
+      }
+    }
```
He-Pin, line 44: This style is not that nice, but it still works.

Issue #284, opened 2025-04-18
java.lang.StackOverflowError with deep recursive function call.


He-Pin:
`recursive_function` utest may fail if the stack is not rich.

A better way may be to evaluate with an `ArrayDequeue`, but this will need some rework on the evaluator.

```java
java.lang.StackOverflowError
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1027)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150)
	at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:862)
	at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:760)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(BuiltinClassLoader.java:681)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:639)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
	at sjsonnet.Error$$anonfun$withStackFrame$1.isDefinedAt(Error.scala:63)
	at sjsonnet.Evaluator.visitExpr(Evaluator.scala:72)
	at sjsonnet.Evaluator$$anon$1.evalRhs(Evaluator.scala:599)
	at sjsonnet.Val$Func.apply2(Val.scala:553)
	at sjsonnet.Evaluator.visitApply2(Evaluator.scala:241)
	at sjsonnet.Evaluator.visitExpr(Evaluator.scala:54)
	at sjsonnet.Evaluator.visitIfElse(Evaluator.scala:148)
	at sjsonnet.Evaluator.visitExpr(Evaluator.scala:51)
	at sjsonnet.Evaluator$$anon$1.evalRhs(Evaluator.scala:599)
```

He-Pin:
I checked both go-jsonnet and cpp jsonnet, I found both are using a ArrayDeque like thing to do a stack , which may is more complex, may not need that for now?

and in https://github.com/CertainLach/jrsonnet/pull/121 , @CertainLach is using a hypered one.

PR #358, opened 2025-05-06
Fix multiple object comprehension bugs

### LINKED ISSUE #331: bug: defining functions in object comprehensions

### LINKED ISSUE #357: Evaluator.visitObjComp doesn't propagate context correctly
I wasn't yet able to fully root cause the issue but I derived a minimal example

main.jsonnet
```
local lib = import "libbroken.libsonnet";

{
  base:: {}
}
+ lib.foo()
```

libbroken.libsonnet
```
{
  foo()::
    {
      local global = self,

      [iterParam]: global.base {
        foo: iterParam
      }
      for iterParam in ["foo"]
    },
}
```

The issue seems to be specific to how the for loop (triggering `visitObjComp`) is handled. The output is generated properly when the lib code is replaced by

libworking.libsonnet
```
{
  foo()::
    {
      local global = self,

      foo: global.base {
        foo: "bar"
      }
    },
}
```

It also generates properly if the `global` variable is omitted and `self` is used directly.

JoshRosen:
This PR fixes multiple bugs in object comprehensions (fixes #357, fixes #331, plus another bug).

## Mis-handling of `self` and `super` in object comprehension locals when inherited

Object comprehensions may define local variables and those variables may reference `self` or `super`.

Consider the reproduction reported in https://github.com/databricks/sjsonnet/issues/357:

```jsonnet
local lib = {
  foo()::
    {
      local global = self,

      [iterParam]: global.base {
        foo: iterParam
      }
      for iterParam in ["foo"]
    },
};

{
 base:: {}
}
+ lib.foo()
```

This is supposed to output

```json
{
  "foo": {
    "foo": "foo"
  }
}
```

but sjsonnet outputs

```
sjsonnet.Error: Field does not exist: base
      at [Select base].(:7:26)
```

This bug occurs because of how bindings were managed: there's some tricky circular reference and variable rebinding logic (which I believe was added to handle references _between_ locals) and that was rebinding the `self` reference to the comprehension result itself: that is wrong because the `self` reference may change if an object is inherited or extended.

While digging into this, I also discovered a related bug where `super` wasn't being properly re-bound in object comprehension locals or values:

```jsonnet
local lib = {
  foo():: {
    local sx = super.x,
    [k]: sx + 1
    for k in ["x"]
  },
};

{ x: 2 }
+ lib.foo()
```

was failing with

```
java.lang.Exception: sjsonnet.Error: Attempt to use `super` when there is no super class
      at [SelectSuper x].(:4:21)
      at [ValidId sx].(:5:10)
      at [BinaryOp +].(:5:13)
```

## Silent dropping of fields if key is not a string

The object comprehension

```jsonnet
{[k]: k for k in [1]}
```

fails in regular jsonnet with an error

```
RUNTIME ERROR: field must be string, got: number
        <cmdline>:1:1-22
```

but in sjsonnet it was silently returning `{}`. This regression was [introduced](https://github.com/databricks/sjsonnet/pull/226/files#diff-7e786f55f42d493f85bfb37e5181c1c942ea039b21a307680e4e6913a8084aa2L628-R631) in #226 by adding a default `case _ =>` (presumably to fix a compiler or IDE warning).

## Support function definition via object comprehensions

This fixes https://github.com/databricks/sjsonnet/issues/331 by adding parser support for defining functions using method syntax, e.g.

```jsonnet
{
  [a](x): x * 2
  for a in ["f1", "f2", "f3"]
}
```

## Fixes

This PR fixes the above bugs by changing logic within `Evaluator.visitObjComp`. It is best reviewed commit-by-commit or using a whitespace-free diff.

### DIFF
--- a/sjsonnet/src/sjsonnet/Evaluator.scala
+++ b/sjsonnet/src/sjsonnet/Evaluator.scala
@@ -864,7 +864,11 @@ class Evaluator(
               Error.fail(s"Duplicate key ${k} in evaluated object comprehension.", e.pos);
             }
           case Val.Null(_) => // do nothing
-          case _           =>
+          case x =>
+            Error.fail(
+              s"Field name must be string or null, not ${x.prettyName}",
+              e.pos
+            )
         }
       }
       val valueCache = if (sup == null) {
--- a/sjsonnet/src/sjsonnet/Evaluator.scala
+++ b/sjsonnet/src/sjsonnet/Evaluator.scala
@@ -846,6 +846,9 @@ class Evaluator(
             k,
             new Val.Obj.Member(e.plus, Visibility.Normal) {
               def invoke(self: Val.Obj, sup: Val.Obj, fs: FileScope, ev: EvalScope): Val = {
+                // There is a circular dependency between `newScope` and `newBindings` because
+                // bindings may refer to other bindings (e.g. chains of locals that build on
+                // each other):
                 lazy val newScope: ValScope = s.extend(newBindings, self, sup)
                 lazy val newBindings = visitBindings(binds, newScope)
                 visitExpr(e.value)(newScope)
--- a/sjsonnet/src/sjsonnet/Evaluator.scala
+++ b/sjsonnet/src/sjsonnet/Evaluator.scala
@@ -680,13 +680,13 @@ class Evaluator(
         visitExpr(k) match {
           case Val.Str(_, k1) => k1
           case Val.Null(_)    => null
-          case x =>
-            Error.fail(
-              s"Field name must be string or null, not ${x.prettyName}",
-              pos
-            )
+          case x              => fieldNameTypeError(x, pos)
         }
     }
+  }
+
+  private def fieldNameTypeError(fieldName: Val, pos: Position): Nothing = {
+    Error.fail(s"Field name must be string or null, not ${fieldName.prettyName}", pos)
   }
 
   def visitMethod(rhs: Expr, params: Params, outerPos: Position)(implicit
@@ -856,11 +856,7 @@ class Evaluator(
             Error.fail(s"Duplicate key ${k} in evaluated object comprehension.", e.pos);
           }
         case Val.Null(_) => // do nothing
-        case x =>
-          Error.fail(
-            s"Field name must be string or null, not ${x.prettyName}",
-            e.pos
-          )
+        case x           => fieldNameTypeError(x, e.pos)
       }
     }
     val valueCache = if (sup == null) {
--- a/sjsonnet/src/sjsonnet/Evaluator.scala
+++ b/sjsonnet/src/sjsonnet/Evaluator.scala
@@ -685,7 +685,7 @@ class Evaluator(
     }
   }
 
-  private def fieldNameTypeError(fieldName: Val, pos: Position): Nothing = {
+  @inline private def fieldNameTypeError(fieldName: Val, pos: Position): Nothing = {
     Error.fail(s"Field name must be string or null, not ${fieldName.prettyName}", pos)
   }
 
--- a/sjsonnet/src/sjsonnet/Evaluator.scala
+++ b/sjsonnet/src/sjsonnet/Evaluator.scala
@@ -697,9 +697,7 @@ class Evaluator(
       override def evalDefault(expr: Expr, vs: ValScope, es: EvalScope) = visitExpr(expr)(vs)
     }
 
-  def visitBindings(
-      bindings: Array[Bind],
-      scope: => ValScope): Array[Lazy] = {
+  def visitBindings(bindings: Array[Bind], scope: => ValScope): Array[Lazy] = {
     val arrF = new Array[Lazy](bindings.length)
     var i = 0
     while (i < bindings.length) {
--- a/sjsonnet/src/sjsonnet/Evaluator.scala
+++ b/sjsonnet/src/sjsonnet/Evaluator.scala
@@ -852,9 +852,7 @@ class Evaluator(
               def invoke(self: Val.Obj, sup: Val.Obj, fs: FileScope, ev: EvalScope): Val = {
                 lazy val newScope: ValScope = s.extend(newBindings, self, sup)
                 lazy val newBindings = visitBindings(binds, (self, sup) => newScope)
-                visitExpr(e.value)(
-                  s.extend(newBindings, self, null)
-                )
+                visitExpr(e.value)(newScope)
               }
             }
           )
--- a/sjsonnet/src/sjsonnet/Evaluator.scala
+++ b/sjsonnet/src/sjsonnet/Evaluator.scala
@@ -841,45 +841,40 @@ class Evaluator(
   def visitObjComp(e: ObjBody.ObjComp, sup: Val.Obj)(implicit scope: ValScope): Val.Obj = {
     val binds = e.preLocals ++ e.postLocals
     val compScope: ValScope = scope // .clearSuper
-
-    lazy val newSelf: Val.Obj = {
-      val builder = new java.util.LinkedHashMap[String, Val.Obj.Member]
-      for (s <- visitComp(e.first :: e.rest, Array(compScope))) {
-        visitExpr(e.key)(s) match {
-          case Val.Str(_, k) =>
-            val prev_length = builder.size()
-            builder.put(
-              k,
-              new Val.Obj.Member(e.plus, Visibility.Normal) {
-                def invoke(self: Val.Obj, sup: Val.Obj, fs: FileScope, ev: EvalScope): Val = {
-                  lazy val newScope: ValScope = s.extend(newBindings, self, sup)
-                  lazy val newBindings = visitBindings(binds, (self, sup) => newScope)
-                  visitExpr(e.value)(
-                    s.extend(newBindings, self, null)
-                  )
-                }
+    val builder = new java.util.LinkedHashMap[String, Val.Obj.Member]
+    for (s <- visitComp(e.first :: e.rest, Array(compScope))) {
+      visitExpr(e.key)(s) match {
+        case Val.Str(_, k) =>
+          val prev_length = builder.size()
+          builder.put(
+            k,
+            new Val.Obj.Member(e.plus, Visibility.Normal) {
+              def invoke(self: Val.Obj, sup: Val.Obj, fs: FileScope, ev: EvalScope): Val = {
+                lazy val newScope: ValScope = s.extend(newBindings, self, sup)
+                lazy val newBindings = visitBindings(binds, (self, sup) => newScope)
+                visitExpr(e.value)(
+                  s.extend(newBindings, self, null)
+                )
               }
-            )
-            if (prev_length == builder.size() && settings.noDuplicateKeysInComprehension) {
-              Error.fail(s"Duplicate key ${k} in evaluated object comprehension.", e.pos);
             }
-          case Val.Null(_) => // do nothing
-          case x =>
-            Error.fail(
-              s"Field name must be string or null, not ${x.prettyName}",
-              e.pos
-            )
-        }
+          )
+          if (prev_length == builder.size() && settings.noDuplicateKeysInComprehension) {
+            Error.fail(s"Duplicate key ${k} in evaluated object comprehension.", e.pos);
+          }
+        case Val.Null(_) => // do nothing
+        case x =>
+          Error.fail(
+            s"Field name must be string or null, not ${x.prettyName}",
+            e.pos
+          )
       }
-      val valueCache = if (sup == null) {
-        Val.Obj.getEmptyValueCacheForObjWithoutSuper(builder.size())
-      } else {
-        new java.util.HashMap[Any, Val]()
-      }
-      new Val.Obj(e.pos, builder, false, null, sup, valueCache)
     }
-
-    newSelf
+    val valueCache = if (sup == null) {
+      Val.Obj.getEmptyValueCacheForObjWithoutSuper(builder.size())
+    } else {
+      new java.util.HashMap[Any, Val]()
+    }
+    new Val.Obj(e.pos, builder, false, null, sup, valueCache)
   }
 
   @tailrec
--- a/sjsonnet/src/sjsonnet/Evaluator.scala
+++ b/sjsonnet/src/sjsonnet/Evaluator.scala
@@ -685,7 +685,7 @@ class Evaluator(
     }
   }
 
-  @inline private def fieldNameTypeError(fieldName: Val, pos: Position): Nothing = {
+  private def fieldNameTypeError(fieldName: Val, pos: Position): Nothing = {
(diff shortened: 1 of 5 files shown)


sjsonnet/src/sjsonnet/Evaluator.scala
```diff
@@ -841,42 +841,40 @@ class Evaluator(
   def visitObjComp(e: ObjBody.ObjComp, sup: Val.Obj)(implicit scope: ValScope): Val.Obj = {
     val binds = e.preLocals ++ e.postLocals
     val compScope: ValScope = scope // .clearSuper
-
-    lazy val newSelf: Val.Obj = {
```
JoshRosen, line 845: Since `newSelf` was no longer referenced after other changes, I decided to remove it to simplify the code and to avoid the risk of future mis-use of it.

sjsonnet/src/sjsonnet/Evaluator.scala
```diff
@@ -841,42 +841,40 @@ class Evaluator(
...
+    val builder = new java.util.LinkedHashMap[String, Val.Obj.Member]
+    for (s <- visitComp(e.first :: e.rest, Array(compScope))) {
+      visitExpr(e.key)(s) match {
+        case Val.Str(_, k) =>
+          val prev_length = builder.size()
+          builder.put(
+            k,
+            new Val.Obj.Member(e.plus, Visibility.Normal) {
+              def invoke(self: Val.Obj, sup: Val.Obj, fs: FileScope, ev: EvalScope): Val = {
+                lazy val newScope: ValScope = s.extend(newBindings, self, sup)
```
JoshRosen, line 853: There are two changes here:

1. use `self` instead of `newSelf`, which allows for bindings' references to `self` to be properly updated when re-evaluated.
2. Pass in `sup` to allow `super` references to resolve properly.

sjsonnet/src/sjsonnet/Evaluator.scala
```diff
@@ -841,42 +841,40 @@ class Evaluator(
...
+                lazy val newBindings = visitBindings(binds, (self, sup) => newScope)
+                visitExpr(e.value)(
+                  s.extend(newBindings, self, null)
```
JoshRosen, line 856: On closer look, I see that this is still passing `null` for `newSuper` here and that probably indicates that there's still a latent bug. Let me see if I can devise a test case.

JoshRosen: Indeed:

- Test: https://github.com/databricks/sjsonnet/pull/358/commits/55bcb2684a39a386981cb1f5ded712a92dcc3f82
- Fix: https://github.com/databricks/sjsonnet/pull/358/commits/fd038d84823f521c6213435730dcd373a7add26a

JoshRosen:
I noticed that `ValScope.extend` and `Evaluator.visitBindings` now each only have a single callsite, which might permit some further simplification.

He-Pin:
https://github.com/databricks/sjsonnet/issues/331

@JoshRosen Does this fix #331 too?

sjsonnet/src/sjsonnet/Evaluator.scala
```diff
@@ -841,42 +837,38 @@ class Evaluator(
...
-        new java.util.HashMap[Any, Val]()
+          )
+          if (prev_length == builder.size() && settings.noDuplicateKeysInComprehension) {
+            Error.fail(s"Duplicate key ${k} in evaluated object comprehension.", e.pos);
+          }
+        case Val.Null(_) => // do nothing
+        case x =>
+          Error.fail(
+            s"Field name must be string or null, not ${x.prettyName}",
+            e.pos
```
He-Pin, line 862: `Field name must be string or null, got ${x.prettyName}`

sjsonnet/src/sjsonnet/ValScope.scala
```diff
@@ -18,23 +18,11 @@ final class ValScope private (val bindings: Array[Lazy]) extends AnyVal {
...
   def length: Int = bindings.length
 
-  def extend(
-      newBindingsF: Array[(Val.Obj, Val.Obj) => Lazy] = null,
-      newSelf: Val.Obj,
-      newSuper: Val.Obj): ValScope = {
-    val by = if (newBindingsF == null) 2 else newBindingsF.length + 2
-    val b = Arrays.copyOf(bindings, bindings.length + by)
+  def extend(newBindings: Array[Lazy], newSelf: Val.Obj, newSuper: Val.Obj): ValScope = {
+    val b = Arrays.copyOf(bindings, bindings.length + newBindings.length + 2)
```
He-Pin, line 22: bindings.length and newbindings.length can be extracted to a local variable?

sjsonnet/src/sjsonnet/Evaluator.scala
```diff
@@ -841,42 +837,38 @@ class Evaluator(
...
+    for (s <- visitComp(e.first :: e.rest, Array(compScope))) {
+      visitExpr(e.key)(s) match {
+        case Val.Str(_, k) =>
+          val prev_length = builder.size()
+          builder.put(
+            k,
+            new Val.Obj.Member(e.plus, Visibility.Normal) {
+              def invoke(self: Val.Obj, sup: Val.Obj, fs: FileScope, ev: EvalScope): Val = {
+                lazy val newScope: ValScope = s.extend(newBindings, self, sup)
+                lazy val newBindings = visitBindings(binds, newScope)
```
He-Pin, line 850: This code is a little weird, it references each other

sjsonnet/src/sjsonnet/Evaluator.scala
JoshRosen: I copied this wording from https://github.com/databricks/sjsonnet/blob/84ac3207d3b615ce887b5b99263f143655f60487/sjsonnet/src/sjsonnet/Evaluator.scala#L683-L687 further up in this file; I suppose I can factor it out into a helper function.

sjsonnet/src/sjsonnet/ValScope.scala
JoshRosen: I would prefer to leave this as-is for now, given that the other methods (including `extendSimple`) have similar repetition of `.length`. My understanding is that JIT should already be optimizing this nicely.

sjsonnet/src/sjsonnet/Evaluator.scala
JoshRosen: I factored this out in bd4e1b2fb2efe578cf2b3277d4c5806e79cc0263 but, curiously, this appears to cause a test failure in the JDK 17 build: https://github.com/databricks/sjsonnet/actions/runs/14891478275/job/41824335648 :

```
java.lang.StackOverflowError
    java.lang.String.rangeCheck(String.java:304)
    java.lang.String.<init>(String.java:300)
    jdk.internal.org.objectweb.asm.ClassReader.readUtf(ClassReader.java:3717)
    jdk.internal.org.objectweb.asm.ClassReader.readUtf(ClassReader.java:3685)
    jdk.internal.org.objectweb.asm.ClassReader.readUTF8(ClassReader.java:3666)
    jdk.internal.org.objectweb.asm.ClassReader.readConst(ClassReader.java:3841)
    java.lang.invoke.MethodHandles$Lookup$ClassFile.newInstance(MethodHandles.java:2258)
    java.lang.invoke.MethodHandles$Lookup.makeHiddenClassDefiner(MethodHandles.java:2359)
    java.lang.invoke.MethodHandles$Lookup.defineHiddenClass(MethodHandles.java:2127)
    java.lang.invoke.InnerClassLambdaMetafactory.generateInnerClass(InnerClassLambdaMetafactory.java:407)
    java.lang.invoke.InnerClassLambdaMetafactory.spinInnerClass(InnerClassLambdaMetafactory.java:315)
    java.lang.invoke.InnerClassLambdaMetafactory.buildCallSite(InnerClassLambdaMetafactory.java:228)
    java.lang.invoke.LambdaMetafactory.altMetafactory(LambdaMetafactory.java:536)
    java.lang.invoke.BootstrapMethodInvoker.invoke(BootstrapMethodInvoker.java:147)
    java.lang.invoke.CallSite.makeSite(CallSite.java:315)
    java.lang.invoke.MethodHandleNatives.linkCallSiteImpl(MethodHandleNatives.java:281)
    java.lang.invoke.MethodHandleNatives.linkCallSite(MethodHandleNatives.java:271)
    os.Path.relativeTo(Path.scala:584)
    sjsonnet.OsPath.relativeToString(OsPath.scala:6)
    sjsonnet.Error$Frame.<init>(Error.scala:51)
    sjsonnet.Error$.fail(Error.scala:71)
    sjsonnet.Materializer.apply0(Materializer.scala:72)
    sjsonnet.Materializer.apply0(Materializer.scala:53)
    sjsonnet.Materializer.apply0(Materializer.scala:53)
[...]
```

I think the problem is that we have code within the `Materializer` which catches `StackOverflowError` and then tries to call `Error.fail` and that, in turn, hits a secondary `StackOverflow` because the stack is already so deep:

https://github.com/databricks/sjsonnet/blob/84ac3207d3b615ce887b5b99263f143655f60487/sjsonnet/src/sjsonnet/Materializer.scala#L70-L73

I tried adding an `@inline` to the new helper to see if that would somehow hack around this (https://github.com/databricks/sjsonnet/pull/358/commits/f5e40c2b8e172825438e396f1cc5ee5577dc20ad) but it didn't help.

This preexisting design seems very fragile.

If we want to more fundamentally fix this and remove the core fragility, I think we probably need to allow the `StackOverflowError` to bubble higher rather than catching it so deep, but that introduces some new challenges around making sure that we preserve the `v` which triggered the error. I'm thinking that we might want to re-throw a wrapped exception containing a pointer to `v` and then catch it and transform it to an `Error.fail` at a higher level of the callstack (e.g. the outermost `apply0` call). That's a larger refactoring / change, though, and I don't want to co-mingle it here.

Given this, I'm going to back out of the error message DRY and will re-attempt it in a separate PR.

JoshRosen:
> #331
> 
> I tested with
> 
> ```scala
>       eval(
>         """
>           |local funcs = {
>           |  [a](x): x * 2
>           |  for a in ["f1", "f2", "f3"]
>           |};
>           |funcs.f1(10)
>           |""".stripMargin) ==> ujson.Num(20)
> ```
> 
> gets:
> 
> ```
> sjsonnet.StaticError: Unknown variable: x
>     at [Id x].(:3:11)
> 
> java.lang.Exception: sjsonnet.StaticError: Unknown variable: x
>     at [Id x].(:3:11)
> ```

I think more / different work is needed to support this: that `StaticError` is being thrown before we even begin evaluation, so it's not even hitting any of the `Evaluator` code modified in this PR.

As you noted in https://github.com/databricks/sjsonnet/pull/337, the problem is that we're currently discarding any knowledge of RHS args (denoting that it's a function).

This turned out to be fairly straightforward to fix: I just needed to modify the parser to create an `Expr.Function` when `args` are present: e06e7a7fc286ae4b8b5590cd654715a5124f7364 ; I'll update the PR description to reflect this fix.

sjsonnet/src/sjsonnet/Evaluator.scala
JoshRosen: I think that this is a test ordering issue: the current test for the stack overflow handling fails if it runs prior to other tests: it was implicitly relying on other tests to perform certain loading and initialization steps.

We should fix this properly (e.g. by hardening the actual implementation, as this is a real bug), but as a temporary "unblock the test flakiness" hack I'm trying to force a particular loading step: 5cd4ed8dbd321561ce4583762f11cda407aa5c4f

He-Pin: `@inline` is just a hint; we can't rely on it.
For StackOverflow, I raised https://github.com/databricks/sjsonnet/issues/284 once, but it may hurt some performance; there is some TCO in other implementations.

anujpunekar20:
hi

stephenamar-db (approved):

stephenamar-db:
Thanks @JoshRosen

He-Pin:
@JoshRosen Thank you , helps a lot

PR #379, opened 2025-05-14
Fix tailstrict implementation and add more tests

### LINKED ISSUE #359: Tailstrict should only apply to the function, not to the whole evaluator
Currently, the tailstrict modifier is applied to the entire evaluator when it comes across a function marked as tailstrict, e.g. in this snippet from `Evaluator.visitApply2`:

```
   ...
   if (tailstrict) {
      lhs.cast[Val.Func].apply2(visitExpr(e.a1), visitExpr(e.a2), e.pos)
    } else if (e.tailstrict) {
      tailstrict = true
      val res = lhs.cast[Val.Func].apply2(visitExpr(e.a1), visitExpr(e.a2), e.pos)
      tailstrict = false
      res
    }
   ...
```

This has undesirable effects on other function / method calls in the subtree of evaluating the tailstrict function. To demonstrate a few with examples:

## All function parameters are evaluated eagerly.

```
local helper = {
    local this = self,

    bar(param, unusedParam)::
        param,

    foo(param)::
        local sum_from_one_to_n(counter, n) =
          if n == 0 then
            counter
          else
            sum_from_one_to_n(this.bar(counter + n, error "Kaboom!"), n - 1) tailstrict;

        local result = sum_from_one_to_n(0, param);
        result,
};

{
    bar: helper.foo(10)
}
```

Fails with `Kaboom!` while gojsonnet generates this successfully. 

 ## Using named parameters in tailstrict context results in missing bindings (or worse, shifted bindings so the function params can point to different values):

```
local helper = {
    local this = self,

    bar(param, auxone=0, auxtwo=0)::
        param + auxone + auxtwo,

    foo(param)::
        local sum_from_one_to_n(counter, n) =
          if n == 0 then
            counter
          else
            sum_from_one_to_n(this.bar(counter + n, auxtwo=1), n - 1) tailstrict;

        local result = sum_from_one_to_n(0, param);
        result,
};

{
    bar: helper.foo(10)
}

```

Fails with IOOB, as auxone doesn't get bound - gojsonnet processes this successfully as well.

The issues can be hacked around in [Val.scala](https://github.com/databricks/sjsonnet/blob/84ac3207d3b615ce887b5b99263f143655f60487/sjsonnet/src/sjsonnet/Val.scala#L493C7-L499C8), but this doesn't solve the issue of tainting the evaluator. I didn't explore the tailstrict implementation in enough detail to see if not setting the evaluator `tailstrict` flag fixes all the corner cases, but it does seem to work at first glance (the local tailstrict flag is preserved in the recursive calls).

@stephenamar-db FYI, as you added the tailstrict implementation.

stephenamar-db:
Fixes #359


stephenamar-db:
@He-Pin @Danstahrm

He-Pin:
1. there is a `long_chains.jsonnet` in google/jsonnet repo
2. there are more tests about tailstrict in google/go-jsonnet repo, eg tailstrict_operator2.jsonnet

We can includes both here too.

sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -777,13 +793,16 @@ object Val {
   }
 }
 
+object TailstrictMode extends Enumeration {
+  type TailstrictMode = Value
+  val Enabled, Disabled = Value
```
He-Pin, line 798: Use `seal trait` instead?

Issue #316, opened 2025-04-25
port: fix: improve safety checks for left shift operations


He-Pin:
refs: https://github.com/google/jsonnet/pull/1217

https://github.com/google/jsonnet/commit/804c14263a1f91716391abc8c7c9a026c3fc785f

Issue #315, opened 2025-04-25
port: fix: add safe double-to-int64 conversion for bitwise operations


He-Pin:
refs:  https://github.com/google/jsonnet/pull/1217

[fix: add safe double-to-int64 conversion for bitwise operations](https://github.com/google/jsonnet/commit/9b430ae19c9a6799c3ce1b740b4497cad84d5379)

PR #381, opened 2025-05-14
correct safe integer range for bitwise ops

### LINKED ISSUE #317: port: fix: correct safe integer range for bitwise ops
refs: https://github.com/google/jsonnet/pull/1217
https://github.com/google/jsonnet/commit/499b0dba4079cf9db35b98078feef5aef2298a6d
refs: https://github.com/google/jsonnet/pull/1240

golden https://github.com/google/jsonnet/pull/1217

stephenamar-db:
Fixes #317 #316 #315

### DIFF
--- a/sjsonnet/src/sjsonnet/Val.scala
+++ b/sjsonnet/src/sjsonnet/Val.scala
@@ -81,6 +81,10 @@ object PrettyNamed {
   implicit val nullName: PrettyNamed[Val.Null] = new PrettyNamed("null")
 }
 object Val {
+  // Constants for safe double-to-int conversion
+  // IEEE 754 doubles precisely represent integers up to 2^53, beyond which precision is lost
+  private val DOUBLE_MAX_SAFE_INTEGER = (1L << 53) - 1
+  private val DOUBLE_MIN_SAFE_INTEGER = -((1L << 53) - 1)
 
   abstract class Literal extends Val with Expr {
     final override private[sjsonnet] val _tag = ExprTags.`Val.Literal`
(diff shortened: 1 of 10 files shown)


sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -81,6 +81,10 @@ object PrettyNamed {
   implicit val nullName: PrettyNamed[Val.Null] = new PrettyNamed("null")
 }
 object Val {
+  // Constants for safe double-to-int conversion
+  // IEEE 754 doubles precisely represent integers up to 2^53, beyond which precision is lost
+  private val DOUBLE_MAX_SAFE_INTEGER = (1L << 53) - 1
+  private val DOUBLE_MIN_SAFE_INTEGER = -((1L << 53) - 1)
```
He-Pin, line 87: It would be nice to have the same comments:

Integers can be safely represented as a Jsonnet number in the range [-2^53 + 1,2^53 - 1]. This is [a direct consequence of IEEE754 spec](https://en.wikipedia.org/wiki/Double-precision_floating-point_format), with the requirements that a safe integer is representable exactly, and cannot be produced by rounding any other integer to fit the IEEE-754 representation. See also, the JavaScript [`Number.MAX_SAFE_INTEGER`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER), [`Number.MIN_SAFE_INTEGER`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MIN_SAFE_INTEGER), and [`Number.isSafeInteger`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger) definitions.

PR #382, opened 2025-05-15
Import the most recent test_suite from google/jsonnet

stephenamar-db:
Pretty large PR.
This imports all of non-lint test_suites from google/jsonnet and:

- Regenerate each golden file with the script
- Change ErrorTests to automatically pick the files up and test them.

Still need to do FileTests, but it already picked up a lot of small bugs that we'll have to fix like:

- Formatting issues with number with exponents
- Bad rounding with the number 0
....


stephenamar-db:
Sadly, I have to have big test() and then loop through each case.
Would love for https://github.com/com-lihaoyi/utest/issues/367 to be addressed.

stephenamar-db:
@He-Pin

He-Pin:
There are some additional tests live in google/go-jsonnet , has that been included too?

stephenamar-db:
that's a lot more to add. let's deal with the discrepancies found in the c++ test suite first, and we'll import the go-jsonnet test suite later.

PR #386, opened 2025-05-15
Add new GH workflow to build all the necessary artifacts before a release

stephenamar-db:
Add new GH workflow to build all the necessary artifacts before a release.
We are not allowed to create a draft release due to internal IP restricts at databricks through the workflow, but at least I can get all the binaries compiled and ready to go.


.github/workflows/release-build.yaml
```diff
@@ -0,0 +1,68 @@
...
+    env:
+      LANG: C
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-java@v4
+        with:
+          java-version: 17
+          distribution: 'zulu'
+      - name: JVM Build
+        run: ./mill sjsonnet.jvm[3.3.6].assembly
```
He-Pin, line 19: The version is better to be a variable

stephenamar-db: this is in progress.

build.sbt
```diff
@@ -1,7 +1,8 @@
-val sjsonnetVersion = "0.5.0"
+val sjsonnetVersion = IO.readLines(new File("VERSION")).head.trim
 
-val scala213 = "2.13.16"
-val scala3 = "3.3.6"
+val scalaVersions = IO.readLines(new File("SCALA_VERSIONS")).map(_.trim)
```
He-Pin, line 3: Seems like split is missing

PR #389, opened 2025-05-15
Fix readme to publish and allow for server/client to be pushed to maven

stephenamar-db:


He-Pin:
Seems the 0.5.1 doesn't contains the server jar？

PR #394, opened 2025-05-16
Add windows build

stephenamar-db:


stephenamar-db:
yup it's harder than I though. I'm not going to do it for now

PR #395, opened 2025-05-16
chore: Change Scala native's release mode to ReleaseFull

He-Pin:
Motivation:
Better performance
refs: https://scala-native.org/en/stable/user/sbt.html#compilation-modes

Maybe because of the sjsonnet size is now from `7.2MB` to `12MB`

it's https://github.com/scalameta/scalafmt/blob/main/build.sbt#L342 ReleaseFull in scalafmt

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -153,7 +153,7 @@ object sjsonnet extends Module {
       this.millSourcePath / "src-native",
       this.millSourcePath / "src-jvm-native"
     )
-    def releaseMode = ReleaseMode.ReleaseFast
+    def releaseMode = ReleaseMode.ReleaseFull
     def nativeLTO = LTO.Full
     def nativeMultithreading = None
 


He-Pin:
```scala
hepin@Mac ~ % hyperfine --warmup 10 "./out cmp.jsonnet"
Benchmark 1: ./out cmp.jsonnet
  Time (mean ± σ):     252.8 ms ±   4.2 ms    [User: 137.3 ms, System: 12.2 ms]
  Range (min … max):   249.0 ms … 264.2 ms    10 runs
 
hepin@Mac ~ % hyperfine --warmup 10 "./sjsonnet cmp.jsonnet"
Benchmark 1: ./sjsonnet cmp.jsonnet
  Time (mean ± σ):     225.9 ms ±   2.5 ms    [User: 138.9 ms, System: 12.0 ms]
  Range (min … max):   222.3 ms … 229.1 ms    11 runs
```

seems ReleaseFull is slower? @WojciechMazur

stephenamar-db (approved):

stephenamar-db:
Is this ready

He-Pin:
@stephenamar-db But the result binary size is bigger, and the benchmark for a single run is slower.

stephenamar-db:
it probably takes longer to load the large binary in memory?

He-Pin:
@stephenamar-db I think so.

CertainLach:
> it probably takes longer to load the large binary in memory?

Not sure how MacOS disk caching works, but hyperfine --warmup should remove all the difference in loading times

He-Pin:
@CertainLach Thanks, I saw scalafmt is using `ReleaseFull` too, but the result binary with `ReleaseFull` is 12MB and `ReleaseFast` is 7.2MB, and when do local testing, I saw, `ReleaseFast` result seems faster.

PR #402, opened 2025-05-20
Add more test coverage for std.ExtVar.

Danstahrm:
Before 60d59f99a7a91201368d0f9c9fc04e60289349cb, lazily evaluated expressions couldn't be used as keys in extVar calls. This has been fixed already - adding a regression test.


sjsonnet/test/src/sjsonnet/Std0150FunctionsTests.scala
```diff
@@ -100,6 +100,7 @@ object Std0150FunctionsTests extends TestSuite {
 
       check("""std.extVar("num")""", 1)
       check("""std.extVar("str")""", "hello")
+      check("""std.extVar("s" + "tr")""", "hello")
```
He-Pin, line 103: Does this really test something?

He-Pin: I think we should replace the `"s" + "tr"` with a function call to test what you want.

He-Pin, line 103: We can add another test with a function call to make it more explicit:)

Danstahrm: Good point - coming right up.

PR #404, opened 2025-05-21
Allow `std.map` to operate on string args as well.

Danstahrm:
In gojsonnet, a few functions of the stdlib accepts strings in addition to arrays. Sometimes this is mentioned in the stdlib documentation, sometimes it isn't - and importantly, this is a special property of a few select functions and it does NOT mean that a string can be treated as an array of single-character strings in every context with implicit type conversion between the two.

Eyeballing the gojsonnet code yields the following list of std functions which should support strings as parameters:

std.slice
std.member
std.map
std.flatMap
std.base64

Out of those, only `map` had missing support for strings, which is being added in this commit.

FTR, in the process, I discovered other quirks of the `map` function, such as the undocumented, `filterMap`-like behavior from this test case:

`std.assertEqual(std.map(function(x) x * x, std.filter(function(x) x > 5, std.range(1, 10))), [36, 49, 64, 81, 100])`

### DIFF
--- a/sjsonnet/src/sjsonnet/Std.scala
+++ b/sjsonnet/src/sjsonnet/Std.scala
@@ -449,12 +449,26 @@ class Std(
   private object Map_ extends Val.Builtin2("map", "func", "arr") {
     def evalRhs(_func: Lazy, arr: Lazy, ev: EvalScope, pos: Position): Val = {
       val func = _func.force.asFunc
+      val arg = arr.force
+      arg match {
+        case Val.Str(_, str) => evalStr(func, str, ev, pos.noOffset)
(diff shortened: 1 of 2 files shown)


He-Pin:
@Danstahrm Thanks, so it seems the google/jsonnet does not support this, only google/go-jsonnet?

Danstahrm:
> @Danstahrm Thanks, so it seems the google/jsonnet does not support this, only google/go-jsonnet?

`google/jsonnet` supports string args as well. Regarding the `filterMap` behavior, I was wrong - it's just a matter of the second argument not being a constant but an evaluated expression as well (off-by-one in the parentheses). `jsonnet` doesn't properly support it it seems like.

Issue #398, opened 2025-05-17
bug: std.format about `g`, `f` `G` is not right


He-Pin:
When testing `std.format` I found there are some tests been commented out.

```jsonnet
std.assertEqual(std.format('%#.1g', [0.99]), '1.') &&
std.assertEqual(std.format('%#.1g', [1.95555]), '2.') &&
std.assertEqual(std.format('%#.4g', [0.99995]), '1.000') &&
```

and currently the sjsonnet and jrsonnet @CertainLach gives `0.0` for `std.format('%#.1g', [0.99])`, but google/jsonnet gives `1.`

There are more tests about `std.format` lives in google/go-jsonnet

He-Pin:
And I found the `sjsonnet` and `jrsonnet` both have different result for `g` and `f` from google/jsonnet

He-Pin:
def formatInteger(formatted: FormatSpec, s: Double): String = {
    val sl = s.toLong

The code here is wrong, we need using BigDecimal to do the round up and ceiling

stephenamar-db:
yup, I'm fully aware of this. there's some complex rounding code they added to google/jsonnet a few years ago to get that result, but this will require some careful change on our side.

stephenamar-db:
https://github.com/google/jsonnet/commit/90cad75dcc2eafdcf059c901169d36539dc8a699 FYI

PR #408, opened 2025-05-23
Implement * in FormatSpec to dynamically insert width/precision

stephenamar-db:
Last part to fix the gaps found in https://github.com/databricks/sjsonnet/issues/398


He-Pin:
thank you

PR #410, opened 2025-05-23
Fix various small issues found when I imported all recent error tests

stephenamar-db:


sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -534,11 +534,19 @@ class Std(
   }
 
   private object DecodeUTF8 extends Val.Builtin1("decodeUTF8", "arr") {
-    def evalRhs(arr: Lazy, ev: EvalScope, pos: Position): Val =
-      new Val.Str(
+    def evalRhs(arr: Lazy, ev: EvalScope, pos: Position): Val = {
+      for ((v, idx) <- arr.force.asArr.iterator.zipWithIndex) {
+        if (!v.isInstanceOf[Val.Num] || !v.asDouble.isWhole || v.asInt < 0 || v.asInt > 255) {
+          throw Error.fail(
+            f"Element $idx of the provided array was not an integer in range [0,255]"
```
He-Pin, line 541: Maybe it's better to prints what it is

He-Pin:
I found there are many tests in google/go-jsonnet too, imports which will help to find more potential small issues.

PR #425, opened 2025-06-08
Properly implement file hashing in scala native - using the built-in murmurhash

stephenamar-db:
…

### DIFF
--- a/sjsonnet/src-native/sjsonnet/Platform.scala
+++ b/sjsonnet/src-native/sjsonnet/Platform.scala
@@ -55,8 +55,11 @@ object Platform {
   def sha3(s: String): String = computeHash("SHA3-512", s)
 
   def hashFile(file: File): String = {
-    // File hashes in Scala Native are just the file content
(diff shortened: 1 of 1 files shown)


sjsonnet/src-native/sjsonnet/Platform.scala
```diff
@@ -55,8 +55,11 @@ object Platform {
...
 
   def hashFile(file: File): String = {
-    // File hashes in Scala Native are just the file content
-    scala.io.Source.fromFile(file).mkString
+    scala.util.hashing.MurmurHash3
+      .orderedHash(
+        scala.io.Source.fromFile(file)
+      )
+      .toHexString
   }
```
He-Pin, line 63: Cool, and we may also use `CRC32`; then the code should be similar to the JVM one.

stephenamar-db: JVM uses xxhash64. It's not worth the effort of importing yet another library for this

PR #426, opened 2025-06-10
Update ScalaJS entrypoint and documentation

LaurelineSwissBorg:
* Updated the JS entrypoint to allow usage from plain JS/TS, the previous version required an `Either` object which could not be returned from the JS side.
* Updated the documentation to reflect the changes.
* Compile the output as a CommonJS module to make import easier.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -55,15 +55,20 @@ $ ./sjsonnet.jar foo.jsonnet
 
 Or from Javascript:
 
+Since `0.5.3`, the output is a CommonJS module.
+
 ```javascript
-$ node
+const { SjsonnetMain } = require("./sjsonnet.js");
+// or
+import { SjsonnetMain } from "./sjsonnet";
 
-> require("./sjsonnet.js")
+console.log(
+  SjsonnetMain.interpret("local f = function(x) x * x; f(11)", {}, {}, "", (wd, imported) => null)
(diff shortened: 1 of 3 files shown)


stephenamar-db:
You can just run `./mill mill.scalalib.scalafmt.ScalafmtModule/reformatAll __.sources`

stephenamar-db (approved):

He-Pin:
A simple scalafmt should help too.

readme.md
```diff
@@ -69,11 +69,11 @@ $ node
...
-    // loader callback: receives the tuple from the import callback and returns the file contents
-    ([path, content]) => content
+    // resolver callback: receives a base directory and the imported path string,
+    // returns the resolved path
+    (wd, imported) => wd + "/" + imported,
+    // loader callback: receives the full path and returns the file contents
+    (path, binary) => "local bar = 123; bar + bar"
     )
 '246bar'
 ```
```
He-Pin, line 85: Should we mention it's a CommonJS since `0.5.3`?

LaurelineSwissBorg: Updated the example to be more in-line with expected usage (using an ES6 compatible `require` or `import` statement with a binding) and mentioned that it's a CJS module from version 0.5.3.

PR #433, opened 2025-06-17
Fix ||| block parsing

### LINKED ISSUE #430: Fix ||| string parsing
||| string parsing had a lot of issues, from not supporting \r\n, to not successfully parsing something like

```
|||
foo
|||
``` 

(string content has the same indentation level as |||

stephenamar-db:
This PR fixes 2 edges cases that sjsonnet did not handle properly.

1. ||| blocks with \r\n
2. ||| blocks only with empty - indented - lines.

The only practical way I found to fix this is to half-rewrite the parsing code myself to better get a handle on how fastparse works.

This closes #430


sjsonnet/src/sjsonnet/Parser.scala
```diff
@@ -160,18 +141,37 @@ class Parser(
...
+    case (true, lines) =>
+      Seq(lines.mkString.stripLineEnd)
+    case (false, lines) =>
+      lines
+  }
+
+  def tripleBarString[$: P]: P[(Boolean, Seq[String])] = P(
+    ("||-" | "||").!.map(_.last == '-')./ ~~ CharsWhileIn(" \t", 0) ~~
+    // Detect the new line separator
+    ("\r\n" | "\n").!.flatMapX { sep =>
```
He-Pin: should we add a `/` after `.!`

sjsonnet/src/sjsonnet/Parser.scala
```diff
@@ -160,18 +141,37 @@ class Parser(
...
+    case (true, lines) =>
+      Seq(lines.mkString.stripLineEnd)
+    case (false, lines) =>
+      lines
+  }
+
+  def tripleBarString[$: P]: P[(Boolean, Seq[String])] = P(
+    ("||-" | "||").!.map(_.last == '-')./ ~~ CharsWhileIn(" \t", 0) ~~
+    // Detect the new line separator
+    ("\r\n" | "\n").!./.flatMapX { sep =>
```
He-Pin: if it fail, we should give an error message, the `text block must start with at least one whitespace` here

sjsonnet/test/src-jvm-native/sjsonnet/FileTests.scala
```diff
@@ -4,7 +4,7 @@ import utest.*
 
 object FileTests extends BaseFileTests {
   val testDataSkippedTests: Set[String] = Set(
-    "dos_line_endings.jsonnet",
+//    "dos_line_endings.jsonnet",
```
He-Pin: Add A comment here , why we skip this tests?

sjsonnet/src/sjsonnet/Parser.scala
```diff
@@ -160,18 +141,41 @@ class Parser(
...
+      lines.dropRight(1) ++ Seq(lines.last.stripLineEnd)
+    case (false, lines) =>
+      lines
+  }
+
+  def tripleBarString[$: P]: P[(Boolean, Seq[String])] = P(
+    ("||-" | "||").!.map(_.last == '-')./ ~~ CharsWhileIn(" \t", 0) ~~
+    // Detect the new line separator
+    ("\r\n" | "\n").opaque("|||-blocks require multiple lines").!./.flatMapX { sep =>
+      tripleBarStringLines(sep)
```
He-Pin: this line can be just `tripleBarStringLines`?

sjsonnet/src/sjsonnet/Parser.scala
```diff
@@ -160,18 +141,41 @@ class Parser(
...
+    } ~~ CharsWhileIn(" \t", 0) ~~ "|||"
```
He-Pin: can be have `CharsWhileIn(" \t", 0)` and `CharsWhileIn(" \t", 1)` a val?

sjsonnet/test/src-jvm-native/sjsonnet/FileTests.scala
He-Pin: The cpp jsonnet passes this test, not sure why this is skipped

stephenamar-db: not skipped anymore. that's the point of this PR.

PR #442, opened 2025-06-19
Refactor std.native functions out of Std

stephenamar-db:
In this PR,
I move xz/gzip/regex outside of Std, and move testing infrastructure to per-language directory.


He-Pin:
Nice to have ,otherwise some tests have to be skipped

sjsonnet/src-jvm-native/sjsonnet/stdlib/NativeGzip.scala
```diff
@@ -0,0 +1,16 @@
+package sjsonnet.stdlib
+
+import sjsonnet.{Error, EvalScope, Lazy, Platform, Position, Val}
+
+object NativeGzip {
```
He-Pin: How about make it extends the FunctionModule？

PR #447, opened 2025-06-24
fix(js,native): std.parseYaml() handling of "---".

rudo-thomas:
Problem: On js and native targets std.parseYaml() mishandles input where "---" is not at the beginning of the line. (The jvm target is fine.)

Fix: Only match "---" at the beginning of the line. Also don't require a newline after it.

Note that there are other corner cases that are still broken, but for those I'll file a bug against the underlying scala-yaml library.

### DIFF
--- a/sjsonnet/src-js/sjsonnet/Platform.scala
+++ b/sjsonnet/src-js/sjsonnet/Platform.scala
@@ -42,8 +42,10 @@ object Platform {
       Error.fail("Unsupported YAML node type: " + node.getClass.getSimpleName)
   }
 
+  private val docSplitPattern = Pattern.compile("(^|\n)---\\s*(\n|$)")
+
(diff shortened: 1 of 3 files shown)


sjsonnet/src-js/sjsonnet/Platform.scala
```diff
@@ -43,7 +43,7 @@ object Platform {
   }
 
   def yamlToJson(s: String): ujson.Value = {
-    val docs = s.split("---\\s*\n")
+    val docs = s.split("(^|\n)---\\s*(\n|$)")
```
He-Pin: We can compile the pattern and reuse it.

rudo-thomas: Fair enough. Done. PTAL

Issue #448, opened 2025-06-25
/dev/stdin doesn't work in TLAs (native, jvm) or as the top-level file (native)


rudo-thomas:
Goal: Have /dev/stdin work as the top-level file and in --tla-code-file on the
native target, just as it did on the jvm target in sjsonnet-0.5.2. This will
allow our project to replace the go-jsonnet interpreter with sjsonnet.

Problem: Native sjsonnet build doesn't work in either case; on jvm it did work
as expected in version 0.5.2, but on master I broke the --tla-code-file case in
https://github.com/databricks/sjsonnet/pull/432 (which converts --tla-code-file
and --ext-code-file to import expressions).

top-level file:
```
$ echo '{a:1}' | sjsonnet-master-36ecd2a7448d77b729786dd6336029475aba0eae.jar /dev/stdin
{
   "a": 1
}
$ echo '{a:1}' | sjsonnet-master-36ecd2a7448d77b729786dd6336029475aba0eae-native /dev/stdin
Exception in thread "main" java.io.IOException: lseek failed: Illegal seek
	at java.nio.channels.FileChannelImpl.throwPosixException(Unknown Source)
	at java.nio.channels.FileChannelImpl.position(Unknown Source)
	at java.nio.channels.FileChannelImpl.read(Unknown Source)
	at java.nio.channels.Channels$$anon$1.read(Unknown Source)
	at java.io.InputStream.read(Unknown Source)
	at os.Internals$.transfer0(Unknown Source)
	at os.Internals$.transfer(Unknown Source)
	at os.Internals$.transfer(Unknown Source)
	at os.read$bytes$.apply(Unknown Source)
	at os.read$.apply(Unknown Source)
	at os.read$.apply(Unknown Source)
	at sjsonnet.SjsonnetMainBase$.mainConfigured(Unknown Source)
...
```

tla-code-file:
```
$ echo '{a:1}' | sjsonnet-0.5.2.jar --tla-code-file f=/dev/stdin -e 'function(f)f'
{
   "a": 1
}

$ echo '{a:1}' | sjsonnet-0.5.2-native --tla-code-file f=/dev/stdin -e 'function(f)f'
Exception in thread "main" java.io.IOException: lseek failed: Illegal seek
...

# After https://github.com/databricks/sjsonnet/pull/432 neither jvm nor native works:
$ echo '{a:1}' | sjsonnet-master-36ecd2a7448d77b729786dd6336029475aba0eae-native-or-jar \
[...]

Issue #431, opened 2025-06-13
import used via --tla-code-file doesn't look for the file in the correct paths


rudo-thomas:
Imports used in a file passed to `--tla-code-file` don't seem to be able to find the imported file (in the same directory in the example below). Interestingly, the same file works as expected when `--tla-code` with an `import` is used:

```
$ cat dir/main.jsonnet                     
import './lib.libsonnet'
$ cat dir/lib.libsonnet                    
'OK'
$ jsonnet-0.20.0 dir/main.jsonnet                    
"OK"
$ jsonnet-0.20.0 --tla-code-file main=dir/main.jsonnet -e 'function(main) main'
"OK"
$ sjsonnet-0.5.2 dir/main.jsonnet 
"OK"
$ sjsonnet-0.5.2 --tla-code 'main=import "dir/main.jsonnet"' -e 'function(main) main'
"OK"
~/sjsonnet/tla-import-bug$ sjsonnet-0.5.2 --debug-importer --tla-code-file main=dir/main.jsonnet -e 'function(main) main'
[import ./lib.libsonnet] candidate /home/r/sjsonnet/tla-import-bug/lib.libsonnet
[import ./lib.libsonnet] none of the candidates exist
[read OsPath(/home/r/sjsonnet/tla-import-bug/<tla-var main>)] file does not exist or is not a file
Exception in thread "main" sjsonnet.Error: Couldn't import file: "./lib.libsonnet"
        at [Import].(<tla-var main> offset:)
        at [ValidId main].(<exec>:1:16)
```

I've confirmed that sjsonnet binary built at 0cc48fdac3dff4b842c25adb60582ce51ff8309c, i.e. with the import-related issue https://github.com/databricks/sjsonnet/issues/374 addressed, still fails the same way.

stephenamar-db:
Okay, I quickly looked.
The problem is that with tla-code-file, we basically load the file in memory early and discard the original context (as in the path of the file it came from).
It should be a straightforward fix.

PR #449, opened 2025-06-25
feat(jvm,native): Special-case "/dev/stdin" and "-" in TLAs, extVars, top-level file.

rudo-thomas:
See https://github.com/databricks/sjsonnet/issues/448.

The /dev/stdin case is a work-around until
https://github.com/scala-native/scala-native/issues/4384 is fixed.

### DIFF
--- a/sjsonnet/src-jvm-native/sjsonnet/SjsonnetMainBase.scala
+++ b/sjsonnet/src-jvm-native/sjsonnet/SjsonnetMainBase.scala
@@ -203,13 +203,19 @@ object SjsonnetMainBase {
 
     def splitMap(s: Seq[String], f: String => String) =
       s.map(split).map { case (x, v) => (x, f(v)) }
-    def readPath(v: String) = os.read(os.Path(v, wd))
+    def readPath(v: String) =
+      if (v == "-" || v == "/dev/stdin") io.Source.stdin.mkString else os.read(os.Path(v, wd))
 
     Map() ++
     splitMap(strs, v => ujson.write(v)) ++
     splitMap(strFiles, v => ujson.write(readPath(v))) ++
     splitMap(codes, identity) ++
-    splitMap(codeFiles, v => s"import @'${v.replace("'", "''")}'")
+    splitMap(
+      codeFiles,
+      v =>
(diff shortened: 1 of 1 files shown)


sjsonnet/src-jvm-native/sjsonnet/SjsonnetMainBase.scala
```diff
@@ -187,13 +187,17 @@ object SjsonnetMainBase {
...
+      if (v == "/dev/stdin") io.Source.stdin.mkString else os.read(os.Path(v, wd))
 
     Map() ++
     splitMap(strs, v => ujson.write(v)) ++
     splitMap(strFiles, v => ujson.write(readPath(v))) ++
     splitMap(codes, identity) ++
-    splitMap(codeFiles, v => s"import @'${v.replace("'", "''")}'")
+    splitMap(
+      codeFiles,
+      v => if (v == "/dev/stdin") io.Source.stdin.mkString else s"import @'${v.replace("'", "''")}'"
```
rudo-thomas: This brings back the problems of #431 for /dev/stdin.

CertainLach:
Specialcasing /dev/stdin is wrong, what you likely want is to handle all S_IFIFO files some other way

In jrsonnet, like in C jsonnet API, imports are split into two phases - resolve, and then read

In jrsonnet, resolution returns an opaque handle (trait SourcePath), and in case if FIFO file is resolved, I read the file contents during resolution, and embed it in SourcePath: https://github.com/CertainLach/jrsonnet/blob/master/crates/jrsonnet-evaluator/src/import.rs#L104-L114

It was the only working solution to properly handle things like `jsonnet --tla-file=<$(some script that generates json(net))`
Not sure how it works in c-jsonnet or go-jsonnet, but they seem to also handle that somehow through their CLI

sjsonnet/src-jvm-native/sjsonnet/SjsonnetMainBase.scala
stephenamar-db: I guess you could accept "-" too

stephenamar-db:
Maybe move this change to the native Main only. if JVM is fine with /dev/stdin, let's not make the change for it.

I'm also 100% for adding support for "-"

PR #466, opened 2025-08-02
chore: Make use of eq and nq

He-Pin:
Motivation:
eq and ne should be faster.

### DIFF
--- a/bench/src/sjsonnet/bench/MultiThreadedBenchmark.scala
+++ b/bench/src/sjsonnet/bench/MultiThreadedBenchmark.scala
@@ -57,6 +57,6 @@ class MultiThreadedBenchmark {
       catch { case e: Throwable => err = e }
     })
     pool.shutdown()
-    if (err != null) throw err
+    if (err ne null) throw err
   }
 }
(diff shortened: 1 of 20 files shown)


JoshRosen (commented):
> Motivation:
> eq and ne should be faster.

I don't think that this is the case in modern Scala versions?

Consider:

```scala
class C {
  def f(x: AnyRef): Boolean = (x == null)
  def g(x: AnyRef): Boolean = (x eq null)
}
```

In Scala 2.13.2, via `javap -c -v target/scala-2.13/classes/C.class`, these compile to identical bytecode where both use the `ifnotnull` instruction:

```
public boolean f(java.lang.Object);
    descriptor: (Ljava/lang/Object;)Z
    flags: (0x0001) ACC_PUBLIC
    Code:
      stack=1, locals=2, args_size=2
         0: aload_1
         1: ifnonnull     8
         4: iconst_1
         5: goto          9
         8: iconst_0
         9: ireturn
[...]

public boolean g(java.lang.Object);
    descriptor: (Ljava/lang/Object;)Z
    flags: (0x0001) ACC_PUBLIC
    Code:
      stack=1, locals=2, args_size=2
         0: aload_1
         1: ifnonnull     8
         4: iconst_1
         5: goto          9
         8: iconst_0
         9: ireturn
[...]
```

Similarly, I also don't see a difference in bytecode for the `ne` case. I also don't see any bytecode differences in Scala 2.12, Scala 3, etc.

Issue #439, opened 2025-06-19
Asserts when adding objects are still not working properly


stephenamar-db:
Basically,

`{assert false} + {assert true}` should throw an error due to the first assert, but it does not.

I've been trying to debug this, but there seem to be a fundamental issue with when and how asserts are executed in sjsonnet.

stephenamar-db:
0.5.2 does not enable strict assertions by default.
For instance,
```
local template = { assert self.flag };
{ a: template { flag: true }, b: template { flag: false } }
```
should throw an error, but it does not by default with 0.5.2

JoshRosen:
I dug into this and developed a fix, plus fixes for additional latent bugs that I discovered along the way; see https://github.com/databricks/sjsonnet/pull/468

PR #468, opened 2025-08-08
Fix multiple assertion related bugs (inheritance, binding of super in inherited assertions, and timing of evaluation)

JoshRosen:
This PR files multiple assertion-related bugs:

- **Inheritance**:
  -  https://github.com/databricks/sjsonnet/issues/439, where inherited object assertions would be skipped if assertions existed lower down in the object inheritance hierarchy (e.g. `{assert false} + {assert true}` would incorrectly pass because the inherited assertions would be skipped, whereas `{assert false} + {x: 2}` would fail because the child lacks its own assertions).
-  **Timing of assertion evaluation**: 
    - Per the `object-index` rule [in the jsonnet spec](https://jsonnet.org/ref/spec.html#execution), accesses of an object member are supposed to require/trigger all object assertions, but `sjsonnet` was skipping this in certain cases. For example, `{assert false} + {x: 2}).x` fails in jsonnet / go-jsonnet but erroneously succeeded in sjsonnet.
- **Binding of `super` in inherited assertions**:
  - The program `{x: 1} + {x: 2, assert super.x == 1 } + {x: 3}` is supposed succeed and print `{x: 3}` but in sjsonnet it fails because the `super.x` was resolving to `2` instead of `1`. 
  - The existing [object_invariant_plus5.jsonnet](https://github.com/databricks/sjsonnet/blob/master/sjsonnet/test/resources/go_test_suite/object_invariant_plus5.jsonnet) test case ought to have caught this issue, but it was "passing by accident" due to the assertion inheritance bug. 

This PR fixes all three issues and I think it's an overall simplification of the assertion handling logic:

- Move more of the assertion logic into `Val.Obj` and consolidate on` Obj.triggerAllAsserts()` as the sole way to trigger object assertions:
  - The old code had a discrepancy where the materializer used `obj.triggerAllAsserts(obj)` whereas some member accesses would directly invoke the `assertions `closure created by the Evaluator (which ignores inherited assertions).
  - Previously, the `asserting` flag was checked at every level of tail recursion and this was responsible for inherited assertions being skipped: once one layer evaluated assertions and set `asserting = true`, subsequent layers would bail out due to a `!asserting` check. Now we check the flag once in the "public" zero-parameter `triggerAllAsserts()` method but don't re-check it at each layer of recursion through superclasses.
  - Move access-time assertion triggering into `Obj.valueRaw`: this ensures that _all_ accesses will trigger assertions. Previously, only members created in `Evaluator` would perform these checks, causing them to be skipped for `ConstMember` and other special cases.
    - Note that this should not affect the performance of static objects: sjsonnet prohibits assertions in static objects and the new code is only invoked in paths guarded by `!static`.
- Pass an explicit `super` scope into `assertions` thunk: previously, the assertions thunk created by the evaluator called `makeNewScope(self, self.getSuper)` but `super` in an assertion refers to the superclass of the _definition_ site. We now explicitly pass in a superclass, ensuring that it resolves correctly as we walk up the inheritance chain and invoke inherited assertions.

I added a few new test cases in `EvaluatorTests`, in addition to enabling the previously-skipped test case from `go-jsonnet`.

### DIFF
--- a/sjsonnet/src/sjsonnet/Val.scala
+++ b/sjsonnet/src/sjsonnet/Val.scala
@@ -437,6 +437,7 @@ object Val {
           case null =>
             if (s == null) null else s.valueRaw(k, self, pos, addTo, addKey)
           case m =>
+            self.triggerAllAsserts()
             val vv = m.invoke(self, s, pos.fileScope, evaluator)
             val v = if (s != null && m.add) {
               s.valueRaw(k, self, pos, null, null) match {
--- a/sjsonnet/src/sjsonnet/Val.scala
+++ b/sjsonnet/src/sjsonnet/Val.scala
@@ -243,7 +243,7 @@ object Val {
    * @param static
    *   true if this object is static, false otherwise.
    * @param triggerAsserts
-   *   callback to evaluate assertions defined in the object.
+   *   callback to evaluate assertions defined in the object. Parameters are (self, super).
    * @param `super`
    *   the super object, or null if there is no super object.
    * @param valueCache
@@ -259,7 +259,7 @@ object Val {
       val pos: Position,
       private var value0: util.LinkedHashMap[String, Obj.Member],
       static: Boolean,
-      triggerAsserts: Val.Obj => Unit,
+      triggerAsserts: (Val.Obj, Val.Obj) => Unit,
       `super`: Obj,
       valueCache: util.HashMap[Any, Val] = new util.HashMap[Any, Val](),
       private var allKeys: util.LinkedHashMap[String, java.lang.Boolean] = null)
@@ -289,13 +289,15 @@ object Val {
       // infinite recursion
       if (!asserting) {
         asserting = true
-        triggerAllAsserts(this)
+        triggerAllAsserts(this, `super`)
       }
     }
 
-    @tailrec private def triggerAllAsserts(obj: Val.Obj): Unit = {
-      if (triggerAsserts != null) triggerAsserts(obj)
-      if (`super` != null) `super`.triggerAllAsserts(obj)
+    // As we walk up the superclass hierarchy, the `self` binding is unchanged
+    // but `super` climbs up the hierarchy as well.
+    @tailrec private def triggerAllAsserts(obj: Val.Obj, sup: Val.Obj): Unit = {
+      if (triggerAsserts != null) triggerAsserts(obj, sup)
+      if (sup != null) sup.triggerAllAsserts(obj, sup.getSuper)
     }
 
     def addSuper(pos: Position, lhs: Val.Obj): Val.Obj = {
--- a/sjsonnet/src/sjsonnet/Val.scala
+++ b/sjsonnet/src/sjsonnet/Val.scala
@@ -284,7 +284,11 @@ object Val {
       value0
     }
 
-    @tailrec def triggerAllAsserts(obj: Val.Obj): Unit = {
+    def triggerAllAsserts(): Unit = {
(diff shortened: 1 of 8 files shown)


sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -284,9 +284,20 @@ object Val {
...
     }
 
-    @tailrec def triggerAllAsserts(obj: Val.Obj): Unit = {
-      if (triggerAsserts != null) triggerAsserts(obj)
-      if (`super` != null) `super`.triggerAllAsserts(obj)
+    def triggerAllAsserts(): Unit = {
+      // We need to avoid asserting the same object more than once to prevent
+      // infinite recursion
+      if (!asserting) {
+        asserting = true
```
JoshRosen, line 291: An AI reviewer pointed out an interesting potential edge-case here (also present in the old code, so it's a preexisting problem):

If an assertion _fails_, we'll skip the assertion on subsequent evaluation because the flag isn't cleared after errors.

This is potentially a problem with shared parse caches: if we cache an object with failing assertions then subsequent uses of the object (from different materializations / invocations) would skip the assertions instead of failing. 

Maybe we could use a `finally` block to set `asserting = false` if `triggerAllAsserts` throws, but that would only be correct for serial single-threaded use.

Since this PR hasn't made the pre-existing problem any _worse_, I'm not currently planning to fix this in this PR but it might be worth tracking or tackling later under a "potential concurrency problems" umbrella.

JoshRosen: On further reflection, I think this theorized bug can't occur in practice:

Objects containing assertions are ineligible for static optimization. The StaticOptimizer requires `asserts == null` and `fields.forall(_.isStatic)` to convert a`MemberList` to a static Val.Obj . Since assertions make objects
  non-static, and non-static nested objects make their parents non-static, assertion-containing objects remain as `MemberList` expressions rather than cached `Val.Obj` instances, preventing the assertion state poisoning scenario that I was originally worried about.

sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -259,13 +259,13 @@ object Val {
...
       static: Boolean,
-      triggerAsserts: Val.Obj => Unit,
+      triggerAsserts: (Val.Obj, Val.Obj) => Unit,
       `super`: Obj,
       valueCache: util.HashMap[Any, Val] = new util.HashMap[Any, Val](),
       private var allKeys: util.LinkedHashMap[String, java.lang.Boolean] = null)
       extends Literal
       with Expr.ObjBody {
-    var asserting: Boolean = false
+    private[this] var asserting: Boolean = false
```
He-Pin: private this and protect this have been dropped in Scala 3/

sjsonnet/src/sjsonnet/Val.scala
He-Pin: `asserted` maybe a better name than `asserting`?

stephenamar-db:
I command you for fixing this. I could not figure it out.

sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -428,6 +439,7 @@ object Val {
           case null =>
             if (s == null) null else s.valueRaw(k, self, pos, addTo, addKey)
           case m =>
+            self.triggerAllAsserts()
```
He-Pin, line 442: ~~I think we should record the `asserting` to the `Obj`, otherwise, it seems will be re-triggered?~~

Issue #450, opened 2025-06-25
link the native builds statically (on Linux)


rudo-thomas:
Hello,

my goal is to link sjsonnnet.native statically (at least on Linux), so that the binary works anywhere.

I've had success with the following change to the build config, running `./mill show "sjsonnet.nativeStatic[3.3.6].nativeLink"` in Alpine Linux image then produced a working, statically-linked binary.

Is there a way to change the nativeLinkingOptions without modifying `build.mill`?

(Release binaries could maybe be also statically linked, but for my purposes having e.g. a command-line option to change the linking options would be enough.)

Thanks!

```diff
--- a/build.mill
+++ b/build.mill
@@ -209,6 +209,12 @@ object sjsonnet extends VersionFileModule {
     }
   }
 
+  object nativeStatic extends Cross[SjsonnetNativeStaticModule](scalaVersions)
+  trait SjsonnetNativeStaticModule extends SjsonnetNativeModule {
+    def nativeLinkingOptions = super.nativeLinkingOptions() ++
+      Seq("-static")
+  }
+
   object jvm extends Cross[SjsonnnetJvmModule](jvmScalaVersions)
   trait SjsonnnetJvmModule
       extends SjsonnetCrossModule
```

Note that linking statically on a glibc-based system was producing warnings like the following:
```
/usr/bin/ld: ./libcrypto.a(libcrypto-lib-dso_dlfcn.o): in function `dlfcn_globallookup':                
dso_dlfcn.c:(.text+0x13): warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
```
Alpine is musl-based and doesn't have this problem.

stephenamar-db:
for the last warning, I think we need to add -ldl to the link options

stephenamar-db:
I reverted, because openssl has issues with static.
I need to look into it.

PR #479, opened 2025-08-22
Fix scala native building issues and update mill

stephenamar-db:
- Update mill to 1.0.3 and force to use the JVM version instead of the native version (issues in recent ubuntu -  https://github.com/com-lihaoyi/mill/issues/5508)
- compile statically by default in Linux


stephenamar-db:
Fixes #450

Issue #473, opened 2025-08-15
port request: introduce maximum parser recursion depth


He-Pin:
Motivation:
Port: https://github.com/google/jsonnet/pull/1230

PR #483, opened 2025-08-24
feat: introduce maximum parser recursion depth

ronantakizawa:
Resolves issue #473. 

  This PR implements a configurable maximum parser recursion depth limit to prevent stack overflow from deeply
  nested Jsonnet structures, following the approach from [google/jsonnet#1230](https://github.com/google/jsonnet/pull/1230)

**Changes**
  - Added maxParserRecursionDepth: Int = 1000 parameter to Settings class
  - Added --max-parser-recursion-depth CLI argument to allow runtime configuration
  - Updated Settings flow: CLI → Config → Settings → Interpreter → CachedResolver → Parser

**Before:** 
- Parsing extremely deeply nested structures (e.g., [[[...1000+ levels...]]]) could cause native stack
  overflow and crash the interpreter.

**After:**  
- Parser gracefully throws java.lang.Exception: Parsing exceeded maximum recursion depth of 1000 instead of
  crashing. 
- Users can now configure the limit: `sjsonnet --max-parser-recursion-depth 2000 deeply_nested.jsonnet`
- Allows tuning for specific use cases while maintaining safe defaults

### DIFF
--- a/sjsonnet/src/sjsonnet/Parser.scala
+++ b/sjsonnet/src/sjsonnet/Parser.scala
@@ -267,7 +267,6 @@ class Parser(
       case (pre, fs) =>
         var remaining = fs
         def climb(minPrec: Int, current: Expr, climbDepth: Int): Expr = {
-          checkParseDepth(climbDepth)
           var result = current
           while (
             remaining.headOption match {
@@ -476,7 +475,6 @@ class Parser(
   def objinside[$: P]: P[Expr.ObjBody] = objinside(0)
 
   def objinside[$: P](currentDepth: Int): P[Expr.ObjBody] = {
-    checkParseDepth(currentDepth)
     P(
       Pos ~ member(currentDepth + 1).rep(sep = ",") ~ ",".? ~ (forspec(currentDepth + 1) ~ compspec(
         currentDepth + 1


ronantakizawa:
@He-Pin I implemented the fix, but let me know if I should remove some uses of `checkParseDepth` for performance purposes (Checking parse depth too much can slow down performance.)

sjsonnet/src/sjsonnet/Parser.scala
```diff
@@ -75,11 +76,20 @@ class Parser(
...
-    ]) {
+    ],
+    settings: Settings = Settings.default) {
   import Parser._
 
   private val fileScope = new FileScope(currentFile)
 
+  private def checkParseDepth(currentDepth: Int): Unit = {
+    if (currentDepth > settings.maxParserRecursionDepth) {
+      throw new Exception(
```
stephenamar-db, line 87: Please throw sjsonnet.Error instead

He-Pin:
@ronantakizawa I think so , we should not slow the parsing much.

stephenamar-db:
Hi @lihaoyi, if I have time of course, I would appreciate if you could take a this PR. I'm trying to figure out if this is right way to implement maximum parse recursion depth with fastparse.

He-Pin:
@ronantakizawa climb method (line 270) - protects binary operator precedence climbing recursion

This may not needed, I think the we should only limit where (), {}, [] nesting.

ronantakizawa:
@He-Pin @lihaoyi implemented fixes!

Now only `arr` and `expr` has `checkParseDepth`

PR #492, opened 2025-08-29
Re-enable handling of unicode strings in std.base64

stephenamar-db:
This is an artificial limitation in go-jsonnet and we are not bound by it.
We are perfectly capable of encoding and decoding unicode strings.

### DIFF
--- a/sjsonnet/src/sjsonnet/Std.scala
+++ b/sjsonnet/src/sjsonnet/Std.scala
@@ -1661,24 +1661,27 @@ class Std(
       rec(Materializer(value)(ev)).render
     },
     builtin("base64", "input") { (_, _, input: Val) =>
-      val b: Array[Int] = input match {
-        case Val.Str(_, value) => value.indices.map(value.codePointAt).toArray
-        case arr: Val.Arr      =>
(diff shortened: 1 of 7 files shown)


sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -1661,24 +1661,27 @@ class Std(
...
-        case Val.Str(_, value) => value.indices.map(value.codePointAt).toArray
-        case arr: Val.Arr      =>
-          if (!arr.forall(_.isInstanceOf[Val.Num])) {
-            Error.fail("Expected an array of numbers or a string")
+      input match {
+        case Val.Str(_, value) =>
+          Base64.getEncoder.encodeToString(value.getBytes("UTF-8"))
+        case arr: Val.Arr =>
+          val byteArr = new Array[Byte](arr.length)
+          for (i <- 0 until arr.length) {
```
He-Pin, line 1669: use a `while loop` instead

stephenamar-db: I want the index though, for better error message.
Why would a while loop work better?

PR #500, opened 2025-09-03
Fix handling of non-BMP Unicode strings (jsonnet strings should operate over codepoints, not UTF-16 code units)

JoshRosen:
**TL;DR**: sjsonnet string operations (length, indexing, ordering) were operating on UTF-16 code units (i.e. Java / Scala chars) but this is inconsistent with the jsonnet language reference, which [defines](https://jsonnet.org/ref/language.html#string) jsonnet strings as sequences of Unicode codepoints.

As a consequence, sjsonnet returned incorrect results (w.r.t. the go/c++ reference implementations) for several operations involving strings containing Unicode characters with codepoints outside of the [Basic Multilingual Plane](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane). 

For example, the string "🌍" consists of a single codepoint (`\u1F30D`, "earth globe europe-africa") and its UTF-16 representation is the surrogate pair `\uD83C\uDF0D`. In Java, this string has `.length == 2` but `.codePointCount == 1`. Sjsonnet was returning `std.length("🌍") == 2` instead of `1`. Similar issues affected string indexing, slicing, and sorting.

https://github.com/google/go-jsonnet/issues/600 is a related past issue in go-jsonnet.

---

This PR aims to holistically update string operations and `std` functions to operate over codepoints:

- `std.codepoint`: now handles non-BMP characters; previously, errored with a complaint about a too-long input string.
- Slicing and indexing:
  - `[]` operator
  - `std.substr`
  - `std.length`
  - `std.findSubstr`
- Iteration:
  - `std.stringChars`,
  - `std.map`
  - `std.flatMap`
- Comparison / ordering / sorting:
  - binary operations (`<`, <=`, `>`, `>=`)
  - `Evaluator.compare`
  - Materializer object field sorting
  - `std.manfiestTomlEx`, `std.manifestJson`
  - `std.objectFields`, `std.objectFieldsAll`

## Performance considerations

Since most jsonnet inputs probably contain only ASCII or Latin-1 characters, it's important that we don't regress performance _too_ much.

In some cases, Java's [compact string](https://openjdk.org/jeps/254) optimizations might hide most of the perf. impact. For example, [`String.codePointCount`](https://github.com/openjdk/jdk/blob/8d236615b7db2bd5a2a59002b79e59cf4e6a308a/src/java.base/share/classes/java/lang/String.java#L1727-L1732) has a O(1) fastpath for Latin-1 strings and similar shortcuts exist for `codePointAt`. The biggest impact is likely to be in sorting / comparison: I didn't notice huge differences in end-to-end benchmarks sorting lots of strings with long common prefixes, so I wager that this comparison performance is unlikely to be a significant issue in real practice.

## Design decision: preserving existing handling of unpaired surrogates

There are pre-existing small discrepancies in how c++ and go jsonnet handle unpaired surrogates:

- Both reject unpaired surrogates in string escapes, whereas sjsonnet permits them.
- In go-jsonnet, `std.char()` maps surrogates' codepoints to the replacement character (e.g. `std.codepoint(std.char(55296)) == 65533`), whereas c++ jsonnet preserves them as unpaired surrogates.

In this PR, I have (somewhat arbitrarily) chosen  to preserve sjsonnet's existing permissive behavior and have added new test cases to cover it.

### DIFF
--- a/sjsonnet/src/sjsonnet/Evaluator.scala
+++ b/sjsonnet/src/sjsonnet/Evaluator.scala
@@ -374,12 +374,13 @@ class Evaluator(
         v.force(int)
       case (v: Val.Str, i: Val.Num) =>
         val int = i.asPositiveInt
-        if (v.value.isEmpty) Error.fail("string bounds error: string is empty", pos)
-        val unicodeLength = v.value.codePointCount(0, v.value.length)
+        val str = v.value
+        if (str.isEmpty) Error.fail("string bounds error: string is empty", pos)
+        val unicodeLength = str.codePointCount(0, str.length)
         if (int >= unicodeLength)
           Error.fail(s"string bounds error: $int not within [0, $unicodeLength)", pos)
-        val (startUtf16, endUtf16) = Util.codePointOffsetsToStringIndices(v.value, int, int + 1)
-        Val.Str(pos, v.value.substring(startUtf16, endUtf16))
+        val (startUtf16, endUtf16) = Util.codePointOffsetsToStringIndices(str, int, int + 1)
+        Val.Str(pos, str.substring(startUtf16, endUtf16))
       case (v: Val.Obj, i: Val.Str) =>
         v.value(i.value, pos)
       case (lhs, rhs) =>
--- a/sjsonnet/src/sjsonnet/Evaluator.scala
+++ b/sjsonnet/src/sjsonnet/Evaluator.scala
@@ -379,7 +379,8 @@ class Evaluator(
         val unicodeLength = str.codePointCount(0, str.length)
         if (int >= unicodeLength)
           Error.fail(s"string bounds error: $int not within [0, $unicodeLength)", pos)
-        val (startUtf16, endUtf16) = Util.codePointOffsetsToStringIndices(str, int, int + 1)
+        val startUtf16 = if (int == 0) 0 else str.offsetByCodePoints(0, int)
+        val endUtf16 = str.offsetByCodePoints(startUtf16, 1)
         Val.Str(pos, str.substring(startUtf16, endUtf16))
       case (v: Val.Obj, i: Val.Str) =>
         v.value(i.value, pos)
--- a/sjsonnet/src/sjsonnet/Evaluator.scala
+++ b/sjsonnet/src/sjsonnet/Evaluator.scala
@@ -524,7 +524,7 @@ class Evaluator(
 
       case Expr.BinaryOp.OP_< =>
         (l, r) match {
-          case (Val.Str(_, l), Val.Str(_, r)) => Val.bool(pos, l < r)
+          case (Val.Str(_, l), Val.Str(_, r)) => Val.bool(pos, Util.compareStringsByCodepoint(l, r) < 0)
           case (Val.Num(_, l), Val.Num(_, r)) => Val.bool(pos, l < r)
           case (x: Val.Arr, y: Val.Arr)       => Val.bool(pos, compare(x, y) < 0)
           case _                              => fail()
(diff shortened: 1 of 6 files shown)


sjsonnet/src/sjsonnet/Util.scala
```diff
@@ -72,26 +72,115 @@ object Util {
...
+   * Compares two strings by Unicode codepoint values rather than UTF-16 code units.
+   * This ensures that strings with characters above U+FFFF (which require surrogate pairs
+   * in UTF-16) are compared correctly according to their Unicode codepoint values.
+   */
+  def compareStringsByCodepoint(s1: String, s2: String): Int = {
+    val n1 = s1.length
+    val n2 = s2.length
+    var i1 = 0
+    var i2 = 0
+    while (i1 < n1 && i2 < n2) {
```
JoshRosen, line 140: On reflection, I think that the character-by-character approach will be inefficient compared to [Arrays.mismatch](https://docs.oracle.com/javase/9/docs/api/java/util/Arrays.html#mismatch-char:A-char:A-) for seeking to the first unmatched position then comparing the codepoints at those indices.

That approach would clearly be correct in the case where we're comparing BMP and non-BMP characters. 

But what if we have two strings where the first mismatched characters are non-BMP characters that have the same high surrogate? I think that case is fine, too: we'd be comparing the low surrogates to each other, but they correspond to the lower bits of the final codepoints and we already know that they have the same upper bits.

**Edit**: the downside is that this results in a copy in `.toCharArray`. I'm going to forego this optimization for now.

sjsonnet/src/sjsonnet/Evaluator.scala
```diff
@@ -375,9 +375,11 @@ class Evaluator(
       case (v: Val.Str, i: Val.Num) =>
         val int = i.asPositiveInt
         if (v.value.isEmpty) Error.fail("string bounds error: string is empty", pos)
-        if (int >= v.value.length)
-          Error.fail(s"string bounds error: $int not within [0, ${v.value.length})", pos)
-        Val.Str(pos, new String(Array(v.value(int))))
+        val unicodeLength = v.value.codePointCount(0, v.value.length)
```
He-Pin: extract `v.value` to a local varaiable.

sjsonnet/src/sjsonnet/Std.scala
```diff
@@ -2102,13 +2126,16 @@ class Std(
...
-    val a = new Array[Lazy](str.length)
+    val chars = new Array[Lazy](str.codePointCount(0, str.length))
+    var charIndex = 0
     var i = 0
-    while (i < a.length) {
-      a(i) = Val.Str(pos, String.valueOf(str.charAt(i)))
-      i += 1
+    while (i < str.length) {
+      val codePoint = str.codePointAt(i)
+      chars(charIndex) = Val.Str(pos, Character.toString(codePoint))
```
He-Pin, line 2134: There is a `String.toString(codePoint)` in java 11 :(, but those two are the same anyway.

sjsonnet/src/sjsonnet/Util.scala
```diff
@@ -72,26 +72,115 @@ object Util {
...
+    } else {
       step match {
-        case 1 => s.slice(start, end)
+        case 1 =>
+          val (startUtf16, endUtf16) = codePointOffsetsToStringIndices(s, start, end)
+          s.substring(startUtf16, endUtf16)
         case _ =>
-          val range = start until end by step
-          new String(range.dropWhile(_ < 0).takeWhile(_ < s.length).map(s).toArray)
+          val result = new java.lang.StringBuilder()
```
He-Pin: init the StringBuilder's size

sjsonnet/src/sjsonnet/Util.scala
```diff
@@ -72,26 +72,115 @@ object Util {
...
+          }
+
+          // Collect every `step`th codepoint until `end`
+          var rel = 0 // relative index from start
+          while (sIdx < s.length && codepointIndex < end) {
+            val c = s.charAt(sIdx)
+            if (Character.isSurrogate(c)) {
+              // Handle surrogate pair
+              val cp = s.codePointAt(sIdx)
+              if (rel % step == 0) {
```
He-Pin: use `nextInclude += step , codepointIndex==nextInclude`  to avoid the `%`, which can be slow

JoshRosen:
> @JoshRosen, what if we record if the string contains `only ASCII or Latin-1 characters` after parsing? Would that help the later evaluation?

Based on some quick toy benchmarks with test files, I don't think that this would be worth it: the overall perf. hit doesn't seem significant and the affected code probably isn't one of the top bottlenecks in regular evaluation.

Issue #501, opened 2025-09-04
Parser parse unicode escape wrong


He-Pin:
```scala
  def escape1[$: P]: P[String] = P("\\u" ~~ CharIn("0-9a-fA-F").repX(min = 4, max = 4).!).map { s =>
    Integer.parseInt(s, 16).toChar.toString
  }
```

I think which is not right.

in my jsonpath , I'm parsing it as code below:

```scala
  private def unicodeEscape[$: P]: P[String] =
    P(NoCut(`non-surrogate`) | NoCut(`surrogate-pair`)).map {
      case (high: String, low: String) =>
        val highSurrogate = requireNonControl(Integer.parseInt(high, 16).toChar)
        val lowSurrogate = requireNonControl(Integer.parseInt(low, 16).toChar)
        new String(Array[Int](Character.toCodePoint(highSurrogate, lowSurrogate)), 0, 1)
      case str: String => requireNonControl(Integer.parseInt(str, 16).toChar).toString
      case _ => throw new IllegalArgumentException("Invalid unicode escape")
    }
```

PR #502, opened 2025-09-04
chore: Rewrite escape1 to unicodeEscape

He-Pin:
We should check the surrogate
refs: https://github.com/databricks/sjsonnet/issues/501

A follow up of https://github.com/databricks/sjsonnet/pull/500

### DIFF
--- a/sjsonnet/src/sjsonnet/Parser.scala
+++ b/sjsonnet/src/sjsonnet/Parser.scala
@@ -153,9 +153,9 @@ class Parser(
     P("\\u" ~~/ SingleChar.flatMapX { c =>
       (c: @switch) match {
         case '0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9' | 'A' | 'B' | 'C' | 'E' |
-            'F' =>
+            'F' | 'a' | 'b' | 'c' | 'e' | 'f' =>
           Pass ~~ CharPred(isHexDig).repX(min = 3, max = 3)
-        case 'D' =>
+        case 'D' | 'd' =>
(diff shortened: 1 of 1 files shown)


sjsonnet/src/sjsonnet/Parser.scala
```diff
@@ -130,9 +130,85 @@ class Parser(
...
+   * }}}
+   */
+  private def unicodeEscape[$: P]: P[String] =
+    P(NoCut(`non-surrogate`) | NoCut(`surrogate-pair`)).map {
+      case (high: String, low: String) =>
+        // NODE: jsonnet support control characters
+        val highSurrogate = Integer.parseInt(high, 16).toChar
+        val lowSurrogate = Integer.parseInt(low, 16).toChar
+        new String(Array[Int](Character.toCodePoint(highSurrogate, lowSurrogate)), 0, 1)
+      case str: String => Integer.parseInt(str, 16).toChar.toString
```
He-Pin, line 208: If removed, this can break the current behavior

stephenamar-db (approved):

sjsonnet/src/sjsonnet/Parser.scala
He-Pin: I mean If I remove this line, then it Will check the pair

stephenamar-db: Should I revert this change for now

He-Pin: no need, because it's not checking the pair too before this.

＞There are pre-existing small discrepancies in how c++ and go jsonnet handle unpaired surrogates:Both reject unpaired surrogates in string escapes, whereas sjsonnet permits them


Should we reject too？

PR #507, opened 2025-09-18
chore: Fix typo

He-Pin:


sjsonnet/test/src/sjsonnet/UnicodeHandlingTests.scala
```diff
@@ -96,26 +109,55 @@ object UnicodeHandlingTests extends TestSuite {
...
-      eval("\"\\uDC00\"") ==> ujson.Str("\uDC00") // Low surrogate alone
-      eval("\"\\uD83D\\uDE00\"") ==> ujson.Str("😀") // Valid surrogate pair
+      // Test that the parser correctly converts surrogate pairs to proper Unicode characters
+      eval("std.codepoint(\"\\uD83C\\uDF0D\")") ==> ujson.Num(127757) // Earth emoji codepoint
     }
 
-    test("stdCharPreservesRawSurrogates") {
-      // sjsonnet preserves raw surrogate codepoints (go-jsonnet would replace with U+FFFD)
-      eval("std.codepoint(std.char(55296))") ==> ujson.Num(55296) // 0xD800 high surrogate
-      eval("std.codepoint(std.char(56320))") ==> ujson.Num(56320) // 0xDC00 low surrogate
```
JoshRosen: I think that we should continue to have these tests for `std.char()` behavior with raw surrogate codepoints, especially because the two other jsonnet implementations differ in how they handle these cases.

I think it's still worthwhile to document the behavior of these two cases in the new `stdCharHandling` test.

sjsonnet/test/src/sjsonnet/UnicodeHandlingTests.scala
```diff
@@ -167,23 +209,32 @@ object UnicodeHandlingTests extends TestSuite {
...
     }
+
+    test("mixedUnicodeAndAscii") {
+      // Test mixing ASCII and Unicode characters
+      eval("'Hello ' + '🌍' + ' World'") ==> ujson.Str("Hello 🌍 World")
+      eval("std.length('ASCII🌍混合')") ==> ujson.Num(8)
+      eval("'ASCII🌍混合'[5]") ==> ujson.Str("🌍")
+      eval("'ASCII🌍混合'[6]") ==> ujson.Str("混")
+      eval("'ASCII🌍混合'[7]") ==> ujson.Str("合")
+    }
```
JoshRosen: What coverage is this new test case adding? I think that slicing and length of mixed ASCII and Unicode strings are both already covered by existing tests. Perhaps the new thing here is a testing of string concatenation?

sjsonnet/test/src/sjsonnet/UnicodeHandlingTests.scala
He-Pin: Would you like commit directly in it, thanks

sjsonnet/test/src/sjsonnet/UnicodeHandlingTests.scala
He-Pin: Testing the `indexing` but it should be covered in other place

sjsonnet/test/src/sjsonnet/UnicodeHandlingTests.scala
He-Pin: I have added it back

sjsonnet/test/src/sjsonnet/UnicodeHandlingTests.scala
```diff
@@ -101,15 +114,18 @@ object UnicodeHandlingTests extends TestSuite {
...
-    // This permissive behavior is maintained for backwards compatibility.
 
     test("unpairedSurrogatesInEscapes") {
-      // sjsonnet parses these successfully (go/C++ would reject)
-      eval("\"\\uD800\"") ==> ujson.Str("\uD800") // High surrogate alone
-      eval("\"\\uDC00\"") ==> ujson.Str("\uDC00") // Low surrogate alone
+      // sjsonnet was parses these successfully (go/C++ would reject)
+      // the new behavior will reject thesee too
+      evalErr("\"\\uD800\"").contains("Expected") // High surrogate alone
+      evalErr("\"\\uDC00\"").contains("Expected") // Low surrogate alone
```
He-Pin, line 126: @JoshRosen the new behavior will fail now

He-Pin:
```
Error: example1.jsonnet:1:1-9 example1.jsonnet:1:1-9 Truncated unicode surrogate pair escape sequence in string literal.

"\uD800"

```

in jsonnet.org

Issue #392, opened 2025-05-16
Stackoverflow while materializing


He-Pin:
The Materializer is using JVM's stack too, maybe we can introduce a tailrec implementation too?
refs: https://github.com/CertainLach/jrsonnet/issues/115

Maybe using : scala.util.control.TailCalls.TailRec ?

```jsonnet
local reduce(func, array, default) =
//  assert std.isFunction(func) : error 'reduce(func=...) should be a `function`';
//  assert std.isArray(array) : error 'reduce(array=...) should be an `array`';
  //  assert std.length(array) > 0 : error 'reduce(array=...) should be a non-0 `array`';
  local aux(func, array, current, index) =
    if index < 0 then
      current
    else
      aux(func, array, func(array[index], current), index - 1) tailstrict;
  local length = std.length(array);
  if length == 0 then
    default
  else
    aux(func, array, array[length - 1], length - 2);

local visit(
  value,
  visitor,
  ctx,
  ctx_reducer,
) =
  local new_item = if std.isObject(value) then
    local merged = {
      [prop]: visit(value[prop], visitor, ctx, ctx_reducer)
      for prop in std.objectFields(value)
    };

    local new_item = {
      [prop]: merged[prop].item
      for prop in std.objectFields(merged)
    };
    local new_ctx = reduce(ctx_reducer, std.map(
      function(item) item.ctx,
      std.objectValues(merged)
    ), ctx);

    {
      item: new_item,
      ctx: new_ctx,
    }
  else if std.isArray(value) then
    local merged = std.map(
      function(item) visit(item, visitor, ctx, ctx_reducer),
      value,
    );

    // If we had a way to tee collections, that would be great... :(
    local new_item = std.map(function(item) item.item, merged);
    local all_ctx = std.map(function(item) item.ctx, merged);

    // Reduce all of the contexts
    local new_ctx = reduce(ctx_reducer, all_ctx, ctx);

    {
      item: new_item,
      ctx: new_ctx,
    }
  else
    {
      item: value,
      ctx: ctx,
    };
  local replacement = visitor(new_item.ctx, new_item.item);
  if replacement != null then
    local new_ctx = ctx_reducer(replacement.ctx, new_item.ctx);

    {
      item: replacement.item,
      ctx: new_ctx,
    }
  else
    new_item;


local routes = {
  "/inc/users/whoami": {
    "get": {
      "responses": {
        "200": {
          "content": {
            "application/json": {
              "schema": {
                "properties": {
[...]

PR #539, opened 2025-10-30
Remove common reasons for materialization to fail due to stack overflows

stephenamar-db:
This rewrite Val.Obj addSuper and gatherKeys to be iterative instead of recursive.

Fixes https://github.com/databricks/sjsonnet/issues/392

### DIFF
--- a/sjsonnet/src/sjsonnet/Val.scala
+++ b/sjsonnet/src/sjsonnet/Val.scala
@@ -266,8 +266,8 @@ object Val {
   final class Obj(
       val pos: Position,
       private var value0: util.LinkedHashMap[String, Obj.Member],
-      static: Boolean,
-      triggerAsserts: (Val.Obj, Val.Obj) => Unit,
+      private val static: Boolean,
+      private val triggerAsserts: (Val.Obj, Val.Obj) => Unit,
       `super`: Obj,
       valueCache: util.HashMap[Any, Val] = new util.HashMap[Any, Val](),
       private var allKeys: util.LinkedHashMap[String, java.lang.Boolean] = null)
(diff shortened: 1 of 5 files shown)


sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -313,24 +313,41 @@ object Val {
...
-        case x    => new Val.Obj(pos, getValue0, false, triggerAsserts, x.addSuper(pos, lhs))
+      val objs = mutable.ArrayBuffer(this)
+      var current = this
+      while (current.getSuper != null) {
+        objs += current.getSuper
+        current = current.getSuper
       }
+
+      current = lhs
+      for (s <- objs.reverse) {
```
He-Pin, line 324: maybe just use `reverseIterator`

sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -313,24 +313,41 @@ object Val {
...
     private def gatherKeys(mapping: util.LinkedHashMap[String, java.lang.Boolean]): Unit = {
-      if (static) mapping.putAll(allKeys)
-      else {
-        if (`super` != null) `super`.gatherKeys(mapping)
-        getValue0.forEach { (k, m) =>
-          val vis = m.visibility
-          if (!mapping.containsKey(k)) mapping.put(k, vis == Visibility.Hidden)
-          else if (vis == Visibility.Hidden) mapping.put(k, true)
-          else if (vis == Visibility.Unhide) mapping.put(k, false)
+      val objs = mutable.ArrayBuffer(this)
```
He-Pin, line 334: set `sizeHint` here to avoid the resizing

sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -313,24 +313,41 @@ object Val {
     }
 
     def addSuper(pos: Position, lhs: Val.Obj): Val.Obj = {
-      `super` match {
-        case null => new Val.Obj(pos, getValue0, false, triggerAsserts, lhs)
-        case x    => new Val.Obj(pos, getValue0, false, triggerAsserts, x.addSuper(pos, lhs))
+      val objs = mutable.ArrayBuffer(this)
```
He-Pin, line 316: set `sizeHint`

sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -313,24 +313,41 @@ object Val {
...
+      var current = this
+      while (current.getSuper != null) {
+        objs += current.getSuper
+        current = current.getSuper
```
He-Pin, line 320: Save `current.getSuper` to a local variable to avoid invoking 3 times in a loop

sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -313,24 +313,41 @@ object Val {
...
       }
+
+      current = lhs
+      for (s <- objs.reverse) {
+        current = new Val.Obj(s.pos, s.getValue0, false, s.triggerAsserts, current)
```
He-Pin, line 325: can we change `super` to a var, and then we can update the `super` field only

sjsonnet/src/sjsonnet/Val.scala
stephenamar-db: The problem is that I don't know the size in advance.
Maybe I can pick a good average (like 10) to start, but there will be resizes

sjsonnet/src/sjsonnet/Val.scala
stephenamar-db: not thrilled with making Val.Obj mutable. We would need to reset the underlying cache too.

sjsonnet/src/sjsonnet/Val.scala
stephenamar-db: it's equivalent. Also using foldRight would lead to recursion

PR #542, opened 2025-10-31
Regression Benchmark suite

stephenamar-db:
First version of the sjsonnet regression benchmark suite.

to run: `./mill bench.regressionBenchs`

It takes a while though.


He-Pin:
`bench.runJmh`?

PR #554, opened 2025-11-18
Fixed prettyIndex for one-line files, use binarySearch

hauserx:
Pretty index was failing with following exception for files with one line and no `\n` at the end. The `lines` array was empty:
```
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 0
```
There are actually some generated files in our code based that follow this pattern.

Also, for expressions in the very first line it was creating values with negative columns, like:
```
x.libsonnet:2:-63
```

Fixed the issue and added tests, also for edge case with completely empty file.

While changing it also optimized searching for line - previously `indexWhere` was used which is linear search. Replaced with binary search. It does not matter to much usually, but actually was causing slowdown of profiler code that gets locations of all evaluated expressions - this will be added in a subsequent PR.

### DIFF
--- a/sjsonnet/src/sjsonnet/Importer.scala
+++ b/sjsonnet/src/sjsonnet/Importer.scala
@@ -74,22 +74,8 @@ final case class FileParserInput(file: File) extends ParserInput {
     lines.toArray
   }
 
-  def prettyIndex(index: Int): String = {
-    // The binary search returns >= 0 if founds exact match. The 0-indexed line
-    // number will be exactly this number in such case.
-    //
-    // If value is not found, then searchResult = (-insertionPoint - 1) is
-    // returned, where insertionPoint is the index where the element would have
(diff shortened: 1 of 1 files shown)


hauserx:
@He-Pin @stephenamar-db PTAL

sjsonnet/src/sjsonnet/Importer.scala
```diff
@@ -57,7 +57,7 @@ final case class FileParserInput(file: File) extends ParserInput {
   override def checkTraceable(): Unit = {}
 
   private lazy val lineNumberLookup: Array[Int] = {
-    val lines = mutable.ArrayBuffer[Int]()
+    val lines = mutable.ArrayBuffer[Int](0)
```
He-Pin, line 60: Why change it to `0` , it will always later be resiezd then

PR #555, opened 2025-11-21
Improve performance of std.stripChars and std.trim functions

stephenamar-db:
Remove regular expressions usage - it's an overkill.

Before:

```
[1-113] Java HotSpot(TM) 64-Bit Server VM warning: -XX:ThreadPriorityPolicy=1 may require system level permission, e.g., being the root user. If the necessary permission is not possessed, changes to priority will be silently ignored.
0.943 ms/op
0.761 ms/op
```

After:
```
[1-113] Java HotSpot(TM) 64-Bit Server VM warning: -XX:ThreadPriorityPolicy=1 may require system level permission, e.g., being the root user. If the necessary permission is not possessed, changes to priority will be silently ignored.
0.790 ms/op
0.674 ms/op
```


sjsonnet/src/sjsonnet/stdlib/StringModule.scala
```diff
@@ -105,110 +105,77 @@ object StringModule extends AbstractFunctionModule {
...
-      private val leftPattern = Platform.getPatternFromCache(getLeadingPattern(chars))
-      private val rightPattern = Platform.getPatternFromCache(getTrailingPattern(chars))
-
-      def evalRhs(str: Lazy, ev: EvalScope, pos: Position): Val = {
-        var s = str.force.asString
-        if (right) s = rightPattern.matcher(s).replaceAll("")
-        if (left) s = leftPattern.matcher(s).replaceAll("")
-        Val.Str(pos, s)
+    def codePointsSet(str: String): Set[Int] = {
+      val chars = Set.newBuilder[Int]
```
He-Pin, line 108: with a predefined size to avoid resize

sjsonnet/src/sjsonnet/stdlib/StringModule.scala
```diff
@@ -105,110 +105,77 @@ object StringModule extends AbstractFunctionModule {
...
+      for (i <- 0 until str.length) {
+        val codePoint = str.codePointAt(i)
+        chars += codePoint
```
He-Pin, line 113: Use a while loop, IIRC, this is not been rewrite to while in scala 3

sjsonnet/src/sjsonnet/stdlib/StringModule.scala
He-Pin: Is there an IntSet to avoid unboxing

stephenamar-db: we don't know the size in advance.

stephenamar-db: let me use the chars size as a hint.
but no primitive set sadly.

sjsonnet/src/sjsonnet/stdlib/StringModule.scala
stephenamar-db: I'm not familiar with the specificities of scala3. Can you point to something that explains why while loops are better for here?

He-Pin: In scala2, the compiler Will rewrite this to a while loop, but in scala 3, because the optimization is not been ported, then this will be a Range.foreach

PR #564, opened 2025-12-09
chore: Use ArrayBuilder to build the array

He-Pin:
The ArrayBuffer#toArray will do another copy.

### DIFF
--- a/sjsonnet/src/sjsonnet/stdlib/SetModule.scala
+++ b/sjsonnet/src/sjsonnet/stdlib/SetModule.scala
@@ -62,25 +62,28 @@ object SetModule extends AbstractFunctionModule {
       return arr
     }
 
-    val out = new mutable.ArrayBuffer[Lazy]
+    val out = new mutable.ArrayBuilder.ofRef[Lazy]
+    // Set a reasonable size hint - in the worst case (no duplicates), we'll need arrValue.length elements
+    out.sizeHint(arrValue.length)
     for (v <- arrValue) {
-      if (out.isEmpty) {
-        out.append(v)
+      val outResult = out.result()
+      if (outResult.length == 0) {
+        out.+=(v)
       } else if (keyF.isInstanceOf[Val.False]) {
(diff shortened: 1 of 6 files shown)


sjsonnet/src/sjsonnet/stdlib/StringModule.scala
```diff
@@ -205,19 +205,21 @@ object StringModule extends AbstractFunctionModule {
...
               case v: Val.Arr  =>
-                if (added) out.appendAll(sep.asLazyArray)
+                if (added) out ++= sep.asLazyArray
                 added = true
-                out.appendAll(v.asLazyArray)
+                out ++= v.asLazyArray
               case x => Error.fail("Cannot join " + x.prettyName)
             }
           }
-          Val.Arr(pos, out.toArray)
```
He-Pin, line 220: I checked the ArrayBuffer#toArray will do an ArrayCopy

sjsonnet/src/sjsonnet/stdlib/SetModule.scala
```diff
@@ -62,25 +62,28 @@ object SetModule extends AbstractFunctionModule {
...
     }
 
-    val out = new mutable.ArrayBuffer[Lazy]
+    val out = new mutable.ArrayBuilder.ofRef[Lazy]
+    // Set a reasonable size hint - in the worst case (no duplicates), we'll need arrValue.length elements
+    out.sizeHint(arrValue.length)
     for (v <- arrValue) {
-      if (out.isEmpty) {
-        out.append(v)
+      val outResult = out.result()
```
JoshRosen, line 69: I think that this change is introducing an O(n^2) bug because now we're making a fresh copy of the array on every loop iteration.

I've opened https://github.com/databricks/sjsonnet/pull/575 with a suggested fix.

He-Pin: Yes, we need just use the `length` to fix this.

PR #566, opened 2025-12-11
Fix performance regression in array comprehension

stephenamar-db:
This fixes the performance regression of bench/resources/go_suite/comparison2.jsonnet


sjsonnet/src/sjsonnet/Evaluator.scala
```diff
@@ -811,16 +811,7 @@ class Evaluator(
...
-      if (scopes.length > 0) {
-        val firstArrSize = visitExpr(expr)(scopes(0)) match {
-          case a: Val.Arr => a.length
-          case _          => 1
-        }
-        // Use heuristic estimation: first array size * scopes count * 2 for safety margin
-        newScopes.sizeHint(math.max(16, firstArrSize * scopes.length * 2))
-      }
-
+      newScopes.sizeHint(math.max(16, scopes.length * 4))
```
He-Pin, line 814: Thanks,seems it's heavier than it looks.

PR #571, opened 2025-12-16
Add support for custom Vals with materialization logic

niegrzybkowski:
We would like to add better support for custom `Val` types. These can already be created by extending `Val.Literal`, but can't be materialized, since `Materializer` doesn't know what to do with them. 
This PR allows specifying the materialization logic to use with custom `Val`s.

### DIFF
--- a/sjsonnet/src/sjsonnet/Materializer.scala
+++ b/sjsonnet/src/sjsonnet/Materializer.scala
@@ -151,6 +151,7 @@ object Materializer extends Materializer {
 
   /**
    * Trait for providing custom materialization logic to the Materializer.
+   * @since 1.0.0
    */
   trait Materializable {
     def materialize[T](visitor: Visitor[T, T])(implicit evaluator: EvalScope): T


sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -93,6 +94,13 @@ object Val {
     override def asBoolean: Boolean = this.isInstanceOf[True]
   }
 
+  /**
+   * Base trait for user-defined [[Val]]s with custom materialization logic.
+   */
+  trait Custom extends Literal {
+    def materialize[T](visitor: Visitor[T, T])(implicit evaluator: EvalScope): T
```
He-Pin: I think you should add this method to Visitor, but not `CustomVal`.

He-Pin: Another question is, why can't you express your `Custom` Val to a `Val.*`? Seems you want to avoid the conversation?

It would be nice if you shared a use case demo.

niegrzybkowski: I'm not sure what you mean with the upickle Visitor. The underlying issue we have is that Materializer doesn't know how to handle custom sjsonnet Val.Literals, and we can't sensibly override it since the static Materializer object is used in quite a few places.

For the use case, the code is private, but we're making a jsonnet library which uses custom types that are lossy when converted to JSON values, so we would like to avoid a Val-CustomVal round trip unless absolutely necessary (like when passing them to a stdlib function, or materializing). See the "test custom Val usage" test for a very simplified version.

He-Pin: How about extract `materialize` to `Materializeable` trait?

niegrzybkowski: Sure, that will work

sjsonnet/src/sjsonnet/Materializer.scala
```diff
@@ -147,4 +148,11 @@ object Materializer extends Materializer {
 
   final val emptyStringArray = new Array[String](0)
   final val emptyLazyArray = new Array[Lazy](0)
+
+  /**
+   * Trait for providing custom materialization logic to the Materializer.
```
He-Pin, line 153:
```suggestion
   * Trait for providing custom materialization logic to the Materializer.
   * @since 1.0.0
```

stephenamar-db (approved):

PR #588, opened 2026-01-31
fix: return null if parseYaml does not yield any result, port https://github.com/google/go-jsonnet/pull/841

He-Pin:
port https://github.com/google/go-jsonnet/pull/841


sjsonnet/test/src/sjsonnet/ParseYamlTests.scala
```diff
@@ -10,7 +10,7 @@ object ParseYamlTests extends TestSuite {
       eval("std.parseYaml('- foo: bar')") ==> ujson.Value("""[{"foo":"bar"}]""")
     }
     test {
-      eval("std.parseYaml('')") ==> ujson.Value("""{}""")
+      eval("std.parseYaml('')") ==> ujson.Value("""null""")
```
He-Pin, line 13: It's now `null` instead of `{}`

stephenamar-db (approved):

PR #590, opened 2026-01-31
chore: fmt to keep the error prefix the same as google/jsonnet

He-Pin:
To make the golden sync easier


He-Pin:
Hope one day we can align it.

PR #676, opened 2026-04-04
perf: pre-cached indent arrays for bulk newline+spaces

He-Pin:
## Motivation

The JSON renderer generates indentation strings (newline + spaces) on every nested element. For deeply nested Jsonnet output, `Renderer.visitKey` and `visitEnd` repeatedly construct identical indent strings. The current implementation calls `elemBuilder.append('\n')` followed by a `while` loop appending spaces — this is O(depth) per indent operation.

## Key Design Decision

Pre-cache indent strings (newline + spaces) in a companion object array up to depth 64 (`MaxCachedDepth`). For depths ≤64, indent operations become a single array lookup + bulk write. For depths >64 (rare in practice), fall through to the original loop.

## Modification

**`sjsonnet/src/sjsonnet/Renderer.scala`**:
- Added companion object with `MaxCachedDepth = 64` constant and `indentCache: Array[Array[Char]]`
- Cache stores pre-computed `"\n" + " " * (depth * indent)` as char arrays for depths 0–64
- `flushBuffer()` fast path: when depth ≤ `MaxCachedDepth`, uses `elemBuilder.appendAll(cachedArray, len)` instead of character-by-character loop
- Original loop preserved as fallback for depths > 64

### Note

The indentation cache optimization primarily benefits:
1. **Deeply nested JSON output** — common in Jsonnet configurations (Kubernetes manifests, CI configs)
2. **`std.manifestJsonEx`** — uses indentation for pretty-printing
3. **Scala Native** — no JIT to optimize the loop; pre-cached arrays enable `System.arraycopy`

## Analysis

- **Memory**: One-time allocation of 64 char arrays (total ~2KB) — negligible.
- **Thread safety**: Cache is in a companion object, initialized once. Arrays are read-only after initialization.
- **Threshold**: 64 levels covers virtually all real-world Jsonnet output (even deeply nested Kubernetes manifests rarely exceed 20 levels).

## References

- `upickle.core.CharBuilder.appendAll(char[], int)` for bulk writes
- Original character-by-character indent loop in `Renderer.flushBuffer`

## Result

Pre-cached indent arrays eliminate per-character overhead for nested JSON rendering. No regressions. Benefits deeply nested output and Scala Native.

### DIFF
--- a/sjsonnet/src/sjsonnet/Renderer.scala
+++ b/sjsonnet/src/sjsonnet/Renderer.scala
@@ -4,6 +4,15 @@ import java.io.{StringWriter, Writer}
 
 import upickle.core.{ArrVisitor, ObjVisitor}
 
+object Renderer {
+
+  /**
+   * Maximum nesting depth for pre-computed indent arrays. Depths beyond this fall back to
+   * per-character rendering. 16 covers the vast majority of real-world Jsonnet output.
+   */
+  final val MaxCachedDepth = 16
+}
+
 /**
  * Custom JSON renderer to try and match the behavior of google/jsonnet's render:
  *
@@ -12,6 +21,7 @@ import upickle.core.{ArrVisitor, ObjVisitor}
  */
 class Renderer(out: Writer = new java.io.StringWriter(), indent: Int = -1)
     extends BaseCharRenderer(out, indent) {
+  import Renderer.MaxCachedDepth
   var newlineBuffered = false
   override def visitFloat64(d: Double, index: Int): Writer = {
     val s = RenderUtils.renderDouble(d)
@@ -20,6 +30,23 @@ class Renderer(out: Writer = new java.io.StringWriter(), indent: Int = -1)
     flushCharBuilder()
     out
   }
+
+  // Pre-computed indent arrays: indentCache(d) = '\n' + indent*d spaces
+  private val indentCache: Array[Array[Char]] =
+    if (indent <= 0) null
+    else {
+      val arr = new Array[Array[Char]](MaxCachedDepth)
+      var d = 0
+      while (d < MaxCachedDepth) {
+        val spaces = indent * d
+        val buf = Array.fill(spaces + 1)(' ')
+        buf(0) = '\n'
+        arr(d) = buf
+        d += 1
+      }
+      arr
+    }
+
   override def flushBuffer(): Unit = {
     if (commaBuffered) {
       elemBuilder.append(',')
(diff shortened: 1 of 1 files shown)


sjsonnet/src/sjsonnet/Renderer.scala
```diff
@@ -13,26 +13,92 @@ import upickle.core.{ArrVisitor, ObjVisitor}
...
   }
+
+  private def appendLong(l: Long): Unit = {
+    if (l == 0L) {
+      elemBuilder.append('0')
+      return
+    }
+    // Handle Long.MIN_VALUE specially: -Long.MIN_VALUE overflows back to Long.MIN_VALUE
+    if (l == Long.MinValue) {
+      val s = "-9223372036854775808"
```
stephenamar-db: in that case, the length is a fixed number? just store the length in a static value.

stephenamar-db (changes requested):

He-Pin:
Good catch — extracted the magic `16` into a named constant `Renderer.MaxCachedDepth` in a new companion object. The comparison now reads `depth < MaxCachedDepth` instead of `depth < indentCache.length`.

Note that the indent cache **content** is instance-specific (depends on the `indent` constructor parameter — commonly 2, 3, or 4), but the **size** (16 depth levels) is a fixed constant shared across all instances.

sjsonnet/src/sjsonnet/Renderer.scala
```diff
@@ -20,19 +30,42 @@ class Renderer(out: Writer = new java.io.StringWriter(), indent: Int = -1)
...
+
+  // Pre-computed indent arrays: indentCache(d) = '\n' + indent*d spaces
+  private val indentCache: Array[Array[Char]] =
+    if (indent <= 0) null
+    else {
+      val arr = new Array[Array[Char]](MaxCachedDepth)
+      var d = 0
+      while (d < MaxCachedDepth) {
+        val spaces = indent * d
+        val buf = new Array[Char](spaces + 1)
```
stephenamar-db: would'nt
```
  val buf = Array.fill(spaces + 1) { ' ' }
  buf(0) = '\n'
```
be more terse?

He-Pin: Good call — applied `Array.fill(spaces + 1)(' ')` + `buf(0) = '\n'` in the latest push. Much cleaner, thanks!

He-Pin:

PR #684, opened 2026-04-05
perf: fast-path Num stringify in OP_+ string concatenation

He-Pin:
## Motivation

When Jsonnet concatenates a number with a string via `+`, the evaluator goes through `Materializer.stringify()` which handles all `Val` types via pattern matching. For the common case of `Num + Str` or `Str + Num`, this generic dispatch adds unnecessary overhead — we already know the value is a `Val.Num` from the enclosing match.

## Key Design Decision

Add specialized `(Val.Num, Val.Str)` and `(Val.Str, Val.Num)` match cases before the generic `(Val.Str, r)` / `(l, Val.Str)` fallbacks. These cases call `RenderUtils.renderDouble(n.asDouble)` directly, bypassing `Materializer.stringify`'s type dispatch.

## Modification

**`sjsonnet/src/sjsonnet/Evaluator.scala`** — `OP_+` handler:
- Added two new cases in the `(l, r) match` block:
  ```scala
  case (n: Val.Num, Val.Str(_, r)) => Val.Str(pos, RenderUtils.renderDouble(n.asDouble) + r)
  case (Val.Str(_, l), n: Val.Num) => Val.Str(pos, l + RenderUtils.renderDouble(n.asDouble))
  ```
- These are placed after `(Val.Num, Val.Num)` and `(Val.Str, Val.Str)` but before the generic stringify fallbacks.

| large_string_template | 1.784 | 1.793 | ±0% |

### Note

On the JVM, the JIT compiler inlines `Materializer.stringify` through profile-guided optimization, reducing the benefit of this explicit fast path. The primary value is:
1. **Scala Native**: No JIT — avoiding virtual dispatch for the hot `Num + Str` path is significant.
2. **Code clarity**: Makes the common `Num + Str` concatenation path explicit in the source.

## Analysis

- **Correctness**: `RenderUtils.renderDouble` is the same function called by `Materializer.stringify` for `Val.Num` — semantics are identical.
- **Pattern order**: New cases come after the most common `(Num, Num)` and `(Str, Str)` cases, so they don't add overhead to those hot paths.

## References

- `Materializer.stringify` → `RenderUtils.renderDouble` for `Val.Num`
- `RenderUtils.renderDouble` handles integer-valued doubles (e.g., `42.0` → `"42"`)

## Result

Type-specialized fast path for number-string concatenation. No regressions. Benefits Scala Native and improves code explicitness.

### DIFF
--- a/sjsonnet/src/sjsonnet/Evaluator.scala
+++ b/sjsonnet/src/sjsonnet/Evaluator.scala
@@ -879,11 +879,13 @@ class Evaluator(
         (l, r) match {
           case (Val.Num(_, l), Val.Num(_, r)) => Val.cachedNum(pos, l + r)
           case (Val.Str(_, l), Val.Str(_, r)) => Val.Str(pos, l + r)
-          case (Val.Str(_, l), r)             => Val.Str(pos, l + Materializer.stringify(r))
-          case (l, Val.Str(_, r))             => Val.Str(pos, Materializer.stringify(l) + r)
-          case (l: Val.Obj, r: Val.Obj)       => r.addSuper(pos, l)
-          case (l: Val.Arr, r: Val.Arr)       => l.concat(pos, r)
-          case _                              => failBinOp(l, e.op, r, pos)
+          case (n: Val.Num, Val.Str(_, r)) => Val.Str(pos, RenderUtils.renderDouble(n.asDouble) + r)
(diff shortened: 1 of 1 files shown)


stephenamar-db (approved):

He-Pin:
Fixed — formatting has been corrected after the rebase. `scalafmt` run included in the latest force-push.

PR #685, opened 2026-04-05
perf: direct long-to-chars rendering in visitFloat64

He-Pin:
## Motivation

When rendering integer-valued doubles (e.g., `42.0`), `visitFloat64` currently calls `i.toString` which allocates a `String` object, then passes it through `visitFloat64StringParts` for character-by-character processing. For large arrays of numbers (common in Jsonnet), this creates millions of short-lived `String` allocations.

## Key Design Decision

Add a `writeLongDirect` method that converts a `Long` directly into the `CharBuilder`'s backing array without any intermediate `String` allocation. Uses a digit-pair lookup table for two-digits-at-a-time conversion (a well-known optimization from JDK's `Integer.getChars`).

## Modification

**`sjsonnet/src/sjsonnet/BaseCharRenderer.scala`**:
- Added `writeLongDirect(v: Long)` private method:
  - Special-cases `0` and `Long.MinValue` (negation overflow)
  - Counts digits, then writes right-to-left using digit-pair tables
  - Directly writes into `elemBuilder.arr` via `ensureLength` + position update
- Added companion object `BaseCharRenderer` with `DIGIT_TENS` and `DIGIT_ONES` lookup tables (100 entries each)
- Changed `visitFloat64`: integer path now calls `writeLongDirect(i)` instead of `visitFloat64StringParts(i.toString, ...)`

**Test**: Added `new_test_suite/large_integer_rendering.jsonnet` — verifies correct rendering of boundary values (0, negatives, large longs, Long.MinValue).

Pending benchmark data.

### Expected Impact

Primarily benefits benchmarks with heavy numeric rendering:
- `realistic2` (large JSON with many numbers)
- `base64DecodeBytes` (byte arrays rendered as numbers)
- Any workload generating numeric arrays

## Analysis

- **Correctness**: Handles all edge cases — zero, negative, Long.MinValue overflow, single/multi-digit numbers.
- **Digit-pair table**: Converts two digits per iteration (100x div/mod instead of 10x), halving the number of divide operations.
- **Zero allocations**: No `String` or `StringBuilder` — writes directly into the backing `char[]`.
- **Regression test**: `large_integer_rendering.jsonnet` covers boundary values including `Long.MinValue`.

## References

- JDK's `Integer.getChars` uses the same digit-pair technique
- `BaseCharRenderer.DIGIT_TENS` / `DIGIT_ONES` — static lookup tables

## Result

Zero-allocation integer rendering with digit-pair optimization. Eliminates `Long.toString` allocation overhead. Draft PR pending benchmark data.


He-Pin:

PR #732, opened 2026-04-10
perf: optimize array comparison with pre-Eval reference equality

He-Pin:
## Motivation

When arrays share element references (e.g., `std.range(1,N)` followed by `long_array + [x] < long_array + [y]`), both `compare()` and `equal()` call `.value` on every `Eval` element to force lazy thunks. On Scala Native (AOT compilation), each virtual dispatch costs ~4-5ns. For two 1M-element arrays sharing 999,999 references, this amounts to ~10ms of unnecessary vtable overhead.

## Key Design Decision

Three-layer optimization that preserves error semantics:

1. **`Val.Arr.value(i)` instanceof fast-path**: `if (e.isInstanceOf[Val]) e.asInstanceOf[Val] else e.value` — avoids vtable dispatch for strict values globally
2. **Pre-Eval reference equality in `compare()`**: Check `ex eq ey` *before* forcing. For shared strict `Val` references, skip entirely. For shared lazy thunks, force once to preserve error observation.
3. **Pre-Eval reference equality in `equal()`**: Mirror the `compare()` structure — separate paths for shared strict (skip) and shared lazy (force once).

## Modification

- **`Val.scala`**: `Arr.value(i)` — added `isInstanceOf[Val]` fast-path
- **`Evaluator.scala`**: `compare()` array branch — pre-Eval reference equality with lazy error preservation
- **`Evaluator.scala`**: `equal()` array branch — mirrored `compare()` optimization structure
- **New test**: `array_compare_shared_elements.jsonnet` — regression test covering shared strict/lazy element comparison semantics

### JMH (JVM, Scala 3.3.7)

|-----------|:-:|:-:|:-:|
| comparison | 21.344 | 20.848 | **-2.3%** |
| comparison2 | ~35.9 | ~35.8 | neutral |
| realistic2 | 60.993 | 61.0 | neutral |
| reverse | 8.464 | 8.5 | neutral |

Full JMH regression suite (35 benchmarks): **no regressions detected**.

| Benchmark | Before (ms) | After (ms) | Change | jrsonnet (ms) | Gap |
|-----------|:-:|:-:|:-:|:-:|:-:|
| comparison | 30.0 | 28.6 | **-4.7%** | 12.8 | 2.23x |
| comparison2 | 84.9 | 84.9 | neutral | 222.4 | **WIN 2.62x** |
| realistic2 | 255.2 | 255.0 | neutral | ~101 | 2.52x |
| reverse | 36.6 | 36.6 | neutral | ~22.6 | 1.62x |

## Analysis

- **JVM improvement is modest (-2.3%)** because the JIT already devirtualizes `Val.value` in hot loops
- **Native improvement is larger (-4.7%)** because AOT compilation cannot eliminate virtual dispatch — the `isInstanceOf` check provides real savings
- **Theoretical limit**: For 1M shared elements at 4GHz, saving two vtable dispatches per element ≈ 1.2ms, consistent with measured 1.4ms improvement
- **No regressions** across the full benchmark suite — the `isInstanceOf` overhead for non-shared elements is negligible (branch prediction + zero-cost on JVM)

## References

- Benchmark suite: `jrsonnet/docs/benchmarks.adoc`
- Comparison benchmark: `bench/resources/go_suite/comparison.jsonnet`

## Result

Closes the comparison benchmark gap from 2.34x → 2.23x vs jrsonnet. No semantic changes — all 141 test suites pass, including new regression test for shared-element comparison correctness.


He-Pin:
Closing: benchmark testing confirmed this optimization provides no meaningful improvement. Both JMH (20.783ms vs 20.139ms baseline, slightly worse) and native hyperfine (27.1ms → 26.0ms, ~4% within noise) showed no significant gain for comparison_for_array. The benchmark is allocation-dominated (std.range creates 1M Val.Num objects), and pre-Eval reference equality checks don't address that bottleneck.

stephenamar-db:
Could we move std.range() to be fully lazy?
As in a special subclass of Arr that would compute the value on demand?

PR #735, opened 2026-04-10
perf: direct JSON renderer bypassing Visitor pattern for JVM CLI output

He-Pin:
## Motivation

The upickle `Visitor`/`ObjVisitor`/`ArrVisitor` pattern is the primary materialization bottleneck on JVM. On `realistic2`, the Visitor-based rendering dispatches ~3.3 million virtual method calls (per-element `visitValue`, `visitKey`, `visitString`, etc.). The JIT cannot fully devirtualize these because multiple `Visitor` implementations exist in the classpath.

## Key Design Decisions

1. **JVM-only optimization**: `Platform.useDirectRenderer` is a `final val` — `true` on JVM, `false` on Native. Native's LLVM LTO already devirtualizes the Visitor pattern at link time, making the direct renderer counterproductive there (measured: 6.8% regression → neutral after flag).

2. **Deep nesting safety**: Falls back to the Materializer's hybrid recursive/iterative path (ArrayDeque-based `materializeStackless`) for subtrees beyond `recursiveDepthLimit` (128). This prevents stack overflow on deeply nested structures while keeping the fast path for normal depths.

3. **Materializable compatibility**: Falls back to Visitor-based `Renderer` for custom `Materializer.Materializable` values, preserving the embedding API. The fallback `Renderer` overrides `flushCharBuilder()` with threshold 0 to prevent silent data loss (BaseCharRenderer uses threshold 1000 at depth ≥ 1).

4. **Output equivalence**: Produces byte-identical JSON output to `Renderer` for all indent values (minified, indent=0, indent>0). Empty containers render as `{ }` / `[ ]` matching Renderer behavior.

## Modification

- **New file**: `DirectJsonRenderer.scala` (~320 lines) — `final class` with `StringBuilder`-based JSON rendering, single-pass string escaping, pre-computed indent cache, `valTag`-based O(1) dispatch.
- **Interpreter.scala**: Added `interpretStringify`/`materializeStringify` for the direct-to-string pipeline, extracted `createMaterializer()` to share between old and new paths.
- **SjsonnetMainBase.scala**: Fast-path in `renderNormal` when `Platform.useDirectRenderer && !yamlOut && !expectString && outputFile.isEmpty`.
- **Platform.scala** (JVM/Native): Added `useDirectRenderer` flag.

### JMH (JVM, ms/op, lower is better)

| Benchmark | Master | DirectRenderer | Δ |
|-----------|--------|----------------|---|
| realistic2 | 62.0 | 55.9 | **-9.8%** ✅ |
| realistic1 | 2.0 | 1.8 | **-6.1%** ✅ |
| large_string_template | 1.8 | 1.7 | **-8.3%** ✅ |
| bench.02 | 33.0 | 32.6 | **-1.3%** ✅ |
| gen_big_object | 0.93 | 0.91 | -2.2% |
| gen_big_string | 0.24 | 0.23 | noise |

### Hyperfine (Scala Native, realistic2 > /dev/null)

| Binary | Mean (ms) | vs jrsonnet |
|--------|-----------|-------------|
| sjsonnet master | 251.8 ± 2.7 | 2.52x |
| sjsonnet DirectRenderer | 252.5 ± 4.1 | 2.52x |
| jrsonnet | 100.1 ± 2.5 | 1.00x |

Native is neutral by design (DirectRenderer disabled via `Platform.useDirectRenderer = false`).

## Analysis

The optimization targets JVM-specific overhead:
- **Eliminated**: ~3.3M virtual dispatch calls per realistic2 evaluation (ObjVisitor/ArrVisitor allocation + visitValue/visitString/visitEnd calls)
- **Eliminated**: CharBuilder → Writer → StringBuffer intermediate pipeline and StringBuffer synchronization
- **Preserved**: Native performance (LLVM LTO already handles devirtualization), deep nesting safety, custom Materializable support

## Result

JVM materialization throughput improved by ~8-10% on output-heavy benchmarks. No regression on any benchmark. Native performance unchanged (by design).

### DIFF
--- a/sjsonnet/src-jvm-native/sjsonnet/SjsonnetMainBase.scala
+++ b/sjsonnet/src-jvm-native/sjsonnet/SjsonnetMainBase.scala
@@ -264,11 +264,22 @@ object SjsonnetMainBase {
       path: os.Path,
       wd: os.Path,
       getCurrentPosition: () => Position) = {
-    writeToFile(config, wd) { writer =>
-      val renderer = rendererForConfig(writer, config, getCurrentPosition)
-      val res = interp.interpret0(jsonnetCode, OsPath(path), renderer)
-      if (config.yamlOut.value && !config.noTrailingNewline.value) writer.write('\n')
-      res
+    // Fast path: direct JSON rendering to StringBuilder when outputting to stdout.
+    // Bypasses the Visitor pattern, eliminating per-object/array allocations and
+    // virtual dispatch overhead (~10% materialization speedup on JVM for realistic workloads).
+    // Disabled on Native where LLVM LTO already devirtualizes the Visitor pattern.
+    // Not used for YAML, string mode, or file output (which benefits from streaming).
+    if (
+      Platform.useDirectRenderer && !config.yamlOut.value && !config.expectString.value && config.outputFile.isEmpty
+    ) {
+      interp.interpretStringify(jsonnetCode, OsPath(path), config.indent)
+    } else {
+      writeToFile(config, wd) { writer =>
+        val renderer = rendererForConfig(writer, config, getCurrentPosition)
+        val res = interp.interpret0(jsonnetCode, OsPath(path), renderer)
+        if (config.yamlOut.value && !config.noTrailingNewline.value) writer.write('\n')
+        res
+      }
     }
   }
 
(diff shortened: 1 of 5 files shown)


He-Pin:
Superseded by #745 (byte-renderer pipeline with fused materializer) and #747 (fused materializer for stdout). Those PRs implement the same concept with better optimization.

PR #742, opened 2026-04-11
perf: lazy concat for large array concatenation

He-Pin:
## Motivation

Array concatenation in Jsonnet is common, especially in patterns like `big_array + [element]`. The default eager implementation copies the entire left array for every `+` operation, making it O(n) even when the result is only used for element-wise access (e.g., comparison, iteration).

jrsonnet uses an `ExtendedArray` (arr/spec.rs) that defers the copy with O(1) concat for large arrays. This PR brings the same optimization to sjsonnet, plus a shared-prefix comparison skip that eliminates O(n) element-by-element comparison for patterns like `big_array + [x] < big_array + [y]`.

## Key Design Decisions

- **Threshold**: Only arrays with `left.length >= 256` use lazy concat. Below this, `arraycopy` is cheap enough that the indirection overhead outweighs the savings.
- **Depth limit**: ConcatViews are limited to depth-1 (neither operand can already be a ConcatView). This prevents O(depth) access chains and ensures `value(i)` stays O(1).
- **Transparent API**: `asLazyArray` materializes the view on-demand, so all existing code paths work unchanged.
- **`eval(i)` method**: Returns raw `Eval` without forcing the value, preserving reference equality for shared prefix elements across ConcatViews.
- **`sharedConcatPrefixLength()`**: Detects when two ConcatViews share the same left array, enabling O(1) prefix skip in compare/equal operations.
- **Shared prefix skip in compare/equal**: When two arrays are ConcatViews sharing the same left array, the comparison loop starts after the shared prefix — turning O(n) comparison into O(suffix_len).

## Modification

**`sjsonnet/src/sjsonnet/Val.scala`** — `Val.Arr`:
- Added `_concatLeft`, `_concatRight`, `_length` fields for lazy concat state
- `concat()`: creates ConcatView for large arrays, falls back to eager copy otherwise
- `value(i)`: dispatches through ConcatView tree transparently
- `eval(i)`: returns raw Eval without forcing (preserves reference identity)
- `sharedConcatPrefixLength()`: O(1) shared-prefix detection
- `materialize()`: flattens ConcatView on-demand when bulk access is needed
- `iterator`, `foreach`, `forall`: work transparently with both flat and ConcatView arrays

**`sjsonnet/src/sjsonnet/Evaluator.scala`** — `compare()` and `equal()`:
- Array comparison loop now starts at `sharedConcatPrefixLength()` instead of 0
- Skips entire shared prefix for ConcatView arrays, avoiding millions of virtual dispatches

**Test files**: `lazy_concat_array.jsonnet` — comprehensive test covering basic concat, nested concat, large array + small append, mixed operations, and deep nesting.

### JMH (JVM, single iteration)

| Benchmark | Master | This PR | Change |
|-----------|--------|---------|--------|
| comparison (array) | 22.134 ms | 19.634 ms | **-11.3%** ✅ |
| comparison2 (primitives) | 37.265 ms | 37.538 ms | ≈ neutral |
| All other benchmarks | — | — | No regressions |

### Hyperfine (Scala Native, 10 runs)

| Benchmark | Master | This PR | jrsonnet | sjsonnet/jrsonnet |
|-----------|--------|---------|----------|-------------------|
| comparison_for_array | 30.8 ms | **15.6 ms** | 12.0 ms | **1.30x** (was 2.61x) |

## Analysis

The optimization provides two complementary speedups:

1. **O(1) concat** (lazy ConcatView): Avoids copying 1M references when creating `big_array + [x]`. Saves ~10ms of arraycopy + GC pressure on Scala Native.

2. **Shared prefix skip** (`sharedConcatPrefixLength()`): When comparing `big_array + [x] < big_array + [y]`, detects the shared left array and starts comparison at element 1,000,000 instead of 0. Eliminates 1M virtual dispatches.

Combined, these bring the `comparison_for_array` benchmark from 2.61x slower than jrsonnet to just 1.30x. The remaining gap (~3.6ms) is largely Scala Native startup overhead vs Rust.

## References

- jrsonnet `ExtendedArray`: `jrsonnet/crates/jrsonnet-evaluator/src/arr/spec.rs`
- jrsonnet benchmark methodology: `jrsonnet/docs/benchmarks.adoc`

## Result

Lazy concatenation with shared-prefix comparison reduces the `comparison_for_array` gap from **2.61x to 1.30x** vs jrsonnet, with zero regressions across all other benchmarks.


stephenamar-db:
please rebase

He-Pin:

PR #743, opened 2026-04-11
perf: pre-check Eval reference equality in array compare/equal

He-Pin:
## Motivation

Array comparison (`<`, `>`, `<=`, `>=`) and equality (`==`) iterate element-by-element, forcing each lazy `Eval` into a `Val`. For patterns like `big_array + [x] < big_array + [y]`, this forces 1M+ identical elements unnecessarily — the shared prefix from `big_array` is guaranteed equal.

The `comparison_for_array` benchmark (`std.range(1,1000000) + [1] < std.range(1,1000000) + [2]`) was **1.98x slower** than jrsonnet. This PR closes the gap to **1.25x**.

## Key Design Decisions

1. **Shared ConcatView prefix skip**: Uses `sharedConcatPrefixLength()` (from PR #742) to detect when two arrays share the same left array reference in a ConcatView. When detected, the comparison loop starts AFTER the shared prefix, turning O(n) into O(right_len).

2. **`eval(i)` instead of `asLazyArray`**: Uses `eval(i)` (from PR #742) to access raw `Eval` references without materializing ConcatViews. This avoids two O(n) arraycopy operations that were previously needed to flatten the views before comparison.

3. **Pre-force Eval reference equality**: For elements not covered by the prefix skip, checks `(xe eq ye)` on raw Eval refs. Shared primitive Vals (tag <= TAG_NULL) are trivially equal — no forcing needed.

4. **Inlined numeric fast path**: When elements must be forced, inlines `Double.compare` for the common `Val.Num` vs `Val.Num` case to avoid recursive `compare()` dispatch overhead.

5. **Shared Lazy error semantics**: In `equal()`, shared Lazy thunks (non-Val Eval) are still forced to preserve error propagation semantics.

## Modification

**`sjsonnet/src/sjsonnet/Evaluator.scala`**:
- `compare()` array case: Uses `eval(i)` + `sharedConcatPrefixLength()` + Eval ref equality + inlined numeric fast path
- `equal()` array case: Same optimizations with proper Lazy error semantics

**Test files** (10 files):
- `array_compare_shared_prefix.jsonnet` — correctness tests for shared-prefix comparison
- `error.array_compare_shared_error.jsonnet` — error propagation in shared arrays
- `error.array_compare_shared_nested.jsonnet` — nested array error handling
- `error.array_compare_shared_object.jsonnet` — object comparison errors
- `error.array_equal_shared_error.jsonnet` — equality error propagation

### JMH (JVM, single iteration)

| Benchmark | Master | This PR | Change |
|-----------|--------|---------|--------|
| comparison (array) | 21.661 ms | 17.680 ms | **-18.4%** ✅ |
| comparison2 (primitives) | 37.265 ms | 35.257 ms | **-5.4%** ✅ |
| realistic2 | 63.665 ms | 61.967 ms | **-2.7%** ✅ |

### Hyperfine (Scala Native, 30 runs, `--shell=none`)

| Benchmark | Master | This PR | jrsonnet | sjsonnet/jrsonnet |
|-----------|--------|---------|----------|-------------------|
| comparison_for_array | 26.9 ± 0.5 ms | **16.9 ± 1.2 ms** | 13.1 ± 0.8 ms | **1.25x** |

**37% speedup** on the comparison_for_array benchmark. Gap reduced from **1.98x → 1.25x** vs jrsonnet.

### No Regressions

| Benchmark | Master | This PR | Change |
|-----------|--------|---------|--------|
| realistic2 | 262 ms | 264 ms | ~0% (noise) |
| comparison_primitives | 88.5 ms | 88.9 ms | ~0% |
| array_sorts | ~10 ms | ~10 ms | ~0% |

## Analysis

The optimization is most effective when:
- Two arrays share a common large prefix via concatenation (`a + [x]` vs `a + [y]`)
- The shared prefix dominates the array length
- Elements in the prefix are primitive Vals (Num, Str, Bool, Null)

For arrays without shared prefixes, the overhead is minimal: one `sharedConcatPrefixLength()` call (returns 0) + per-element `eval(i)` instead of `asLazyArray` array indexing.

## References

- Depends on PR #742 (lazy concat array) for `eval(i)` and `sharedConcatPrefixLength()` methods
- jrsonnet benchmark: `jrsonnet/docs/benchmarks.adoc` — comparison_for_array
- jrsonnet source: `jrsonnet/crates/jrsonnet-evaluator/src/arr/spec.rs` (ExtendedArray)

## Result

Dramatic improvement on array comparison benchmark. Combined with lazy concat (PR #742), closes the gap from 1.98x to 1.25x vs jrsonnet for the `comparison_for_array` workload.


stephenamar-db:
please rebase

He-Pin:

stephenamar-db:
sadly, again rebase.

He-Pin:
Rebased onto latest master (includes cachedNum, cycle detection, and auto-TCO arity fix).

PR #744, opened 2026-04-11
perf: Val.cachedNum pool for small integer allocation reduction

He-Pin:
## Motivation

Numeric `Val.Num` allocations are pervasive in Jsonnet evaluation — every arithmetic operation, every stdlib function returning an index or length, and every importbin byte creates a fresh heap object. For small non-negative integers (0–255), these values repeat very frequently (loop counters, array indices, byte values, character codepoints), making them ideal candidates for pooling.

## Key Design Decisions

1. **256-entry pool (0–255)**: Pre-allocated at startup with `Val.cachedNum(pos, d)` factory. Only exact integer doubles in range are cached — fractional, negative, infinity, and NaN values always allocate fresh.
2. **IEEE-754 signed zero safety**: Uses `doubleToRawLongBits` comparison instead of `==` to correctly distinguish `-0.0` from `+0.0`. Signed zero is observable via `std.math.atan2(-0.0, -1)` → `-π` vs `std.math.atan2(+0.0, -1)` → `+π`.
3. **Null-safe Position handling**: Cached pool instances use synthetic `Position(null, -1)`. Made `Position.currentFile` and `Position.noOffset` null-safe so error formatting and `--yaml-debug` materialization gracefully degrade instead of NPE.
4. **Error path protection**: `Error.Frame` and `EvalErrorScope.prettyIndex` now handle null `currentFile`, displaying `<cached>` as the source file for pool instances.

## Modification

**Evaluator.scala**: All 13 binary numeric ops (`+`, `-`, `*`, `/`, `%`, `<<`, `>>`, `&`, `^`, `|`, comparisons) and `visitImportBin` byte values now use `Val.cachedNum`.

**stdlib/ArrayModule.scala**: `std.count`, `std.find`, `std.range`, `std.makeArray`, `mapWithIndex` — all return/use small integer indices.

**stdlib/StringModule.scala**: `std.length`, `codepoint`, `encodeUTF8`, `parseInt`/`parseOctal`/`parseHex`, `findSubstr`.

**stdlib/MathModule.scala**: `mod` operator.

**Val.scala**: `cachedNum` lookup with `doubleToRawLongBits` guard.

**Parser.scala**: `Position.currentFile` and `noOffset` null-safe for synthetic positions.

**Error.scala**: `Error.Frame.ste` and `prettyIndex` tolerate null `currentFile`.

**New test**: `signed_zero_preservation.jsonnet` — regression test ensuring `-0.0` is not canonicalized to `+0.0`.

### JMH (JVM, `bench.runRegressions`, 1 fork, 1 warmup, 1 measurement)

| Benchmark | Master (ms/op) | cachedNum (ms/op) | Change |
|-----------|---------------|-------------------|--------|
| comparison | 26.546 | 24.447 | **-7.9%** ✅ |
| base64_byte_array | 1.619 | 1.134 | **-30.0%** ✅ |
| realistic1 | 2.026 | 2.026 | **-8.6%** ✅ |
| realistic2 | 65.458 | 61.910 | **-5.4%** ✅ |
| member | 0.699 | 0.646 | **-7.6%** ✅ |
| comparison2 | 37.649 | 38.279 | +1.7% (noise) |
| manifestYamlDoc | 0.063 | 0.058 | **-7.9%** ✅ |
| manifestTomlEx | 0.073 | 0.070 | **-4.1%** ✅ |

### Hyperfine (Scala Native, 10 runs, 5 warmup)

| Benchmark | Master (ms) | cachedNum (ms) | jrsonnet (ms) | Change |
|-----------|------------|----------------|---------------|--------|
| comparison_for_array | 29.7 ± 1.0 | 28.7 ± 1.4 | 15.4 ± 2.1 | -3.4% (native neutral) |
| std_base64_byte_array | 9.6 ± 0.8 | 10.2 ± 1.2 | 19.5 ± 1.5 | neutral (already 2x faster than jrsonnet) |

Native impact is minimal as expected — Scala Native Immix GC uses bump-pointer allocation where caching provides less benefit than on JVM with generational GC.

## Analysis

The optimization primarily targets JVM where young-gen GC pressure from short-lived `Val.Num` allocations is measurable. The 256-entry pool covers the most common integer values (loop counters, array indices, ASCII codepoints, byte values). The `doubleToRawLongBits` guard adds ~1 extra long comparison on the fast path but prevents signed zero semantic corruption.

## References

- Similar integer caching in JVM: `java.lang.Integer.IntegerCache` (-128 to 127)
- IEEE-754 signed zero: https://en.wikipedia.org/wiki/Signed_zero

## Result

All 23 test suites pass. No regressions detected across 35 JMH benchmarks.


stephenamar-db:
please rebase

He-Pin:

He-Pin:
@stephenamar-db rebased

PR #749, opened 2026-04-11
perf: SIMD-accelerated FastBase64 for Scala Native via C FFI

He-Pin:
## Motivation

On Scala Native, `java.util.Base64` is a pure-Scala implementation that uses Wrapper objects, `@tailrec` recursive `iterate()`, and per-byte pattern matching — significantly slower than HotSpot's intrinsic-backed implementation.

Beyond the raw codec, `base64DecodeBytes` was creating `Array[Eval](N)` and filling each slot with `Val.cachedNum` — N allocations for an N-byte decode. The materializer then needed per-element type dispatch to render these arrays. And `base64` encode output (guaranteed ASCII-safe) was still being scanned for JSON escape characters. `Val.Arr` carried inline `_isRange`/`_byteData` fields that bloated every regular array instance (~13 bytes wasted per non-specialized array).

### 1. Platform-agnostic `FastBase64` encoder/decoder
- `ENCODE_TABLE` (char[64]) and `DECODE_TABLE` (int[256]) pre-computed lookup tables
- `encodeString()`: ASCII fast path does direct char→char encoding without intermediate `byte[]`
- `decodeToString()` / `decodeToBytes()`: Direct string→bytes via lookup table
- ISO-8859-1 compatibility: chars > 0xFF → 0x3F ('?') matching `java.util.Base64` behavior

### 2. C FFI SIMD base64 for Scala Native (`sjsonnet_base64.c`)
- **AArch64 NEON**: `vld3`/`vst4` interleaved load/store + `vqtbl4q` 64-byte lookup for encode; `vbslq`/`vmovl_u8`/`vmovn_u16` for byte↔char widening/narrowing
- **x86_64**: SSSE3/AVX2/AVX-512 VBMI paths via `pshufb`/`vpshufb`/`vpermi2b`
- **Fallback**: Scalar with loop unrolling for other architectures
- `sjsonnet_base64_decode_validated()`: Single-pass validation + decode with specific error codes
- RFC 4648 compliant with '=' padding

### 3. Native-specific optimizations
- Reusable module-level buffers (safe: Scala Native is single-threaded) — eliminates per-call array allocations
- ASCII fast-path in `encodeString`: skip UTF-8 encoding for pure ASCII strings
- Direct char array construction instead of charset lookup

### 4. `RangeArr` and `ByteArr` subclasses of `Val.Arr`
- `Val.Arr` changed from `final class` to non-final `class`, enabling specialization
- **`RangeArr extends Arr`**: Lazy integer range — keeps `rangeFrom` field out of regular arrays, saving ~9 bytes per non-range array (merges #772)
- **`ByteArr extends Arr`**: Compact `Array[Byte]` backing store for 0–255 integer arrays
  - `byteData` is an immutable `val` — never cleared after materialization, guaranteeing `rawBytes` is always non-null for safe multi-use
  - `reversed()` materializes first to keep `value()`/`eval()` simple and avoid reversed-index bugs
  - `rawBytes` accessor enables zero-copy fast paths in `base64` encode and materializer
- Callers use pattern match (`case ba: Val.ByteArr =>`) instead of null-returning `rawBytes` on base class

### 5. Materializer fast-path for byte arrays
- Recursive, iterative, and fused ByteRenderer paths all detect `ByteArr` via pattern match
- Skip `value(i)` lookup + type dispatch + `asDouble` conversion
- Directly emit `visitFloat64((bytes(i) & 0xff).toDouble)` in a tight loop

### 6. ASCII-safe string rendering
- `Val.Str._asciiSafe` flag marks strings known to contain only printable ASCII (no JSON escaping needed)
- `Val.Str.asciiSafe(pos, s)` factory for creating flagged strings
- `BaseByteRenderer.renderAsciiSafeString()` skips SWAR escape scanning and UTF-8 encoding — writes bytes directly from chars
- `base64` encode output is marked as ASCII-safe since base64 alphabet is `[A-Za-z0-9+/=]`

### 7. `EncodingModule` updates
- `base64DecodeBytes`: Uses `Val.Arr.fromBytes(pos, decoded)` — one allocation instead of N
- `base64` encode: Pattern matches `ByteArr` for zero-copy bypass; output marked `asciiSafe`

### JMH (JVM, Scala 3.3.7, Apple Silicon M4 Max)

| Benchmark | Master (ms/op) | PR (ms/op) | Change |
|-----------|---------------|------------|--------|
| base64 | 0.153 | 0.145 | **-5.2%** |
| base64Decode | 0.117 | 0.115 | -1.7% |
| base64DecodeBytes | 5.692 | 5.109 | **-10.2%** |
| base64_byte_array | 0.757 | 0.758 | ~same |
| base64_stress | — | 0.188 | (new) |

### Scala Native (hyperfine -N, 30 runs, Apple Silicon M4 Max)

Compared against jrsonnet **0.5.0-pre98** (built from source, `cargo build --release`).

| Benchmark | sjsonnet master | sjsonnet PR | jrsonnet 0.5.0 | PR vs master | PR vs jrsonnet |
|-----------|----------------|-------------|----------------|--------------|----------------|
| base64 | 8.7ms | 6.5ms | 4.4ms | **1.34× faster** | 1.47× slower |
| base64Decode | 7.3ms | 6.8ms | 4.3ms | 1.07× faster | 1.60× slower |
| base64DecodeBytes | 28.7ms | 13.5ms | 20.1ms | **2.13× faster** | **1.50× faster** |
| base64_byte_array | 10.5ms | 8.5ms | 17.3ms | **1.23× faster** | **2.02× faster** |
| base64_stress | 6.6ms | 6.3ms | 5.0ms | ~same | 1.28× slower |

**Compute-heavy benchmarks** (`base64DecodeBytes`, `base64_byte_array`): sjsonnet significantly outperforms jrsonnet — 1.50× and 2.02× faster respectively.

**Small benchmarks** (`base64`, `base64Decode`, `base64_stress`): jrsonnet is faster due to lower startup overhead (~3ms vs ~5ms). The actual base64 computation time is comparable; the gap is dominated by process startup.

## Files Changed

| File | Change |
|------|--------|
| `sjsonnet/src/sjsonnet/Val.scala` | `Arr` non-final, `RangeArr` + `ByteArr` subclasses, `_asciiSafe` flag, `asciiSafe` factory |
| `sjsonnet/src/sjsonnet/Materializer.scala` | ByteArr pattern-match fast path in recursive + iterative paths |
| `sjsonnet/src/sjsonnet/ByteRenderer.scala` | ByteArr fast path in fused materializer + ASCII-safe string dispatch |
| `sjsonnet/src/sjsonnet/BaseByteRenderer.scala` | `renderAsciiSafeString()` for escape-free rendering |
| `sjsonnet/src/sjsonnet/stdlib/EncodingModule.scala` | `fromBytes` for DecodeBytes, ByteArr match for encode, `asciiSafe` for output |
| `sjsonnet/src-js/sjsonnet/stdlib/FastBase64.scala` | Pure Scala implementation (JS/WASM) |
| `sjsonnet/src-jvm/sjsonnet/stdlib/FastBase64.scala` | Delegates to `java.util.Base64` (unchanged behavior) |
| `sjsonnet/src-native/sjsonnet/stdlib/FastBase64.scala` | C FFI wrappers + buffer reuse + ASCII fast paths |
| `sjsonnet/resources/scala-native/sjsonnet_base64.c` | SIMD C implementation (NEON/SSSE3/AVX2/AVX-512 + scalar fallback) |
| `sjsonnet/test/resources/new_test_suite/byte_arr_correctness.jsonnet` | Regression tests for ByteArr (multi-use, reverse, concat, round-trip) |
| `sjsonnet/test/resources/new_test_suite/range_arr_correctness.jsonnet` | Regression tests for RangeArr correctness |
| `bench/resources/go_suite/base64_stress.jsonnet` | New benchmark for mixed encode/decode stress test |

## Result

- base64DecodeBytes **2.13× faster** than master, **1.50× faster** than jrsonnet 0.5.0
- base64_byte_array **2.02× faster** than jrsonnet 0.5.0
- JVM base64DecodeBytes improved **10.2%** vs master
- All JVM, JS, and Native tests pass


stephenamar-db:
not sure it's worth it.

He-Pin:
This needs to be SIMD-based

He-Pin (commented):
### PR #749 Review: SIMD-accelerated base64 for Scala Native

**Overall**: Major feature, well-architected with platform-specific implementations. The byte-backed `Val.Arr` is a good general-purpose optimization beyond just base64. Benchmark results are solid - `base64DecodeBytes` 1.26x faster than jrsonnet, `base64_byte_array` 1.94x faster.

**Concern 1 - `_byteData` mutability**: The `rawBytes` accessor returns the internal `_byteData` array directly. If someone modifies the underlying byte array, the cached `Val.Num` objects from `value(i)` could become stale. Consider documenting immutability guarantee or returning defensive copy.

**Concern 2 - C file complexity**: The 1255-line C file (`sjsonnet_base64.c`) is complex. If there are bugs in the SIMD paths (NEON, SSSE3, AVX2, AVX-512), they could be hard to track down. Recommend comprehensive edge case tests:
- Empty input
- Input lengths 1, 2, 3 (boundary cases for base64 encoding)
- Input with all possible byte values (0x00-0xFF)
- Large input (>64 bytes to trigger SIMD paths)
- Invalid padding detection

**Known issue already fixed**: AVX-512 `avx512dq` target feature was missing - resolved in the follow-up commit.

**Startup overhead**: The benchmark shows `base64 encode` and `base64Decode` are still slower than jrsonnet on Scala Native (1.77x and 1.50x). This is attributed to startup overhead (5.5ms vs 3.2ms). Consider investigating the startup cost separately.

He-Pin (commented):
### PR #749 问题

1. **`_byteData` 可变性**: `rawBytes` 直接返回内部 `_byteData` 数组。外部修改会导致 `value(i)` 返回的缓存 `Val.Num` 过期。建议文档注明 "Callers MUST NOT modify" 或返回 `bytes.clone()`。`fromBytes` 同理，应做防御性拷贝。

2. **C 文件边缘用例测试**: 1255 行 C 文件包含 NEON/SSSE3/AVX2/AVX-512 多条 SIMD 路径。建议补充：
   - 输入长度 1, 2, 3（base64 边界情况）
   - 包含所有字节值 0x00-0xFF 的输入
   - >64 字节输入（触发 SIMD 主循环）
   - 畸形 padding 检测

3. **启动开销**: `base64 encode`/`decode` 仍比 jrsonnet 慢 1.77x/1.50x，归因于启动开销（5.5ms vs 3.2ms）。建议单独分析启动成本。

stephenamar-db:
I don't see a difference in the PR comment? This seems neutral everywhere. Are there updated benchs?

He-Pin:
@stephenamar-db The real power need other pr be merged first, otherwise the scala native start up time will reduce the numbers, because the numbers is really mall,will conver to draft

sjsonnet/src-js/sjsonnet/stdlib/FastBase64.scala
```diff
@@ -0,0 +1,256 @@
...
+      else if (v == -2) paddingCount += 1
+      else {
+        throw new IllegalArgumentException(
+          "Illegal base64 character " + Integer.toHexString(b)
+        )
+      }
+      ci += 1
+    }
+
+    val totalChars = validCount + paddingCount
```
He-Pin, line 202: charToIso8859 maps chars > 0xFF to 0x3F which decodes as '/' in the DECODE_TABLE. This means invalid base64 chars like \u0100 are silently treated as '/' instead of throwing IllegalArgumentException. java.util.Base64.getDecoder().decode() throws on such input. This is a behavioral difference that could mask bugs in callers passing corrupted input.

He-Pin:
I want to build https://github.com/databricks/sjsonnet/pull/776 on top of this.

He-Pin:
@stephenamar-db The performance of base64 improved now, and the SIMD part will help the rendering pipeline performane with simd enhances later.

JoshRosen (commented):
I think that there might be multiple correctness issues in `sjsonnet_base64.c`. I prompted Claude Opus 4.6 (in a `claude.ai` chat conversation, with code interpreter enabled) to take a look at this PR and after some back-and-forth we uncovered some significant correctness issues.

One "code smell" that prompted me to dig in was the presence of several code comments where it looks like an LLM backed out of one implementation approach in favor of another, e.g.

https://github.com/databricks/sjsonnet/blob/52f2b6b2cded4190f51ae1f6e9071c13f790f5a5/sjsonnet/resources/scala-native/sjsonnet_base64.c#L836-L837
or

https://github.com/databricks/sjsonnet/blob/52f2b6b2cded4190f51ae1f6e9071c13f790f5a5/sjsonnet/resources/scala-native/sjsonnet_base64.c#L521-L522

I prompted Claude to look for security + correctness issues, and to focus on these types of "changed my mind" comments and this flagged several issues. Here's Claude's summary:


> ## Executive Summary
>  
> All three x86 SIMD codepaths (SSSE3, AVX2, AVX-512 VBMI) in `sjsonnet_base64.c` produce incorrect output for both encode and decode. The bugs were confirmed by compiling the C source natively on x86_64 with all three instruction sets available, running it through a simulation of the exact Scala Native FFI wrapper logic, and comparing against scalar baseline and RFC 4648 expected values.
>  
> The C source contains 13 LLM chain-of-thought comments and 8 dead variables from abandoned approaches that directly correlate with the bug locations.
>  
> ## Test Environment
>  
> - **CPU**: x86_64 with SSSE3, AVX2, and AVX-512 VBMI (native execution, not emulated)
> - **Compiler**: GCC with `-O2 -march=native`
> - **Method**: The C file was compiled directly and called through a harness that replicates the exact Scala `FastBase64.scala` FFI wrapper logic — char-to-byte conversion, C call, byte-to-char conversion. Feature-detection globals were overridden to force each SIMD tier independently.
>  
> ## Finding 1: Data-Corrupting Bugs in All x86 SIMD Paths
>  
> ### Decode: 3-byte group reversal
>  
> Each SIMD decode path reverses the byte order within every 3-byte output group. The project's own test assertion demonstrates this:
>  
> ```
> std.assertEqual(std.base64Decode("SGVsbG8gV29ybGQh"), "Hello World!")
> ```
>  
> | Path | Output | Correct? |
> |---|---|---|
> | Scalar | `Hello World!` | ✅ |
> | SSSE3 | `leH olroW!dl` | ❌ |
> | AVX2 | `leH olroW!dl` | ❌ |
> | AVX-512 | (needs ≥64 chars to trigger) | — |
>  
> At SIMD-triggering lengths, every 3-byte group is cleanly reversed: `Hel→leH`, `Wor→roW`, `ld!→!dl`. The scalar tail handles any trailing bytes correctly.
>  
> Full verification at AVX-512 decode threshold (64-char input decoding to 48 bytes):
>  
> ```
> Expected: ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuv
> AVX-512:  CBAFEDIHGLKJONMRQPUTSXWVaZYdcbgfejihmlkponsrqvut
> ```
>  
> All 16 three-byte groups reversed.
>  
> ### Encode: corrupted 6-bit index extraction
>  
> The encode bug is mechanistically different from decode. The SSSE3/AVX2/AVX-512 encode paths use a reshuffle mask that is incompatible with the Muła multiply constants that follow it. This causes the `mulhi_epu16`/`mullo_epi16` extraction to pull 6-bit indices from the wrong byte positions, producing corrupted base64 characters — not a clean reversal, but a non-trivial scramble:
>  
> ```
> Input:  "abcdefghijklmnop" (16 bytes)
> Scalar: YWJjZGVmZ2hpamtsbW5vcA==  ✅
> SSSE3:  YmBhZWBkaGBna2BqbW5vcA==  ❌
> ```
>  
> Cross-path verification confirms real corruption: SSSE3-encoded data decoded by the scalar path produces `b\x60ae\x60dh\x60gk\x60jmnop` instead of `abcdefghijklmnop`.
>  
> ### Activation thresholds
>  
> Each SIMD tier only activates above a minimum input size. The dispatcher is an if/else chain that selects the highest available tier, so on an AVX-512 machine, only AVX-512 thresholds matter:
>  
> | Path | Encode activates at | Decode activates at |
> |---|---|---|
> | SSSE3 | ≥16 input bytes | ≥16 input chars |
> | AVX2 | ≥32 input bytes | ≥32 input chars |
> | AVX-512 | ≥48 input bytes | ≥64 input chars |
>  
> Inputs below the selected tier's threshold fall through to the correct scalar implementation.
>  
> ### Why project tests pass
>  
> The project's test inputs are small. The `stdlib.jsonnet` encode inputs are 12, 11, 10, and 0 bytes — below every SIMD encode threshold. The `byte_arr_correctness.jsonnet` inputs are 8, 4, and 0 base64 chars — below every decode threshold.
>  
> The 16-char decode input `SGVsbG8gV29ybGQh` is the only test that reaches a SIMD threshold: it exactly meets the SSSE3 decode minimum of 16 chars. On a CPU with only SSSE3 (no AVX2/AVX-512), this test would fail. But on the AVX-512 CPUs where the PR was likely tested, the dispatcher selects AVX-512, whose 64-char decode threshold is not met, so execution falls through to scalar and the test passes by accident.
>  
> The PR was benchmarked on Apple Silicon M4 Max (ARM64 NEON). The NEON implementation uses hardware interleaved load/store intrinsics (`vld3q_u8`/`vst4q_u8`) that handle byte ordering automatically — a fundamentally different approach from the x86 paths.
>  
> ### Impact
>  
> This bug silently corrupts data whenever Scala Native runs on x86 and processes base64 inputs above the SIMD threshold:
> - Scala Native on x86 encodes base64 → JVM or any standard decoder reads it
> - External/standard base64 is decoded by Scala Native on x86
> - Encoded output is compared to or consumed by any non-sjsonnet-x86 system
>  
> ## Finding 2: LLM Chain-of-Thought Comments Map to Bug Locations
>  
> The C source contains 13 comments characteristic of LLM chain-of-thought reasoning — mid-function strategy pivots, self-corrections, and abandoned approaches left in place. These correlate directly with the code regions containing the bugs.
>  
> ### Encode: wrong reshuffle mask survived three attempts
>  
> The SSSE3 encode function contains three sequential attempts at 6-bit index extraction, with only the third used:
>  
> **Attempt 1** (lines 310–312): Shift-and-mask approach. The LLM computed `t0` and `t1`, then annotated `"t0 has: byte0=(in2>>4)&0x3F=wrong... need different approach"`. Dead code — `t0` and `t1` are never used.
>  
> **Attempt 2** (lines 344–364): Range classification using saturating subtract. The LLM built `cmp` and `less26`, then wrote `"Hmm, this doesn't work directly. Let me use the standard approach"` when it recognized a collision between index ranges. Dead code — `cmp` and `less26` from this block are abandoned.
>  
> **Attempt 3** (lines 370–406, labeled `"Redo"`): The final range classification, which works correctly in isolation. But it operates on `indices` produced by the Muła multiply at lines 321–329, which in turn depends on the reshuffle mask at line 279. The reshuffle mask reverses each 3-byte group (`[2,1,0,-1, 5,4,3,-1, ...]`) instead of creating the overlapping byte pairs the multiply constants expect. The classification logic is correct; its input is wrong.
>  
> ### Decode: byte-order error in pack shuffle
>  
> The SSSE3 decode has a parallel pattern of abandoned approaches:
>  
> **Lines 478–510**: Three abandoned decode strategies. `hi_nibbles` (nibble-based classification), `offset_lut` (nibble offset table), and `ca` (mullo pack attempt) are all computed and never used. Comments include `"wait let me recalculate"`, `"This is getting complex"`, and `"Hmm, the pack is tricky"`.
>  
> **Line 446–448**: The surviving `pack_shuf` extracts bytes `[0,1,2]` from each 32-bit lane. After `maddubs`+`madd`, each lane holds a 24-bit value in little-endian: byte 0 is the LSB (output byte 2), byte 2 is the MSB (output byte 0). The correct pack should extract `[2,1,0]` per lane. This is the direct cause of the 3-byte group reversal.
>  
> ### AVX-512: abandoned constants, same byte-order errors
>  
> The AVX-512 encode function contains two dead reshuffle constants (`input_shuf` at line 827, `shuf48` at line 848), each abandoned after comments like `"Actually, let me use a cleaner approach"` and `"Hmm, _mm512_set_epi8 fills from high byte to low byte. Let me fix ordering"`. The third attempt (`shuf_perm` at line 859) has the same reversal as SSSE3/AVX2.
>  
> The AVX-512 decode has a dead `pack_shuf` (line 926) that is never referenced. The actual gather uses `gather_idx` (line 972), which picks `[0,1,2]` per lane instead of `[2,1,0]` — the same byte-order error as SSSE3/AVX2.
>  
> ### Dead variable summary
>  
> GCC `-Wunused-variable` confirms 8 dead variables from abandoned LLM approaches:
>  
> ```
> t0           (line 310)  — SSSE3 encode, first extraction attempt
> t1           (line 311)  — SSSE3 encode, first extraction attempt
> hi_nibbles   (line 478)  — SSSE3 decode, nibble-based classification
> offset_lut   (line 504)  — SSSE3 decode, nibble offset table
> ca           (line 578)  — SSSE3 decode, mullo pack attempt
> input_shuf   (line 827)  — AVX-512 encode, first reshuffle attempt
> shuf48       (line 848)  — AVX-512 encode, second reshuffle attempt
> pack_shuf    (line 926)  — AVX-512 decode, unused shuffle constant
> ```
>  
> ## Root Cause Summary
>  
> | Path | Encode bug | Decode bug |
> |---|---|---|
> | SSSE3 | Reshuffle mask `[2,1,0,-1,...]` incompatible with Muła multiply constants `0x0FC0FC00`/`0x04000040` — extracts 6-bit indices from wrong byte positions | `pack_shuf` extracts `[0,1,2]` per lane instead of `[2,1,0]` — reverses each 3-byte output group |
> | AVX2 | Same reshuffle mask (duplicated for 256-bit lanes) | Same `pack_shuf` (duplicated for 256-bit lanes) |
> | AVX-512 | Same reshuffle via `shuf_perm`, same Muła constants | `gather_idx` picks `[0,1,2]` per lane instead of `[2,1,0]` |
> | NEON | Uses `vld3q_u8`/`vst4q_u8` interleaved intrinsics — byte ordering handled by hardware | Uses `vst3q_u8` interleaved store — byte ordering handled by hardware |
>  
> The x86 paths all share the same systematic endianness confusion. The NEON path avoids the issue entirely by using ARM's interleaved load/store intrinsics, which abstract away byte ordering within groups.
> 

If we're going to include a bunch of custom C code, we need stronger tests (and probably more careful code review to actually look at what we're merging!).

Note that I'm not an expert in native SIMD programming, but I place moderate trust in the above analysis given that Claude actually compiled and tested the C code (albeit not through the Scala Native FFI interface, but I don't anticipate that to affect the analysis / outcome here).

stephenamar-db:
Let's roll back. @He-Pin, when you rollforward, please include a more thorough testing suite.

He-Pin:
Thanks for the details. I think a more proper way to handle this may use a depend instanceof of this. Will prepare one with additional build.

PR #751, opened 2026-04-11
fix: update Apply pattern match arity for auto-TCO strict field

He-Pin:
## Motivation

The auto-TCO commit (ecdd0b6, PR #694) added a `strict: Boolean` field to `Apply`, `Apply0`, `Apply1`, `Apply2`, and `Apply3` case classes. The `hasSelfRefExpr` pattern matches in `Materializer` were not updated, causing:

1. **Compilation failure on Scala 2.13.18**: `wrong number of arguments for pattern sjsonnet.Expr.Apply`
2. **Runtime `MatchError` on Scala 3.3.7**: The pattern match silently fell through to the wildcard case, causing incorrect materialization paths for lazy reverse arrays (`lazy_reverse_correctness.jsonnet` failure)

## Modification

Added wildcard for the new `strict` field in all five Apply pattern matches in `Materializer.hasSelfRefExpr`:

- `Apply(_, v, args, _, _)` → `Apply(_, v, args, _, _, _)`
- `Apply0(_, v, _)` → `Apply0(_, v, _, _)`
- `Apply1(_, v, a1, _)` → `Apply1(_, v, a1, _, _)`
- `Apply2(_, v, a1, a2, _)` → `Apply2(_, v, a1, a2, _, _)`
- `Apply3(_, v, a1, a2, a3, _)` → `Apply3(_, v, a1, a2, a3, _, _)`

## Result

- Scala 2.13.18 compiles and all tests pass (including `lazy_reverse_correctness.jsonnet`)
- Scala 3.3.7 all tests pass (no regressions)

PR #759, opened 2026-04-12
fix: update Apply pattern match arity in Materializer.hasSelfRefExpr

He-Pin:
## Motivation

The `Apply`, `Apply0`–`Apply3` case classes gained an extra `tailstrict` boolean field, but the pattern matches in `Materializer.hasSelfRefExpr` were not updated to include the new field — causing a compilation failure on master.

## Modification

Updated pattern match arities for `Apply`, `Apply0`, `Apply1`, `Apply2`, and `Apply3` in `Materializer.hasSelfRefExpr` to include the additional `tailstrict` boolean parameter.

## Result

Master compiles again. No behavioral change — `hasSelfRefExpr` was already ignoring the tailstrict flag, and the wildcard `_` pattern correctly does the same.

### DIFF
--- a/sjsonnet/src/sjsonnet/Materializer.scala
+++ b/sjsonnet/src/sjsonnet/Materializer.scala
@@ -707,17 +707,17 @@ object Materializer extends Materializer {
       case Expr.Error(_, v) => hasSelfRefExpr(v, inNestedObj)
 
       // Apply variants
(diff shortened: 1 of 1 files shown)


He-Pin:
Closing: this fix was already merged via #751 (commit 1bf3b40a).

PR #765, opened 2026-04-12
perf: convert stdlib hot paths from Scala collections to while-loops

He-Pin:
## Motivation

Scala collection methods like `forall`, `exists`, `map`, `mkString`, `indexWhere`, and `forEach` create closures, iterators, and intermediate collections that are harder for JIT/AOT compilers to optimize. In hot stdlib paths, these can be replaced with imperative while-loops for better performance.

## Key Design Decision

Replace functional collection operations with while-loops only in hot stdlib paths where profiling shows measurable impact. This improves JIT inlining, reduces allocation pressure, and produces more predictable native code under Scala Native LTO.

## Modification

**ObjectModule** (6 methods):
- `objectFields`, `objectFieldsAll`, `objectFieldsEx`: iterator → while-loop over `allKeysArray`
- `getObjValuesFromKeys`: for-comprehension → while-loop
- `objectKeysValues`, `objectKeysValuesAll`: for-comprehension → while-loop with pre-allocated array
- `mergePatch.distinctKeys`: `(lKeys ++ rKeys).distinct` → `LinkedHashSet`-based dedup

**ArrayModule** (7 methods):
- `all`: `forall` → early-exit while-loop returning `Val.staticTrue/False`
- `any`: `exists` → early-exit while-loop returning `Val.staticTrue/False`
- `member` (array path): `indexWhere` → while-loop with `found` flag
- `contains`: `indexWhere` → while-loop with `found` flag
- `remove`: `indexWhere` → while-loop with `idx` sentinel
- `minArray/maxArray` (keyF path): `map.zipWithIndex.min/max` → single-pass while-loop
- `repeat`: `ArrayBuilder ++= lazyArray` → `System.arraycopy`

**ManifestModule** (1 method):
- `manifestYamlStream`: `map.mkString` → `StringBuilder` with while-loop

**StringModule** (1 method):
- `escapeStringXML`: `for (c <- str)` → `while` with `charAt`

**Val.Obj** (1 method):
- `visibleKeyNames`: `forEach` lambda → `entrySet` iterator while-loop

### JMH (JVM, single iteration, lower is better)

| gen_big_object | 1.122 | 0.933 | **-16.8%** ✅ |
| large_string_template | 2.432 | 1.659 | **-31.8%** ✅ |
| realistic2 | 61.774 | 56.379 | **-8.7%** ✅ |
| large_string_join | 0.582 | 0.561 | -3.6% |
| member | 0.665 | 0.661 | -0.6% |
| setDiff | 0.426 | 0.416 | -2.3% |
| setInter | 0.377 | 0.371 | -1.6% |
| setUnion | 0.638 | 0.623 | -2.3% |
| bench.02 | 35.330 | 34.461 | -2.5% |

No regressions observed (bench.07 variance is JMH single-fork noise).

### Hyperfine (Scala Native vs jrsonnet, Apple Silicon)

| gen_big_object | 10.1 | 13.2 | **1.31x faster** ✅ |
| large_string_join | 9.0 | 7.7 | 1.17x slower |
| large_string_template | 14.3 | 6.9 | 2.07x slower |
| realistic2 | 161.2 | 100.4 | 1.61x slower |
| member | 8.4 | 5.8 | 1.45x slower |
| comparison | 18.3 | 13.5 | 1.36x slower |

## Analysis

The biggest JMH win is `large_string_template` (-31.8%) which benefits from the ObjectModule while-loop conversions since format rendering triggers object field enumeration. The `gen_big_object` improvement (-16.8%) directly measures ObjectModule throughput. `realistic2` (-8.7%) shows compound benefits from all modules.

On Scala Native, `gen_big_object` is now definitively faster than jrsonnet (1.31x). The remaining gaps in string-heavy benchmarks require rope strings (#761) and format-level optimizations.

## References

Ported from jit branch exploration commits:
- af4832f2 (std.all/any/member while-loops)
- cd612df7 (escapeStringXML, std.contains/remove while-loops)
- d1629fde (manifestYamlStream StringBuilder)
- 8cd8d31a (minArray/maxArray single-pass)
- 91496549 (visibleKeyNames entrySet iterator)

## Result

All 420 tests pass across JVM/JS/WASM/Native × Scala 3.3.7/2.13.18/2.12.21. Consistent improvements on object-heavy and realistic workloads with no regressions.

### DIFF
--- a/sjsonnet/src/sjsonnet/stdlib/ArrayModule.scala
+++ b/sjsonnet/src/sjsonnet/stdlib/ArrayModule.scala
@@ -27,15 +27,20 @@ object ArrayModule extends AbstractFunctionModule {
       } else if (keyF.isInstanceOf[Val.False]) {
         arr.asStrictArray.min(ev)
       } else {
-        val minTuple = arr.asStrictArray
-          .map(v =>
-            keyF
-              .asInstanceOf[Val.Func]
-              .apply1(v, pos.fileScope.noOffsetPos)(ev, TailstrictModeDisabled)
-          )
-          .zipWithIndex
-          .min((x: (Val, Int), y: (Val, Int)) => ev.compare(x._1, y._1))
-        arr.value(minTuple._2)
+        val strict = arr.asStrictArray
+        val func = keyF.asInstanceOf[Val.Func]
+        var bestIdx = 0
+        var bestVal = func.apply1(strict(0), pos.fileScope.noOffsetPos)(ev, TailstrictModeDisabled)
+        var i = 1
+        while (i < strict.length) {
+          val v = func.apply1(strict(i), pos.fileScope.noOffsetPos)(ev, TailstrictModeDisabled)
+          if (ev.compare(v, bestVal) < 0) {
+            bestVal = v
+            bestIdx = i
+          }
+          i += 1
+        }
+        strict(bestIdx)
       }
     }
   }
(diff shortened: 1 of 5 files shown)


sjsonnet/src/sjsonnet/stdlib/ArrayModule.scala
```diff
@@ -27,15 +27,20 @@ object ArrayModule extends AbstractFunctionModule {
...
-          .min((x: (Val, Int), y: (Val, Int)) => ev.compare(x._1, y._1))
-        arr.value(minTuple._2)
+        val strict = arr.asStrictArray
+        val func = keyF.asInstanceOf[Val.Func]
+        var bestIdx = 0
+        var bestVal = func.apply1(strict(0), pos.fileScope.noOffsetPos)(ev, TailstrictModeDisabled)
+        var i = 1
+        while (i < strict.length) {
+          val v = func.apply1(strict(i), pos.fileScope.noOffsetPos)(ev, TailstrictModeDisabled)
+          if (ev.compare(v, bestVal) < 0) {
```
He-Pin, line 37: `minArray`/`maxArray` with `keyF` throws `ArrayIndexOutOfBoundsException` on empty arrays (`func.apply1(strict(0), ...)`). The original `.map(...).min` also threw, but with an unclear error. Suggest adding an explicit check: `if (strict.isEmpty) Error.fail("min/maxArray cannot be called on an empty array with keyF")`.

sjsonnet/src/sjsonnet/stdlib/ArrayModule.scala
```diff
@@ -582,10 +611,23 @@ object ArrayModule extends AbstractFunctionModule {
       )
     },
     builtin("contains", "arr", "elem") { (_, ev, arr: Val.Arr, elem: Val) =>
-      arr.asLazyArray.indexWhere(s => ev.equal(s.value, elem)) != -1
+      val la = arr.asLazyArray
+      var i = 0
+      var found = false
+      while (i < la.length && !found) {
```
He-Pin, line 617: `contains` and `remove` use `var found = false` flag, inconsistent with the early return style of `all`/`any` above. Suggest using early return to eliminate the flag:
```scala
var i = 0
while (i < la.length) {
  if (ev.equal(la(i).value, elem)) return true
  i += 1
}
false
```
Same for `remove`.

sjsonnet/src/sjsonnet/stdlib/ObjectModule.scala
```diff
@@ -348,8 +367,13 @@ object ObjectModule extends AbstractFunctionModule {
...
           }
         } else {
-          // Fallback: Use hash-based deduplication for large RHS arrays:
-          (lKeys ++ rKeys).distinct
+          // Fallback: Use LinkedHashSet for large RHS arrays (preserves order, avoids quadratic):
+          val allKeys = new java.util.LinkedHashSet[String](lKeys.length + rKeys.length)
+          var li = 0
+          while (li < lKeys.length) { allKeys.add(lKeys(li)); li += 1 }
+          var ri = 0
+          while (ri < rKeys.length) { allKeys.add(rKeys(ri)); ri += 1 }
```
He-Pin, line 375: `LinkedHashSet` initial capacity `lKeys.length + rKeys.length` over-allocates (deduped elements are always fewer than the sum). Suggest using `math.max(lKeys.length, rKeys.length)` as a tighter bound.

PR #773, opened 2026-04-12
perf: add canDirectIterate fast path to fused ByteRenderer materializer

He-Pin:
## Motivation

The fused ByteRenderer (`materializeDirect`) path used by Scala Native bypasses the upickle Visitor interface for direct byte-level JSON rendering. However, it was missing the `canDirectIterate` optimization already present in `Materializer.materializeRecursiveObj` — meaning every object field went through `visibleKeyNames` allocation + `value()` HashMap lookup, even for simple inline objects.

Profiling showed **realistic2 spends 99% of its time in materialization** (155ms materialize vs 1.7ms eval), processing ~125K objects and producing 28MB JSON output. This makes materializer optimization the highest-leverage target.

## Key Design Decision

Mirror the `canDirectIterate` fast path from `Materializer.scala` into `ByteRenderer.scala`, splitting `materializeDirectObj` into three specialized methods that avoid HashMap lookup for the common case of inline objects.

## Modification

Split `ByteRenderer.materializeDirectObj` into:
- **`materializeDirectInlineObj`**: Iterates raw `inlineFieldKeys`/`inlineFieldMembers` arrays directly, invoking members by index. Handles both multi-field and single-field objects.
- **`materializeDirectSortedInlineObj`**: Uses `_sortedInlineOrder` cached sort order (shared across all objects from same MemberList) for sorted output.
- **`materializeDirectGenericObj`**: Fallback to `visibleKeyNames` + `value()` for complex objects with super chains or excludedKeys.

### JMH (JVM, Scala 3.3.7)

Baseline: master @ `0d13274`

| Benchmark | Before (ms) | After (ms) | Change |
|-----------|-------------|-----------|--------|
| **realistic2** | **57.541** | **49.391** | **-14.2%** ✅ |
| comparison2 | 18.681 | 17.606 | -5.8% ✅ |
| base64Decode | 0.123 | 0.118 | -4.1% |
| bench.02 | 35.401 | 32.904 | -7.1% |
| reverse | 6.717 | 6.883 | +2.5% (noise) |

### Scala Native (hyperfine, 15 runs, 5 warmup)

| Binary | realistic2 (ms) | Relative |
|--------|----------------|----------|
| **sjsonnet (this PR)** | **96.1 ± 2.4** | **1.00x** ✅ |
| jrsonnet (Rust) | 112.8 ± 5.5 | 1.17x slower |
| sjsonnet (master) | 171.8 ± 5.6 | 1.79x slower |

**sjsonnet now beats jrsonnet by 17% on realistic2!**

## Analysis

The `canDirectIterate` fast path eliminates:
1. **`visibleKeyNames` allocation**: No more `ArrayBuffer` → `Array` creation per object
2. **`value()` HashMap lookup**: No more key-based cache lookup per field (replaced by direct index invocation)
3. **Validation checks**: Inline fields skip the `value()` validation path

For realistic2 with 125K objects, this removes ~125K HashMap lookups and ~125K array allocations from the hot materialization loop.

## References

- Mirrors `Materializer.materializeInlineObj` / `materializeSortedInlineObj` logic
- Related profiling: `sjsonnet --debug-stats bench/resources/cpp_suite/realistic2.jsonnet`

## Result

All 420 tests pass across JVM/JS/WASM/Native × Scala 2.12/2.13/3.3.7.

### DIFF
--- a/sjsonnet/src/sjsonnet/ByteRenderer.scala
+++ b/sjsonnet/src/sjsonnet/ByteRenderer.scala
@@ -2,6 +2,7 @@ package sjsonnet
 
 import java.io.OutputStream
 
+import scala.inline
 import upickle.core.{ArrVisitor, ObjVisitor}
 
 /**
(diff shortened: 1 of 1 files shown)


He-Pin (commented):
The overall approach is sound — mirroring `canDirectIterate` from Materializer into ByteRenderer is the right direction. I found a few correctness issues that need to be fixed:

1. `materializeDirectSortedInlineObj` sorted order computation is incorrect for the fallback case.

When `_sortedInlineOrder` is null, you call `Materializer.computeSortedInlineOrder(rawKeys, rawMembers)`. But unlike `materializeInlineObj` which first counts visible fields, `materializeDirectSortedInlineObj` returns `order.length` as `visCount`. However, `computeSortedInlineOrder` filters out hidden fields, so `order.length` is the count of *visible* fields, not the total. This is correct.

BUT there's a subtle issue: `computeSortedInlineOrder` expects the raw arrays and returns indices of *visible* fields in sorted order. If a field is hidden, its index is not in the returned array. So iterating `order` and invoking `rawMembers(idx)` is correct.

However, I need to verify: does `computeSortedInlineOrder` handle the case where ALL fields are hidden? In that case it returns an empty array, `visCount = 0`, and the while loop doesn't execute. The resulting output is `{ }` (empty object), which is correct.

2. `_skipFieldCache` check inconsistency between paths.

In `materializeDirectInlineObj` and `materializeDirectSortedInlineObj`, you check `if (!obj._skipFieldCache) obj.cacheFieldValue(...)` per field. In `materializeDirectGenericObj`, you don't — you use `obj.value(key, ctx.emptyPos)` which internally handles caching. This is correct because `value()` already calls `cacheFieldValue` internally.

3. `materializeDirectSortedInlineObj` else branch delegates to `materializeDirectInlineObj` for single-field objects.

When `rawKeys` is null (single-field object), the sorted path falls back to `materializeDirectInlineObj`. This is correct because sorting a single field is trivial. However, note that `materializeDirectInlineObj` checks `rawKeys != null` for multi-field, and handles single-field in the else branch. This delegation is fine.

4. Hidden field handling in `materializeDirectInlineObj` multi-field path.

When all fields are hidden, the while loop executes but never calls `markNonEmpty()` or `materializeChild`. The `wasEmpty` check at the end will be true, producing `{ }`. This is correct.

5. Potential issue: `materializeDirectInlineObj` multi-field path vs single-field path both invoke `cacheFieldValue`. 

In `materializeDirectInlineObj`, for the multi-field path (`rawKeys != null`), `cacheFieldValue` is called with `rawKeys(i)`. For the single-field path (else branch), it's called with `obj.singleKey`. Both are correct because these are the actual key names used for caching.

6. One correctness concern I want to flag: `ctx.emptyPos.fileScope` is used as the `FileScope` parameter for `m.invoke(...)`. In the original `Materializer.materializeInlineObj`, the same pattern is used (`ctx.emptyPos.fileScope`). This is correct because the `fs` parameter to `invoke` is only used for position information in error messages, and `emptyPos` is a valid sentinel.

Actually, after thorough review, I don't see correctness bugs. The implementation faithfully mirrors the Materializer pattern. The benchmark results confirm the performance improvement (14.2% on JMH, 1.79x -> 1.17x on Scala Native).

One style suggestion (not correctness): consider extracting the "open brace + reset" and "close brace + flush" boilerplate into helper methods to reduce duplication across the three materializeDirect* methods. Each of them has ~10 lines of identical setup/teardown. This doesn't affect performance but makes future maintenance easier.

He-Pin:
Thanks for merge this.

PR #780, opened 2026-04-15
perf: O(1) sort and equality for range arrays

He-Pin:
Motivation:

Avoid materializing and sorting arrays when the evaluator already knows an array is a numeric range with deterministic order.

Key Design Decision:

Use the post-#778 `RangeArr` representation directly rather than adding a separate range flag. This keeps the optimization aligned with the current array model.

Modification:

- Add `Arr.asSortedIfKnown` for O(1) sorted range detection.
- Add `Arr.rangeEquals` for O(1) equality on range arrays.
- Fast-path `std.set` sorting when the input range is already known sorted or can be reversed to sorted order.
- Add a direct `eq` check before array structural comparison.

Benchmark Results:

JMH throughput, normalized from `ms/op` to `ops/ms` (`ops/ms = 1 / ms_per_op`; higher is better):

| Benchmark | master ms/op | PR ms/op | master ops/ms | PR ops/ms | Delta ops/ms |
| `bench.06` | 0.216 | 0.189 | 4.63 | 5.29 | +14.3% |
| `setUnion` | 0.608 | 0.581 | 1.64 | 1.72 | +4.7% |
| `comparison` | 0.028 | 0.029 | 35.71 | 34.48 | -3.4% |

Scala Native hyperfine against source-built jrsonnet (`git@github.com:CertainLach/jrsonnet.git`, origin/master `80cd36a`; lower is better):

| `bench.06` | 6.16 +/- 0.50 ms | 5.80 +/- 1.33 ms | 4.30 +/- 0.34 ms | PR faster than master; jrsonnet faster |
| `setUnion` | 6.07 +/- 0.52 ms | 5.94 +/- 0.86 ms | 4.18 +/- 0.39 ms | PR slightly faster than master; jrsonnet faster |

Analysis:

The core change is semantics-preserving because it only skips work for arrays whose sorted order is known from their `RangeArr` representation. JMH is positive on the intended range/set workloads, and Native hyperfine is also directionally positive though jrsonnet still has a lower startup/runtime cost on these small cases.

References:

- PR branch: `perf/jrsonnet-optimizations`
- Head: `e950edb0c362f46c8c3967f1b209e3a3d395040e`
- Inspired by jrsonnet range-array representation.
- Benchmark toolchain: `./mill --no-server bench.runRegressions`; `hyperfine --warmup 10 --min-runs 50 -N`; source-built `jrsonnet 0.5.0-pre98`

Result:

Ready. JMH is positive and Native hyperfine is non-negative on the targeted cases.

### DIFF
--- a/sjsonnet/src/sjsonnet/Val.scala
+++ b/sjsonnet/src/sjsonnet/Val.scala
@@ -652,20 +652,21 @@ object Val {
      * it directly. If it is a reversed range, return the equivalent forward range in O(1). Returns
      * null if sort order is unknown and a full sort is needed.
      */
-    private[sjsonnet] def asSortedIfKnown(newPos: Position): Arr = {
-      if (isRange) {
-        if (_reversed) reversed(newPos) else this
-      } else null
+    private[sjsonnet] def asSortedIfKnown(newPos: Position): Arr = this match {
+      case _: RangeArr => if (_reversed) reversed(newPos) else this
+      case _           => null
     }
 
     /**
      * O(1) equality check for range arrays. Two ranges are equal if they describe the same integer
      * sequence (same start, length, and direction). Returns false if either array is not a range.
      */
-    private[sjsonnet] def rangeEquals(other: Arr): Boolean = {
-      isRange && other.isRange && _length == other._length &&
-      _reversed == other._reversed && _rangeFrom == other._rangeFrom
-    }
+    private[sjsonnet] def rangeEquals(other: Arr): Boolean =
+      (this, other) match {
+        case (r1: RangeArr, r2: RangeArr) =>
+          r1._length == r2._length && r1._reversed == r2._reversed && r1.rangeFrom == r2.rangeFrom
+        case _ => false
+      }
 
     /**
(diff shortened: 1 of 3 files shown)


sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -674,6 +674,26 @@ object Val {
...
+      } else null
+    }
+
+    /**
+     * O(1) equality check for range arrays. Two ranges are equal if they describe the same integer
+     * sequence (same start, length, and direction). Returns false if either array is not a range.
+     */
+    private[sjsonnet] def rangeEquals(other: Arr): Boolean = {
+      isRange && other.isRange && _length == other._length &&
+      _reversed == other._reversed && _rangeFrom == other._rangeFrom
```
He-Pin: would be better do  with patterns after 778

He-Pin: Processed by the rebase fix. The current PR no longer depends on the old `isRange` flag; it uses `RangeArr` pattern matching / subclass checks after the #778-era array representation changes. Resolving this outdated thread.

stephenamar-db (approved):

PR #784, opened 2026-04-18
perf: cache sorted visible key names on Val.Obj

He-Pin:
## Summary
- Cache the result of `visibleKeyNames.sorted(Util.CodepointStringOrdering)` on `Val.Obj` to avoid re-sorting on every materialization
- Non-inline objects (>8 fields, super chain, excludedKeys) previously called `.sorted()` every time they were materialized via `std.manifestJson` or the default output path
- The cached field uses `@volatile` for safe publication (benign race: worst case is a redundant sort)

## Changes
- `Val.scala`: Add `_sortedVisibleKeyNames` field and `sortedVisibleKeyNames` accessor method
- `Materializer.scala`: Use `obj.sortedVisibleKeyNames` instead of `obj.visibleKeyNames.sorted(...)` in both visitor-based and fused materializer paths
- `ByteRenderer.scala`: Use `obj.sortedVisibleKeyNames` in fused materializer generic object path

## Test plan
- [x] `./mill 'sjsonnet.jvm[3.3.7]'.test` — all tests pass
- [x] `./mill 'sjsonnet.jvm[3.3.7]'.compile` — compiles clean
- [x] Objects with `canDirectIterate` already use `_sortedInlineOrder` cache; this extends caching to the generic (non-inline) path

### DIFF
--- a/sjsonnet/src/sjsonnet/Val.scala
+++ b/sjsonnet/src/sjsonnet/Val.scala
@@ -884,7 +884,18 @@ object Val {
      * Cached sorted field order for inline objects. Shared across all objects from the same
      * MemberList to avoid per-object sort + allocation.
      */
-    @volatile private[sjsonnet] var _sortedInlineOrder: Array[Int] = null
+    private[sjsonnet] var _sortedInlineOrder: Array[Int] = null
+
+    private[sjsonnet] var _sortedVisibleKeyNames: Array[String] = null
+
+    private[sjsonnet] def sortedVisibleKeyNames: Array[String] = {
+      var r = _sortedVisibleKeyNames
+      if (r == null) {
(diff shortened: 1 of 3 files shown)


sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -886,6 +886,17 @@ object Val {
      */
     @volatile private[sjsonnet] var _sortedInlineOrder: Array[Int] = null
 
+    @volatile private[sjsonnet] var _sortedVisibleKeyNames: Array[String] = null
```
stephenamar-db: plrease remove volatiles. they cause so much performance issues, and we have no intention of making this thread safe.

stephenamar-db (changes requested):

stephenamar-db (approved):

PR #786, opened 2026-04-18
perf: change Expr from trait to abstract class

He-Pin:
Motivation:

Reduce hot-path evaluator dispatch overhead by moving `Expr` from a trait/interface shape to an abstract class hierarchy.

Key Design Decision:

Make `Val` extend `Expr` directly. This preserves optimized AST use cases for already-computed values while avoiding `Val with Expr` mixins on literals and functions.

Modification:

- Change `Expr` from `trait` to `abstract class`.
- Make `Val` extend both `Expr` and `Eval`.
- Simplify static optimizer matches from `Val with Expr` to `Val`.
- Update README architecture notes so `Expr`, `Eval`, and `Val` documentation matches the new hierarchy.

Benchmark Results:

JMH throughput, normalized from `ms/op` to `ops/ms` (`ops/ms = 1 / ms_per_op`; higher is better):

| Benchmark | master ms/op | PR ms/op | master ops/ms | PR ops/ms | Delta ops/ms |
| `realistic2` | 48.139 | 40.982 | 0.0208 | 0.0244 | +17.5% |
| `setUnion` | 0.603 | 0.594 | 1.66 | 1.68 | +1.5% |
| `bench.06` | 0.216 | 0.215 | 4.63 | 4.65 | +0.5% |
| `large_string_join` | 0.542 | 0.539 | 1.85 | 1.86 | +0.6% |
| `large_string_template` | 1.566 | 1.575 | 0.64 | 0.63 | -0.6% |
| `comparison2` | 17.397 | 17.686 | 0.0575 | 0.0565 | -1.6% |

Scala Native hyperfine against source-built jrsonnet (`git@github.com:CertainLach/jrsonnet.git`, origin/master `80cd36a`; lower is better):

| `realistic2` | 83.48 +/- 1.64 ms | 83.51 +/- 1.41 ms | 97.25 +/- 1.44 ms | Native neutral; sjsonnet faster than jrsonnet |
| `bench.06` | 6.19 +/- 0.59 ms | 6.11 +/- 0.47 ms | 4.47 +/- 0.40 ms | PR slightly faster than master; jrsonnet faster |

Analysis:

This is primarily a JVM dispatch optimization. The JMH signal is positive on `realistic2`, which is the important broad evaluator workload, while several tiny benchmarks are neutral or within noise. Scala Native is mostly neutral, which is expected because this change targets JVM trait/interface dispatch more than Native code generation.

References:

- PR branch: `perf/expr-abstract-class-pr`
- Head: `60f149dd00a04fba7dcbc469fc48d991a4aefbd8`
- Benchmark toolchain: `./mill --no-server bench.runRegressions`; `hyperfine --warmup 10 --min-runs 50 -N`; source-built `jrsonnet 0.5.0-pre98`

Result:

Ready. JMH is positive; Scala Native is neutral and remains open for future optimization.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -169,7 +169,7 @@ These three phases are encapsulated in the `sjsonnet.Interpreter` object.
 
 Some notes on the values used in parts of the pipeline:
 
-- `sjsonnet.Expr`: this represents `{...}` object literal nodes, `a + b` binary
+- `sjsonnet.Expr`: this abstract class represents `{...}` object literal nodes, `a + b` binary
   operation nodes, `function(a) {...}` definitions and `f(a)` invocations, etc..
   Also keeps track of source-offset information so failures can be correlated
   with line numbers.
@@ -182,9 +182,10 @@ Some notes on the values used in parts of the pipeline:
   raise an error because the first (erroneous) entry of the array is un-used and
   thus not evaluated.
 
-- Classes representing literals extend `sjsonnet.Val.Literal` which in turn extends
-  _both_, `Expr` and `Val`. This allows the evaluator to skip over them instead of
-  having to convert them from one representation to the other.
+- `sjsonnet.Val` extends `Expr`, and classes representing literals extend
+  `sjsonnet.Val.Literal`. This allows the evaluator to skip already-computed values
+  instead of having to convert them from one representation to the other, while
+  keeping the hot `Expr` dispatch path on an abstract class hierarchy.
 
 ## Performance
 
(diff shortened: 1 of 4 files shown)


He-Pin:
In Netty, the ByteBuf is an abstract class too.

He-Pin:
Need update the readme too

He-Pin:
Processed in the latest head. The README now documents `Expr` as an abstract class, explains that `Val` extends both `Eval` and `Expr`, and updates the hierarchy diagram so it matches this PR.

He-Pin:
@stephenamar-db wdyt

stephenamar-db (approved):

PR #785, opened 2026-04-18
perf: hybrid hot-path evaluator — up to 40% faster dispatch

He-Pin:
## Summary

- Profile all 66 benchmark files across 5 suites to identify ExprTag visit frequencies. **Top 7 types cover 96.1%** of all `visitExpr` calls: ValidId (30%), BinaryOp (21%), Val.Literal (18%), Select (13%), Apply1 (5%), ObjExtend (4%), IfElse (4%).
- Split `NewEvaluator.visitExpr` into a **hot path** (~120 bytecodes, 7 `instanceof` checks) and a **cold path** (`private visitExprCold` using `tag + @switch` for remaining 30 types).
- The hot path fits within JIT `FreqInlineSize=325` bytecodes, enabling **C2 to inline `visitExpr` into callers** (`visitBinaryOp`, `visitSelect`, etc.). The old evaluator's ~700-bytecode method body never gets inlined.
- Add `--new-evaluator` CLI flag for A/B testing.
- Add `EvaluatorBenchmark` (JMH) and `ExprTagProfile` profiling tool.

## JMH Results

Steady-state performance (1 fork, 8 warmup, 10 measurement iterations):

| Benchmark | Old (ms) | New (ms) | Delta |
|-----------|----------|----------|-------|
| bench.01 | 0.026 | 0.018 | **-31%** |
| bench.02 | 32.58 | 25.73 | **-21%** |
| bench.03 | 9.39 | 5.64 | **-40%** |
| gen_big_object | 0.928 | 0.715 | **-23%** |
| string_render_perf | 0.768 | 0.496 | **-35%** |
| base64_mega | 3.462 | 3.106 | **-10%** |
| realistic1 | 1.850 | 1.764 | **-5%** |
| heavy_string_render | 34.80 | 33.09 | **-5%** |
| realistic2 | 47.32 | 47.78 | ~tied |
| bench.04, 06, 08, 09 | - | - | ~tied |

Evaluator-heavy benchmarks (bench.01–03, gen_big_object, string_render_perf) show **21–40% improvement**. Builtin-dominated benchmarks (bench.04, foldl, comparison) are unaffected — the evaluator dispatch is not their bottleneck.

## Why it works

[...]

PR #788, opened 2026-04-25
perf: promote hot-path evaluator to default; remove dual-evaluator flag

He-Pin:
## Summary

Lifts the `NewEvaluator` hot-path dispatch (introduced in #785 behind `--new-evaluator`) into the default `Evaluator`, then removes the now-redundant subclass, `Settings.useNewEvaluator`, and the `--new-evaluator` CLI flag. Maintaining two evaluators doubled the surface for every future optimization and left the slow path as the user-facing default.

## Benchmarks

JMH (`EvaluatorBenchmark`, 18 files × 1 fork × 5 warmup × 3s + 10 measure × 3s; the single 1-fork outlier `bench.06` re-verified with 3 forks × 8 warmup + 15 measure):

| | count | examples |
|---|---|---|
| ✓ Strictly better (CIs do not overlap) | 7 | bench.01 -28.8%, bench.02 -27.2%, **bench.03 -37.6%**, gen_big_object -18.6% |
| = Within noise (CIs overlap) | 11 | realistic2 +1.3%, comparison2 -1.4%, ... |
| ✗ Strictly worse | **0** | (initial 1-fork run flagged bench.06 +4.0%; 3-fork rerun showed 0.186 ± 0.001 vs 0.186 ± 0.003 — a tie, the gap was a JIT compilation artifact) |

**Geometric mean: -7.27%**.

## Changes

- **`Evaluator.scala`** — `visitExpr` is now the 7-type instanceof hot path (covers ~96.1% of dispatches) plus a `visitExprCold` tag-based tableswitch. **Profiler instrumentation is preserved in the hot path** — `NewEvaluator` previously dropped it, so `--profile --new-evaluator` was silently broken.
- **`Settings.scala`** / **`Config.scala`** / **`SjsonnetMainBase.scala`** — drop the flag plumbing.
- **`Interpreter.createEvaluator`** — always returns the unified `Evaluator`.
- **`TestUtils`** / **`EvaluatorTests`** / **`AggressiveStaticOptimizationTests`** — drop the `useNewEvaluator` parameter; `EvaluatorTests` no longer runs every assertion twice.
- **`bench/EvaluatorBenchmark.scala`** — deleted; ongoing perf coverage stays in `MainBenchmark` and `RegressionBenchmark`.

Net diff: 9 files, +270 / -530.

## Breaking change

`Settings.useNewEvaluator` and the `--new-evaluator` CLI flag are removed. Embedders that explicitly set the flag should drop it; the resulting behavior is the same as the (now-default) hot-path dispatch.

## Test plan

- [x] `./mill 'sjsonnet.jvm[3.3.7]'.compile`
- [x] `./mill 'sjsonnet.js[3.3.7]'.compile`
- [x] `./mill 'sjsonnet.native[3.3.7]'.compile`
- [x] `./mill bench.compile`
- [x] `./mill 'sjsonnet.jvm[3.3.7]'.test` — 12 suites, 0 failures


He-Pin:
@stephenamar-db I have updated this pr to keep only one evaluator.

stephenamar-db (approved):

PR #792, opened 2026-04-26
fix: Ignore hidden fields in std.mapWithKey

He-Pin:
## Motivation

`std.mapWithKey` should match official Jsonnet and only map visible object fields. sjsonnet used `allKeyNames`, so hidden fields were mapped and re-emitted as visible fields.

This also did unnecessary work for hidden fields: creating mapped members and potentially invoking callbacks for fields that official Jsonnet never exposes through `std.mapWithKey`.

## Modification

- Use `visibleKeyNames` in `std.mapWithKey`.
- Preserve the existing single-field and inline-object construction fast paths.
- Add Scala unit coverage for hidden and forced-visible fields.
- Add a directional `new_test_suite` case validated with official Jsonnet, including a hidden-field non-evaluation assertion.

## Result

Hidden fields are no longer exposed by `std.mapWithKey`, forced-visible fields still map normally, and hidden fields avoid unnecessary callback/member evaluation work.

## Validation

- `jsonnet sjsonnet/test/resources/new_test_suite/map_with_key_hidden_fields.jsonnet`
- `./mill --no-server 'sjsonnet.jvm[3.3.7].test.testOnly' sjsonnet.StdMapTests sjsonnet.FileTests`
- `./mill --no-server 'sjsonnet.jvm[3.3.7].reformat'`
- `git diff --check`

## Notes

This is separate from the `std.objectRemoveKey` issue/PR and only covers the stdlib audit finding for `std.mapWithKey`.


He-Pin:
Four-implementation audit confirmation:

```jsonnet
std.mapWithKey(function(k, v) k + std.toString(v), {a::1, b:2})
```

| implementation | result |
| --- | --- |
| official Jsonnet 0.22.0 | `{ "b": "b2" }` |
| go-jsonnet 0.22.0 | `{ "b": "b2" }` |
| source-built jrsonnet 0.5.0-pre98 (`80cd36a`) | `{ "b": "b2" }` |
| sjsonnet PR #800 worktree, without this PR's change | `{ "a": "a1", "b": "b2" }` |

This confirms #792 is still a necessary stdlib compatibility fix and should either merge before #800 or be included when #800 is rebased/stacked. The source-level direction is also right: official `std.jsonnet` uses `std.objectFields(obj)`, which excludes hidden fields.

stephenamar-db (approved):

Issue #793, opened 2026-04-26
Track std.base64 non-ASCII string semantics vs official Jsonnet


He-Pin:
## Motivation

During the stdlib audit, `std.base64` / `std.base64Decode` showed a non-ASCII string semantics difference between sjsonnet and official C++ Jsonnet `v0.22.0`.

We are not changing this in the current stdlib correctness PR because this behavior may be visible to users and jrsonnet currently makes the same UTF-8 choice as sjsonnet. This issue tracks the discrepancy separately.

## Evidence

Official `jsonnet v0.22.0`:

```jsonnet
std.base64("é")                         // "6Q=="
std.base64Decode("w6k=")                 // "Ã©"
std.base64Decode("6Q==")                 // "é"
std.base64DecodeBytes(std.base64("é"))   // [233]
std.base64("Ā")                          // runtime error: Can only base64 encode strings / arrays of single bytes.
```

Current sjsonnet:

```jsonnet
std.base64("é")                         // "w6k="
std.base64Decode("w6k=")                 // "é"
std.base64Decode("6Q==")                 // "�"
std.base64DecodeBytes(std.base64("é"))   // [195, 169]
```

Local jrsonnet check:

```text
jrsonnet 0.5.0-pre98
commit 80cd36abd868507312e2cc2c78cb0f55a684c620
```

jrsonnet matches sjsonnet's UTF-8-byte behavior here:

```jsonnet
std.base64("é")                         // "w6k="
std.base64Decode("w6k=")                 // "é"
std.base64Decode("6Q==")                 // runtime error: bad utf8
std.base64DecodeBytes(std.base64("é"))   // [195, 169]
```

## Root Cause

Official C++ Jsonnet's stdlib implements string base64 as codepoint/char bytes:

```jsonnet
local bytes =
  if std.isString(input) then
    std.map(std.codepoint, input)
  else
    input;

base64Decode(str)::
  local bytes = std.base64DecodeBytes(str);
  std.join('', std.map(std.char, bytes)),
```

sjsonnet currently encodes string input as UTF-8 bytes and decodes bytes as UTF-8 strings.

## Proposed Direction

If sjsonnet decides to align strictly with official Jsonnet:

- Keep byte-array `std.base64` and `std.base64DecodeBytes` behavior unchanged.
- Change string `std.base64` to encode each character as a single byte, rejecting codepoints above `255`.
[...]

PR #794, opened 2026-04-26
fix: Align stdlib edge cases with Jsonnet

He-Pin:
Motivation:
Several stdlib functions diverged from official Jsonnet on lazy callback evaluation, ASCII-only string helpers, string set membership, numeric parsing beyond Long range, and signed zero handling.

Modification:
- Preserve `std.filterMap` callback laziness and avoid evaluating elements that the callbacks do not force.
- Reject non-string results from `std.flatMap` on strings, including `null`.
- Make `std.asciiUpper`, `std.asciiLower`, and `std.equalsIgnoreCase` ASCII-only instead of JVM Unicode case folding.
- Allow `std.setMember` to operate on string sets with set validation enabled.
- Parse `std.parseInt`, `std.parseOctal`, and `std.parseHex` into Jsonnet number semantics instead of Long-bounded integers.
- Normalize `std.sign(-0)` to numeric zero.
- Add focused utest coverage plus new direction tests under `new_test_suite`.

Result:
sjsonnet now matches official Jsonnet v0.22.0 for these stdlib edge cases while preserving performance-oriented implementations for hot paths.

Validation:
- `jsonnet --version` -> `Jsonnet commandline interpreter v0.22.0`
- Official `jsonnet` on new success direction tests returned `true`: filter_map_laziness, ascii_case_non_ascii, equals_ignore_case_ascii, set_member_string, parse_large_numbers, sign_negative_zero.
- Official `jsonnet` on `error.std.flatmap_string_null.jsonnet` errors with `std.flatMap on strings, provided function must return a string, got null`.
- `./mill --no-server sjsonnet.jvm[3.3.7].test` -> SUCCESS
- `./mill --no-server sjsonnet.jvm[3.3.7].checkFormat` -> SUCCESS
- `git diff --check` and `git diff --cached --check` -> clean

Notes:
- Base64 non-ASCII string semantics are intentionally not changed in this PR and are tracked separately in #793.
- The parse tests assert numeric equality semantics like official Jsonnet. This PR does not change sjsonnet large-number manifestation formatting.


He-Pin:
Additional audit note from the four-implementation matrix:

| expression | official Jsonnet 0.22.0 | go-jsonnet 0.22.0 | source-built jrsonnet 0.5.0-pre98 (`80cd36a`) | sjsonnet PR #800 worktree |
| `std.parseInt("+1")` | error | `1` | error | `1` |
| `std.parseHex("+10")` | error | error | error | `16` |

The PR #794 parse implementation appears to cover this because it routes parsing through `parseNat` and only special-cases `-`, matching official `std.jsonnet`:

```jsonnet
if str[0] == '-' then
  -parse_nat(str[1:], 10)
else
  parse_nat(str, 10)
```

So this is not a new issue if #794 lands, but it is worth keeping as an explicit direction test because go-jsonnet accepts `std.parseInt("+1")` while official Jsonnet and jrsonnet reject it.

stephenamar-db (approved):

PR #820, opened 2026-05-01
feat: async import loader for JS via static preload (#476)

### LINKED ISSUE #476: JavaScript: Proper way to use an async loader callback
I am using the JS version of Sjsonnet and am running into an issue trying to call `SjsonnetMain.interpret` with an async loader callback. I have `File`s that I want to read, but the only good way to read them these days is through `FileReader`, which only provides an async interface. Is there a way to use this with Sjsonnet, or do I have to preload all of my files’ contents as strings so that I can make the callback synchronous?

stephenamar-db:
Adds Preloader + ImportFinder to discover imports from each file's AST and drive caller-controlled (async) loading before synchronous evaluation. Exposes SjsonnetMain.interpretAsync for the JS build, accepting a Promise-returning loader so callers can use fetch / FileReader / etc.

Co-authored-by: Isaac

### DIFF
--- a/sjsonnet/src-js/sjsonnet/SjsonnetMain.scala
+++ b/sjsonnet/src-js/sjsonnet/SjsonnetMain.scala
@@ -168,9 +168,19 @@ object SjsonnetMain {
    * Async variant of [[interpret]]. Accepts an `importLoader` that returns a `Promise` of the file
    * contents, and returns a `Promise` resolving to the rendered output.
    *
-   * Implementation: imports are statically discovered by parsing each file's AST, loaded
-   * concurrently via the user-supplied async loader, and inserted into a cache. Once the transitive
-   * closure is loaded, evaluation runs synchronously against the cache.
+   * Imports are eagerly front-loaded: every `import`, `importstr`, and `importbin` reachable from
+   * the entry source (plus from any extVar/tlaVar code snippets) is statically discovered and
+   * loaded before evaluation begins. This includes imports inside branches the evaluator will never
+   * force, e.g. `if false then import 'x' else 1` will still ask the loader for `x`. The tradeoff
+   * is that all I/O happens up front, which is what lets evaluation run synchronously.
+   *
+   *   - Loader rejection (missing file, network error, etc.) fails the returned Promise.
+   *   - A parse error on a discovered (non-entry) file is tolerated; it only surfaces if evaluation
+   *     actually forces that branch.
+   *   - The entry source's own parse error fails immediately.
+   *
+   * Each discovered file is parsed once during preload and again referenced by the evaluator; the
+   * parsed AST is shared so fastparse runs only once per file.
    */
   @JSExport
   def interpretAsync(
--- a/sjsonnet/src-js/sjsonnet/SjsonnetMain.scala
+++ b/sjsonnet/src-js/sjsonnet/SjsonnetMain.scala
@@ -85,13 +85,26 @@ object SjsonnetMain {
         )
     }
 
+  /**
+   * Coerce a JS object whose values are strings into a `Map[String, String]`. Iterates the JS
+   * dictionary directly instead of round-tripping through `ujson` to avoid the intermediate ujson
+   * tree, the `.obj.toMap` copy, and the trailing `.map` on the immutable map.
+   */
   private def parseStringMap(label: String, value: js.Any): Map[String, String] =
     try {
-      ujson.WebJson.transform(value, ujson.Value).obj.toMap.map {
-        case (k, ujson.Str(v)) => (k, v)
-        case (k, _)            =>
-          throw js.JavaScriptException(s"$label '$k' must be a string value, got non-string")
+      val dict = value.asInstanceOf[js.Dictionary[js.Any]]
+      val out = Map.newBuilder[String, String]
+      out.sizeHint(dict.size)
+      val it = dict.iterator
+      while (it.hasNext) {
+        val (k, v) = it.next()
+        (v: Any) match {
+          case s: String => out += k -> s
+          case _         =>
+            throw js.JavaScriptException(s"$label '$k' must be a string value, got non-string")
+        }
       }
+      out.result()
     } catch {
       case e: js.JavaScriptException => throw e
       case e: Exception              =>
@@ -177,7 +190,8 @@ object SjsonnetMain {
    *   - Loader rejection (missing file, network error, etc.) fails the returned Promise.
    *   - A parse error on a discovered (non-entry) file is tolerated; it only surfaces if evaluation
    *     actually forces that branch.
-   *   - The entry source's own parse error fails immediately.
+   *   - The entry source's own parse error is reported through the normal `interpret0` formatting
+   *     path so the error shape and location info match synchronous `interpret`.
    *
    * Each discovered file is parsed once during preload and again referenced by the evaluator; the
    * parsed AST is shared so fastparse runs only once per file.
@@ -200,10 +214,11 @@ object SjsonnetMain {
       val wd = JsVirtualPath(wd0)
       val entryPath = JsVirtualPath("(memory)")
 
-      preloader.add(entryPath, StaticResolvedFile(text), ImportFinder.Kind.Code) match {
-        case Left(err) => return js.Promise.reject(err.getMessage)
-        case Right(_)  =>
-      }
+      // Don't propagate the entry's parse error here — let runInterpret surface it via
+      // interpret0 so the message goes through the same Error.formatError path as synchronous
+      // interpret (root frame, "(memory):line:col", etc.). If parsing the entry fails we still
+      // get an empty pending queue and a fast path to runInterpret, which fails identically.
+      preloader.add(entryPath, StaticResolvedFile(text), ImportKind.Code)
 
       // ext/tla vars are parsed as Jsonnet code (Interpreter.parseVar) and may contain imports.
       // Feed each value through the preloader using the same synthetic path layout so that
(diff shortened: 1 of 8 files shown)


stephenamar-db:
@He-Pin

sjsonnet/src/sjsonnet/Preloader.scala
```diff
@@ -0,0 +1,132 @@
...
+  private def discover(path: Path, content: ResolvedFile): Either[Error, Unit] = {
+    val parser = new Parser(path, internedStrings, internedFieldSets, settings)
+    try {
+      fastparse.parse(content.getParserInput(), parser.document(_)) match {
+        case f: Parsed.Failure =>
+          val traced = f.trace()
+          Left(new ParseError(s"$path: ${traced.msg}", offset = traced.index))
+        case Parsed.Success((expr, _), _) =>
+          ImportFinder.collect(expr).foreach { found =>
+            parentImporter.resolve(path, found.value) match {
```
He-Pin: This should resolve against `path.parent()`, not `path`. During evaluation `Importer.resolveAndReadOrFail` calls `resolve(pos.fileScope.currentFile.parent(), ...)`, so the preloader is passing a different `docBase` to the JS resolver than synchronous `interpret` does. A resolver that joins `docBase` + import name will miss nested imports: `dir/a.libsonnet` importing `b.libsonnet` gets looked up under `dir/a.libsonnet/b.libsonnet` instead of `dir/b.libsonnet`, leaving the cache incomplete before evaluation.

sjsonnet/src-js/sjsonnet/SjsonnetMain.scala
```diff
@@ -59,85 +135,101 @@ object SjsonnetMain {
...
-          js.Dynamic.global.console.error("Sjsonnet evaluation error:", msg)
-          throw js.JavaScriptException(msg)
-        case Right(v) => v
+      def loop(): Future[Unit] = {
+        val batch = preloader.takePendingImports()
+        if (batch.isEmpty) Future.successful(())
+        else Future.sequence(batch.map(loadOne)).flatMap(_ => loop())
       }
+
+      val result: Future[js.Any] = loop().map { _ =>
```
He-Pin, line 256: `loop()` only preloads imports reachable from the entry file AST. `runInterpret` still parses extVars/tlaVars lazily via `Interpreter.parseVar`, and those snippets can contain imports. With the cache-only importer, `std.extVar('cfg')` where `cfg` is `import 'lib.libsonnet'` fails with `Couldn't import file` unless the root file also imports that path. The async path needs to preload parsed ext/tla snippets as code roots too, or otherwise keep variable parsing from hitting a cache-only importer.

sjsonnet/src/sjsonnet/Preloader.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  private def discover(path: Path, content: ResolvedFile): Either[Error, Unit] = {
+    val parser = new Parser(path, internedStrings, internedFieldSets, settings)
+    try {
+      fastparse.parse(content.getParserInput(), parser.document(_)) match {
+        case f: Parsed.Failure =>
+          val traced = f.trace()
+          Left(new ParseError(s"$path: ${traced.msg}", offset = traced.index))
+        case Parsed.Success((expr, _), _) =>
+          ImportFinder.collect(expr).foreach { found =>
```
He-Pin, line 147: Collecting every import node here makes `interpretAsync` eager in places normal Jsonnet evaluation is lazy. For example, `if false then import 'bad.libsonnet' else 1` currently rejects if `bad.libsonnet` has a parse error, while synchronous `interpret` returns `1` because that branch is never forced. If static preloading is the intended tradeoff, this semantic difference should be explicit in the JS docs/API contract; otherwise this needs a design that does not parse/load imports from unforced branches.

He-Pin (commented):
Focused on the Scala.js async import semantics from #476. The added targeted tests pass, but I found a few cases where `interpretAsync` diverges from the existing synchronous `interpret` behavior.

He-Pin:
https://github.com/scala-js/scala-js/issues/4916
https://github.com/scala-js/scala-js/issues/5079

sjsonnet/src/sjsonnet/Preloader.scala
stephenamar-db: I don't think there's an alternative for this
The "right" way would be to find a way to make Lazy a Promise - but this is so invasive that it's not worth it.

He-Pin: Alternatively, one could introduce a library like Cats-effect, where `IO` is naturally lazy, and there is an `Eval` class in cats too. https://github.com/typelevel/cats/blob/main/core/src/main/scala/cats/Eval.scala

CertainLach: Given that imports in jsonnet are statically analyzable by design (They only permit strings, not arbitrary expressions) it seems that it is fine if they are not lazy

CertainLach: You won't be the only one with that behavior, the jrsonnet web bindings I have released use the same async import interface
https://delta.rocks/jrsonnet/playground?state=eyJ2ZXJzaW9uIjoxLCJlbnRyeSI6Im1haW4uanNvbm5ldCIsImZvcm1hdCI6Impzb24iLCJmaWxlcyI6W3sibmFtZSI6Im1haW4uanNvbm5ldCIsImNvbnRlbnQiOiJpZiB0cnVlIHRoZW5cbiAgMVxuZWxzZVxuICBpbXBvcnQgXCIyLmpzb25uZXRcIiJ9XX0

CertainLach: Wait, no, this is not the behavior I have implemented in mine:

> currently rejects if bad.libsonnet has a parse error, while synchronous interpret returns 1 because that branch is never forced

My implementation only fails if the resolver didn't return anything, the actual parsing and evaluation happens when the branch is actually hit

https://delta.rocks/jrsonnet/playground?state=eyJ2ZXJzaW9uIjoxLCJlbnRyeSI6Im1haW4uanNvbm5ldCIsImZvcm1hdCI6Impzb24iLCJmaWxlcyI6W3sibmFtZSI6Im1haW4uanNvbm5ldCIsImNvbnRlbnQiOiJpZiB0cnVlIHRoZW5cbiAgMVxuZWxzZVxuICBpbXBvcnQgXCIyLmpzb25uZXRcIlxuIn0seyJuYW1lIjoiMi5qc29ubmV0IiwiY29udGVudCI6IlN5bnRheCBlcnJvciJ9XX0

Ah, I get it
In my implementation the recursive analysis stops with the imported file leaf in case of parse error

stephenamar-db: good point, let me adjust.

sjsonnet/src-js/sjsonnet/SjsonnetMain.scala
```diff
@@ -49,6 +53,78 @@ object SjsonnetMain {
...
+        }
+      def read(path: Path, binaryData: Boolean): Option[ResolvedFile] =
+        throw new RuntimeException(
+          s"Importer.read should not be called during async preload (path=$path)"
+        )
+    }
+
+  private def parseStringMap(label: String, value: js.Any): Map[String, String] =
+    try {
+      ujson.WebJson.transform(value, ujson.Value).obj.toMap.map {
```
He-Pin: seems this will generate much allocation

sjsonnet/src/sjsonnet/ImportFinder.scala
```diff
@@ -0,0 +1,57 @@
...
+      def isCode: Boolean = true
+    }
+    case object Str extends Kind {
+      def binaryData: Boolean = false
+      def isCode: Boolean = false
+    }
+    case object Bin extends Kind {
+      def binaryData: Boolean = true
+      def isCode: Boolean = false
+    }
```
He-Pin: There are `ExternalVariableKind` ,it would be better to keep the name consitant, with a better name?

sjsonnet/src/sjsonnet/Preloader.scala
```diff
@@ -0,0 +1,150 @@
...
+   * imports.
+   *
+   * The returned `Either` reports parse errors for code files; binary or string imports never fail
+   * here.
+   */
+  def add(
+      path: Path,
+      content: ResolvedFile,
+      kind: ImportFinder.Kind = ImportFinder.Kind.Code): Either[Error, Unit] = {
+    cache.put(path, content)
```
He-Pin: This cache is keyed only by `Path`, but pending discovery is keyed by `(Path, Kind)`. A valid program can reference the same resolved path as both `importstr` and `importbin`, with one side even in an unforced branch. `interpretAsync` preloads both and whichever `cache.put` runs last overwrites the other; the cache-only importer then ignores `binaryData` and can hand a `StaticBinaryResolvedFile` to `importstr` or a text file to `importbin`. I reproduced a sync/async divergence with `if true then importstr "same" else importbin "same"`: sync returns the text, async rejects with `NotImplementedError`. The preload cache needs to account for import kind/binary mode or avoid overwriting incompatible entries.

sjsonnet/src-js/sjsonnet/SjsonnetMain.scala
```diff
@@ -59,85 +135,126 @@ object SjsonnetMain {
...
+      val parsedExtVars = parseStringMap("External variable", extVars)
+      val parsedTlaVars = parseStringMap("Top-level argument", tlaVars)
+
+      val parentImporter = jsResolveImporter(importResolver)
+      val preloader = new Preloader(parentImporter)
+      val wd = JsVirtualPath(wd0)
+      val entryPath = JsVirtualPath("(memory)")
+
+      preloader.add(entryPath, StaticResolvedFile(text), ImportFinder.Kind.Code) match {
+        case Left(err) => return js.Promise.reject(err.getMessage)
```
He-Pin: This rejects entry parse errors with only `err.getMessage`, which bypasses the normal `interpret0` error formatting used by synchronous `interpret`. The async path loses the root frame / `(memory):line:col` context and reports a different error shape for the entry file itself. Consider letting the normal interpreter parse the entry, or format this error through the same path so `interpretAsync` remains compatible with `interpret` for parse failures.

He-Pin (changes requested):
Found two async correctness issues and left inline comments: mixed import kinds can overwrite the preload cache for the same resolved path, and entry parse errors bypass the normal interpreter error formatting.

He-Pin (approved):
LGTM

PR #822, opened 2026-05-05
Add array eval copy API

He-Pin:
## Motivation

#821 adds slice views, but stdlib consumers still need one narrow way to bulk-copy arrays without each call site reimplementing `directBackingArray` / range / view branches.

This PR adds a shared `Eval` copy API that preserves laziness and gives fully-consumed array paths a simple JVM/Scala Native friendly implementation.

Constraints:

- copy references, not values
- do not force thunks
- keep loops simple enough for HotSpot and Scala Native
- avoid widening the API beyond current consumers
- keep the PR stacked and reviewable on top of #821

## Modification

Add a narrow `Arr.copyEvalTo` API for bulk copying `Eval` references without forcing `Val` values.

This PR routes these consumers through the API:

- `std.flattenArrays`
- array `std.flatMap`
- array-separator `std.join`

Specialized copy paths cover concat, repeat, slice, reversed lazy views, range, and byte arrays.

## Result

JMH, JVM harness, lower is better:

| Benchmark | master | this PR | Delta |
| --- | ---: | ---: | ---: |
| `sjsonnet_suite/array_copy_views` | 18.027 ms/op | 13.972 ms/op | -22.5% |

| Benchmark | master-native | this PR | latest jrsonnet | Result |
| `sjsonnet_suite/array_copy_views` | 23.0 +/- 1.6 ms | 9.1 +/- 3.1 ms | 9.4 +/- 3.0 ms | 2.53x faster than master; roughly even with latest jrsonnet under noisy CLI timing |

JIT / GC review:

- The API copies `Eval`, not `Val`, so laziness and callback/error retry behavior stay intact.
- `copyEvalTo(Array[Eval], offset)` lets hot consumers write into one preallocated array with indexed while-loops.
- Concat materialization now copies through the same API, so concat children can provide cheap copy paths without first materializing themselves.
- Consumer code no longer repeats polymorphic branches at every call site.

Verification:

- `git diff --check`
- `./mill --no-server 'sjsonnet.jvm[3.3.7].compile'`
- `./mill --no-server 'sjsonnet.jvm[2.13.18].compile'`
- `./mill --no-server 'sjsonnet.jvm[2.12.21].compile'`
- `./mill -i bench.runRegressions bench/resources/sjsonnet_suite/array_copy_views.jsonnet`

Rollback boundary:

- This PR is infrastructure plus three consumers.
- It does not change string join sizing, callback invocation, sort, or renderer behavior.
- If a consumer regresses, it can move back to the previous local copy path while keeping the API for other consumers.

## References

- Builds on #821 lazy slice array view.
- Follow-up stack: #823 presizes selected copy consumers.
- Head: `4f65897fff2ea8f736cc569de356d963ae78ed7e`

### DIFF
--- a/bench/resources/sjsonnet_suite/array_copy_views.jsonnet
+++ b/bench/resources/sjsonnet_suite/array_copy_views.jsonnet
@@ -1,0 +1,24 @@
+local block = std.range(0, 4095);
+local pair = block + block;
+local mixed = [
+  std.repeat(block, 64),
+  pair,
+  block[100:4000:2],
+  std.repeat(pair, 16),
+];
+local flattened = std.flattenArrays(mixed);
+local joined = std.join([], mixed);
+local len = std.length(flattened);
+local probes = std.range(0, 39999);
+{
+  len: len,
+  joinedLen: std.length(joined),
+  first: flattened[0],
+  middle: flattened[std.floor(len / 2)],
+  last: flattened[len - 1],
+  checksum: std.foldl(
+    function(acc, i) acc + flattened[(i * 7919) % len],
+    probes,
+    0
+  ),
+}
--- a/bench/resources/sjsonnet_suite/array_copy_views.jsonnet.golden
+++ b/bench/resources/sjsonnet_suite/array_copy_views.jsonnet.golden
@@ -1,0 +1,8 @@
+{
+   "checksum": 81774965,
+   "first": 0,
+   "joinedLen": 403358,
+   "last": 4095,
+   "len": 403358,
+   "middle": 975
+}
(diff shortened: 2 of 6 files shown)


He-Pin:
Closing obsolete/conflicting stack PR. This was infrastructure for array-copy synthetic/sjsonnet_suite workloads and is not a current high-priority docs/jrsonnet-gap PR.

He-Pin:
Reopened and rebased onto current upstream/master. This remains draft: the Arr.copyEvalTo idea is useful infrastructure, but it needs fresh current benchmark attribution before ready.

He-Pin:

PR #823, opened 2026-05-05
Presize array copy consumers

He-Pin:
## Motivation

#822 gives consumers a cheap `Eval` copy API, but `std.flattenArrays` and array-separator `std.join` can still pay `ArrayBuilder` growth/copy costs when the outer array has a modest number of large child arrays.

This PR adds a guarded two-pass pre-size path for those consumers. The goal is to remove avoidable intermediate allocation in few-large-array workloads without regressing many-small-array workloads.

Constraints:

- do not force element values
- avoid many-small-array regressions
- guard total length before allocation
- keep hot paths as straight indexed loops
- keep this PR narrowly stacked on #822

## Modification

Stacked on #822.

Use `Arr.copyEvalTo` to presize high-volume array-copy consumers:

- `std.flattenArrays`
- array-separator `std.join`

The pre-sized path uses two linear scans only when the outer part count is modest (`<= 1024`). Large outer arrays fall back to the one-pass `ArrayBuilder + copyEvalTo` path from #822.

## Result

JMH, JVM harness, lower is better:

| Benchmark | master | #822 | this PR | Delta vs #822 |
| --- | ---: | ---: | ---: | ---: |
| `sjsonnet_suite/array_copy_views` | 18.027 ms/op | 13.972 ms/op | 8.068 ms/op | -42.3% |

| Benchmark | master-native | #822-native | this PR | latest jrsonnet | Result |
| --- | ---: | ---: | ---: | ---: | --- |
| `sjsonnet_suite/array_copy_views` | 23.0 +/- 1.6 ms | 9.1 +/- 3.1 ms | 6.6 +/- 1.8 ms | 9.4 +/- 3.0 ms | this PR is 3.49x faster than master; CLI timings are outlier-heavy |
| `realistic2` | 91.4 +/- 13.0 ms | n/a | 83.0 +/- 1.4 ms | 136.6 +/- 1.8 ms | this PR is 1.65x faster than latest jrsonnet |

JIT / GC review:

- The second pass copies `Eval` references into one preallocated `Array[Eval]`; it does not force element values.
- `totalLen` is accumulated as `Long` and checked before allocating the final `Array[Eval]`.
- `PresizedCopyMaxParts = 1024` avoids turning many-small arrays into an always-two-pass workload.
- The fallback path preserves #822 behavior for large outer arrays.
- The hot path is simple counted while-loops plus `copyEvalTo`, so it stays friendly to JIT inlining and Scala Native codegen.

Verification:

- `git diff --check`
- `./mill --no-server 'sjsonnet.jvm[3.3.7].compile'`
- `./mill --no-server 'sjsonnet.jvm[2.13.18].compile'`
- `./mill --no-server 'sjsonnet.jvm[2.12.21].compile'`
- `./mill -i bench.runRegressions bench/resources/sjsonnet_suite/array_copy_views.jsonnet`

Rollback boundary:

- This PR only changes fully-consumed array-copy consumers.
- It does not change string join, renderer, sort, callback invocation, or global array view semantics.
- If a workload shows a many-small regression, the threshold can be lowered or the affected consumer can use the #822 one-pass path.

## References

- Builds on #822 array eval copy API.
- Head: `a5fb04737be227d73a3a08331534acfaf108cc1d`

### DIFF
--- a/sjsonnet/src/sjsonnet/stdlib/ArrayModule.scala
+++ b/sjsonnet/src/sjsonnet/stdlib/ArrayModule.scala
@@ -41,6 +41,7 @@ object ArrayModule extends AbstractFunctionModule {
 
   private val DefaultKeyF = Val.Null(dummyPos)
   private val DefaultOnEmpty = Val.Null(dummyPos)
+  private val PresizedCopyMaxParts = 1024
 
   @inline private def isDefaultKeyF(v: Val): Boolean = v.asInstanceOf[AnyRef] eq DefaultKeyF
   @inline private def isDefaultOnEmpty(v: Val): Boolean =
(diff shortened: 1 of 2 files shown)


He-Pin:
Closing obsolete/conflicting stack PR. It builds on #822 and mainly targets array_copy_views; not a current high-priority docs/jrsonnet-gap PR.

He-Pin:
Reopened and rebased onto current upstream/master. This remains draft and stacked conceptually on #822; keep for the presized-copy idea, but do not treat as ready without fresh current benchmark attribution.

He-Pin:

stephenamar-db:
rebase

He-Pin:
@stephenamar-db rebased

PR #826, opened 2026-05-07
perf: optimize identity function composition

He-Pin:
Motivation:

`bench/resources/cpp_suite/bench.07.jsonnet` builds a lazy array of identity-equivalent composed functions. Current master evaluates the deep function/lazy chain normally; it is slow on JVM and still overflows Scala Native even with `--max-stack 100000`.

Key Design Decision

Recognize only statically safe identity-equivalent unary functions and keep the runtime check conservative. Direct identity (`function(x) x`) and the exact non-tailstrict self-composition shape (`function(x) g(g(...g(x)))`) can be elided only after the captured `g` is proven effectively identity. Recursive composition cycles are treated as non-identity so normal max-stack behavior is preserved.

Modification:

- Tag identity/self-composition function shapes in `StaticOptimizer`.
- Forward the tags from `Evaluator` into `Val.Func` instances.
- Add an iterative cached `isEffectivelyIdentity` probe with explicit Unknown/Yes/No/InProgress byte states.
- Short-circuit unary function application when the callee is effectively identity.
- Preserve laziness, tailstrict forcing, non-function errors, non-identity composition, and recursive-cycle behavior with evaluator tests.

Benchmark Results:

JMH on JDK 21, `./mill -j 1 bench.runRegressions bench/resources/cpp_suite/bench.07.jsonnet`, lower `ms/op` is better; `ops/ms = 1 / ms_per_op`, higher is better.

| Case | master ms/op | PR ms/op | master ops/ms | PR ops/ms | Delta |
| `cpp_suite/bench.07.jsonnet` | 2.447 | 0.039 | 0.409 | 25.64 | +6173% ops/ms (~62.7x faster) |

Scala Native hyperfine, lower is better, comparing locally built Scala Native binary and source-built `jrsonnet 0.5.0-pre98`.

| Target | Result |
| --- | --- |
| master native | `StackOverflowError` with `--max-stack 100000` |
| PR native | 5.2 ± 1.1 ms |
| jrsonnet (`-s 1000000`) | 28.4 ± 1.9 ms |
| PR vs jrsonnet | PR is ~5.46x faster |

Analysis:

This optimization intentionally does not make arbitrary function equality assumptions. It only uses static shape tags plus runtime proof of the captured base function's effective identity. Errors and `tailstrict` are preserved: constructing `f2(error ...)` remains lazy, calling it forces the original error, and `tailstrict` still forces eagerly.

References:

- PR: https://github.com/databricks/sjsonnet/pull/826
- Rebased head: `50403de6fd8689a4cefb3cd39d280821ed33ff7b`
- Base: `0ae7b78a93c4e643f9bcfb6dd1d99d9fe7a522a9`

Result:

- `./mill --no-server -j 1 __.reformat && ./mill --no-server -j 1 __.test` passed locally.
- `./mill --no-server -j 1 'sjsonnet.native[3.3.7]'.nativeLink` passed locally.
- PR comments were reviewed; there are no review threads. The prior safety-update comment remains accurate after rebase.

### DIFF
--- a/sjsonnet/src/sjsonnet/Evaluator.scala
+++ b/sjsonnet/src/sjsonnet/Evaluator.scala
@@ -100,7 +100,15 @@ class Evaluator(
       case ExprTags.Lookup        => visitLookup(e.asInstanceOf[Lookup])
       case ExprTags.Function      =>
         val f = e.asInstanceOf[Function]
-        visitMethod(f.body, f.params, f.pos)
+        val func = visitMethod(f.body, f.params, f.pos)
+        // Forward StaticOptimizer-computed identity tags so apply1 can elide identity calls.
+        // Skip the writes when the optimizer didn't classify this Function (the common case),
+        // keeping zero overhead for non-identity lambdas.
+        if (f.staticIdentityShape != 0) {
+          func.staticIdentityShape = f.staticIdentityShape
+          func.staticIdentityCapturedIdx = f.staticIdentityCapturedIdx
+        }
+        func
       case ExprTags.LocalExpr            => visitLocalExpr(e.asInstanceOf[LocalExpr])
       case ExprTags.Apply                => visitApply(e.asInstanceOf[Apply])
       case ExprTags.Apply3               => visitApply3(e.asInstanceOf[Apply3])
--- a/sjsonnet/src/sjsonnet/Expr.scala
+++ b/sjsonnet/src/sjsonnet/Expr.scala
@@ -387,6 +387,15 @@ object Expr {
   }
   final case class Function(var pos: Position, params: Params, body: Expr) extends Expr {
     final override private[sjsonnet] def tag = ExprTags.Function
+
+    // Set by StaticOptimizer; remains 0 unless the body is statically recognized.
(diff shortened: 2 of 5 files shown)


He-Pin:
Safety update pushed in `53556cfd`:

- Added an `IdentityChecking` state so identity-composition probing detects recursive composition cycles and falls back to normal application instead of spinning in the optimization path.
- Added self-recursive and object-mutual-recursive regression tests.

Validation:
- `./mill -i sjsonnet.jvm[3.3.7].test.testOnly sjsonnet.EvaluatorTests`
- `./mill -i sjsonnet.native[3.3.7].test.testOnly sjsonnet.EvaluatorTests`
- `./mill -i bench.runRegressions bench/resources/cpp_suite/bench.07.jsonnet` -> `0.039 ms/op`
- `./mill -i __.checkFormat`

He-Pin:
Rebased onto latest master and refreshed local validation/benchmark data in the PR description. Master still overflows on the Scala Native bench.07 case; the rebased PR passes and remains strongly positive.

PR #827, opened 2026-05-07
perf: speed up base64 makeArray byte patterns

He-Pin:
## Motivation

`std.base64` can encode an array of byte values. Generated byte arrays from common `std.makeArray(..., function(i) (... i ... ) % 256)` patterns currently allocate one `Val.Num` per element before base64 can copy them into a `byte[]`.

This PR targets that generated-array path. The jrsonnet docs base64 byte-array benchmark is a large literal array, so it is tracked separately and is not treated as the same workload.

## Modification

- Recognize pure non-negative linear `std.makeArray` bodies ending in `% 256`.
- Represent those arrays with a lazy byte-pattern array view.
- Let `std.base64` request raw bytes directly when the array can provide them.
- Keep ordinary array access lazy and fall back to the generic path for non-matching or non-exact expressions.

## Result

Rebased/current against latest `upstream/master` at `8b67cb1e`.

Targeted JMH for the generated byte-array path, lower is better:

| benchmark | upstream/master | this PR | delta |
| --- | ---: | ---: | ---: |
| `go_suite/base64_stress.jsonnet` | 0.168 ms/op | 0.064 ms/op | -61.9% |

Docs-aligned base64 context against latest source-built jrsonnet master `5b43fa8` (`jrsonnet 0.5.0-pre98`), Scala Native hyperfine:

| docs case | sjsonnet native master | latest jrsonnet | result |
| --- | ---: | ---: | --- |
| `std.base64` | 5.0 +/- 0.9 ms | 2.7 +/- 0.4 ms | jrsonnet remains faster |
| `std.base64 (byte array)` | 6.5 +/- 0.3 ms | 14.7 +/- 2.2 ms | sjsonnet remains 2.27x faster |

The docs cases do not show an attributable PR delta, so they are listed only as external context. The ready signal for this PR is the generated `std.makeArray` byte-stream suite above.

## Verification

- `./mill -i __.checkFormat`
- `git diff --check`
- `./mill -i 'sjsonnet.js[3.3.7].test'`
- `JDK_JAVA_OPTIONS="--enable-native-access=ALL-UNNAMED -Xmx8G -XX:+UseG1GC" ./mill -i 'sjsonnet.native[3.3.7].test'`
- `./mill -i bench.runRegressions bench/resources/go_suite/base64_stress.jsonnet`
- `./mill -i bench.runRegressions bench/resources/go_suite/base64.jsonnet bench/resources/go_suite/base64Decode.jsonnet bench/resources/go_suite/base64DecodeBytes.jsonnet bench/resources/go_suite/base64_byte_array.jsonnet`
- `./mill --no-server 'sjsonnet.native[3.3.7]'.nativeLink`

## Boundary Checks

- `base64_stress` is not a jrsonnet docs benchmark; it covers the generated-byte-array path this PR targets.
- The docs `std.base64 (byte array)` case is a literal array, not a `std.makeArray` pattern.
- Only pure AST patterns are recognized: the index parameter, non-negative integer literals, `+`, `*`, and final `% 256`.
- Expressions that can exceed the exact integer range of `Double` fall back to the generic evaluator.
- Construction remains lazy; the byte array is materialized only when `std.base64` requests raw bytes.

## References

- Head: `d8b4db85fcbf7b4936b5d8ec0b5c4ba542809203`

### DIFF
--- a/sjsonnet/src/sjsonnet/Val.scala
+++ b/sjsonnet/src/sjsonnet/Val.scala
@@ -621,6 +621,8 @@ object Val {
       }
     }
 
+    private[sjsonnet] def rawBytes: Array[Byte] = null
+
     /**
      * If both this and other are ConcatViews sharing the same left array, return the shared prefix
      * length. Otherwise return 0. Used by compare/equal to skip identical prefix elements entirely,
@@ -1344,7 +1346,7 @@ object Val {
     @inline private def isMaterialized: Boolean = arr ne null
 
     /** Raw byte backing data for zero-copy extraction (e.g. base64 encode). Always non-null. */
-    def rawBytes: Array[Byte] = byteData
+    override def rawBytes: Array[Byte] = byteData
 
     override def value(i: Int): Val = {
       if (isMaterialized || isConcatView) super.value(i)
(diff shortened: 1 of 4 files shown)


He-Pin:
Closing as low priority for the current jrsonnet/docs gap work. The standout result is from base64_stress, which is not a jrsonnet docs benchmark; current docs base64 cases do not show an attributable standout delta.

He-Pin:
Reopened. This is not negative; keep it as draft for the generated std.makeArray byte-stream path. It should not be treated as a current docs/jrsonnet-gap ready PR because the standout benchmark is base64_stress rather than a jrsonnet docs case.

He-Pin:

stephenamar-db:
rebase

stephenamar-db (approved):

PR #825, opened 2026-05-07
perf: speed up constant string joins

He-Pin:
Motivation:

`std.join` over repeated constant string arrays (for example `std.join('', std.repeat(['e'], n))`) currently walks every element and repeatedly appends the same separator/value pair. This PR keeps the optimization focused on that repeated/constant-array case while preserving Jsonnet laziness and null-skipping behavior.

Key Design Decision

Expose a best-effort `constantEval` from array views and use it only when `std.join` can prove every element is the same already-known string or null. If the element is lazy, non-string, or otherwise unsafe to inspect eagerly, the code falls back to the existing element-by-element join path.

Modification:

- Add constant/repeated array views for repeated values.
- Fast-path string `std.join` for constant/repeated string arrays with exact result sizing.
- Preserve existing pre-sized array-copy behavior for array separators.
- Add directional golden coverage for repeated strings, Unicode strings, null skipping, array joins, and zero-length lazy-error behavior.

Benchmark Results:

JMH on JDK 21, `./mill -j 1 bench.runRegressions ...`, lower `ms/op` is better; `ops/ms = 1 / ms_per_op`, higher is better.

| Case | master ms/op | PR ms/op | master ops/ms | PR ops/ms | Delta |
| --- | ---: | ---: | ---: | ---: | ---: |
| `cpp_suite/bench.09.jsonnet` | 0.041 | 0.037 | 24.39 | 27.03 | +10.8% ops/ms |
| `cpp_suite/large_string_join.jsonnet` | 0.547 | 0.252 | 1.83 | 3.97 | +117.1% ops/ms |
| `sjsonnet_suite/array_copy_views.jsonnet` guard rerun | 7.770 | 7.677 | 0.129 | 0.130 | +1.2% ops/ms |

Scala Native hyperfine, lower is better, comparing locally built Scala Native binaries and source-built `jrsonnet 0.5.0-pre98`.

| Case | master | PR | jrsonnet | Result |
| --- | ---: | ---: | ---: | --- |
| scaled repeated join (`std.repeat(['abcd'], 50000)`) | 8.1 ± 1.2 ms | 6.9 ± 1.0 ms | 4.8 ± 1.2 ms | PR is ~14.8% faster than master; still ~1.44x jrsonnet |
| `cpp_suite/large_string_join.jsonnet` | 7.0 ± 1.2 ms | 5.5 ± 1.0 ms | 5.2 ± 0.9 ms | PR is ~21.4% faster than master; roughly at jrsonnet parity |
[...]

PR #828, opened 2026-05-07
perf: speed up TOML manifestation

He-Pin:
Motivation:

`std.manifestTomlEx` spends avoidable time and allocation while rendering TOML: regex matching for bare keys, intermediate escaped strings, per-value renderer allocation, `Seq` path copies, `partition`, and repeated sorting during table traversal.

Key Design Decision

Keep this PR TOML-only. The earlier repeated-join work is now handled separately in #825, so this branch was intentionally reduced to a single focused TOML rendering commit after rebasing onto latest master.

Modification:

- Write TOML strings and keys directly to the existing `StringWriter`.
- Replace bare-key regex matching with a small ASCII scanner.
- Avoid string concatenations and unnecessary `StringWriter.flush()` calls while rendering arrays.
- Reuse a `TomlRenderer` per table level and keep table paths in a mutable buffer.
- Use sorted visible keys plus section flags instead of `partition` plus extra sorting.
- Add directional golden coverage for TOML key escaping, non-section ordering, nested tables, and table arrays.

Benchmark Results:

JMH on JDK 21, `./mill -j 1 bench.runRegressions bench/resources/go_suite/manifestTomlEx.jsonnet`, lower `ms/op` is better; `ops/ms = 1 / ms_per_op`, higher is better.

| Case | master ms/op | PR ms/op | master ops/ms | PR ops/ms | Delta |
| `go_suite/manifestTomlEx.jsonnet` | 0.067 | 0.053 | 14.93 | 18.87 | +26.4% ops/ms |

Scala Native hyperfine, lower is better, comparing locally built Scala Native binaries and source-built `jrsonnet 0.5.0-pre98`.

| Case | master | PR | jrsonnet | Result |
| `go_suite/manifestTomlEx.jsonnet` | 5.3 ± 1.3 ms | 5.3 ± 1.3 ms | 4.0 ± 1.1 ms | startup/noise dominated; PR and master are neutral |
| scaled TOML manifestation loop | 29.0 ± 1.6 ms | 12.1 ± 1.0 ms | 10.0 ± 0.9 ms | PR is ~58.3% faster than master; still ~1.21x jrsonnet |

Analysis:

The output remains compatible with current sjsonnet behavior: local master and this PR produced identical output on the benchmark inputs and the new directional TOML golden test. The scaled Native comparison includes jrsonnet as a performance reference, but jrsonnet's TOML formatting differs from sjsonnet on these inputs, so sjsonnet-vs-sjsonnet output equality is the semantic gate.

References:

- PR: https://github.com/databricks/sjsonnet/pull/828
- Rebased head: `3efa453a5b6f48d6dfd0e1cd034922d6c8e2ff03`
- Base: `0ae7b78a93c4e643f9bcfb6dd1d99d9fe7a522a9`
- Split dependency: #825 now owns the repeated constant join optimization.

Result:

- `./mill --no-server -j 1 __.reformat` passed locally.
- `./mill --no-server -j 1 __.test` passed locally.
- PR comments were reviewed; there are no review threads. The maintainer rebase request is addressed by this force-pushed TOML-only branch.


stephenamar-db:
rebase

He-Pin:
Rebased onto latest master and reduced this PR to TOML-only so it no longer duplicates #825's repeated-join work. I refreshed the local validation and benchmark data in the PR description.

PR #776, opened 2026-04-13
perf: comprehensive Scala Native render pipeline optimization

He-Pin:
Motivation:

This PR is the complete rebased render/string optimization branch for preserving the full #776 experiment on top of current `master`. It is intentionally broad and should remain draft/source material before we split the useful pieces into focused PRs.

Key Design Decision:

- Keep the full rebased experiment available on GitHub first, then split from it.
- Preserve compatibility-first behavior: any optimization that changes Jsonnet semantics must be excluded during later split PRs.
- Treat this PR as an archive/integration branch, not as a ready-to-merge unit.
- Use focused follow-up PRs for individual JIT/Native-friendly ideas with clean benchmark proof.

Modification:

- Rebased the full render pipeline branch onto `upstream/master` at `0ae7b78a93c4e643f9bcfb6dd1d99d9fe7a522a9`.
- Resolved current renderer/SWAR conflicts across JVM, Native, JS, and shared renderer code.
- Preserved the existing split/source ideas: SWAR/chunked escaping, char renderer paths, format scanning/assembly, ascii-safe propagation, stdlib string helpers, and manifest rendering work.
- Included the rebase repair for the `std.format` no-spec/single-spec path so offset-based scanned formats do not create invalid null strings.

Benchmark Results:

JMH command:

```bash
./mill --no-server -j 1 bench.runRegressions \
  bench/resources/cpp_suite/large_string_template.jsonnet \
  bench/resources/cpp_suite/large_string_join.jsonnet \
  bench/resources/cpp_suite/bench.09.jsonnet \
  bench/resources/jdk17_suite/repeat_format.jsonnet \
  bench/resources/go_suite/manifestTomlEx.jsonnet \
  bench/resources/go_suite/manifestJsonEx.jsonnet \
  bench/resources/go_suite/substr.jsonnet \
  bench/resources/go_suite/parseInt.jsonnet
```

| Case | master ms/op | #776 ms/op | Result |
| --- | ---: | ---: | --- |
| `large_string_template.jsonnet` | 1.265 | 1.205 | +5.0% ops/ms |
| `large_string_join.jsonnet` | 0.554 | 0.397 | +39.5% ops/ms |
| `bench.09.jsonnet` | 0.042 | 0.046 | regression/risk |
| `repeat_format.jsonnet` | 0.202 | 0.170 | +18.8% ops/ms |
| `manifestTomlEx.jsonnet` | 0.068 | 0.073 | regression/risk |
[...]

PR #834, opened 2026-05-10
perf: add ASCII-safe substr fast path

He-Pin:
Motivation:
Long printable ASCII strings can use UTF-16 length and substring indexes directly because every UTF-16 code unit is also one Jsonnet codepoint. `std.substr` currently still pays codepoint offset checks for these proven-safe long literals.

Key Design Decision:
Reuse the existing `_asciiSafe` runtime marker, but only set it for long (>1024 chars) string literals that are printable ASCII and JSON-render safe. This keeps parser overhead off short-string-heavy workloads and avoids broad runtime ASCII scans. JVM uses byte-level SWAR for the long-string safety scan; Scala Native and Scala.js keep a scalar single-pass scan because the byte-buffer SWAR variant regressed Native due UTF-8 allocation cost.

Modification:
- Add `CharSWAR.isAsciiJsonSafe` on JVM, Native, and JS.
- Use JVM byte-SWAR to check long strings for ASCII + JSON-safe bytes.
- Mark long ASCII JSON-safe literals in `Parser.makeStr`.
- Propagate `_asciiSafe` through `Val.Str.concat` when both operands are safe.
- Let `std.length` and `std.substr` use direct `String.length` / `substring` only for proven-safe strings.
- Add Unicode/ASCII regression coverage for long ASCII length, substr bounds, and concat propagation.

Benchmark Results:
JMH, lower is better (`./mill --no-server -j 1 bench.runRegressions ...`):

| benchmark | master ms/op | this PR ms/op | result |
| --- | ---: | ---: | ---: |
| `go_suite/substr.jsonnet` | 0.056 | 0.045 | ~20% faster |
| `jdk17_suite/split_resolve.jsonnet` | 0.147 | 0.146 | neutral |
| `cpp_suite/realistic2.jsonnet` | 44.529 | 45.024 | neutral/no stable regression signal |

Scala Native hyperfine, lower is better (`--shell=none --warmup 5 --runs 20`):

| command | mean | note |
| --- | ---: | --- |
| master native `go_suite/substr.jsonnet` | 7.7 ± 1.8 ms | baseline |
| this PR native `go_suite/substr.jsonnet` | 6.5 ± 1.0 ms | neutral/slightly faster vs master |
| jrsonnet `go_suite/substr.jsonnet` | 4.0 ± 0.6 ms | still ahead |

Analysis:
The broad #776 ASCII/string fast-path ideas were narrowed aggressively. Runtime ASCII scans in `std.length`/`std.substr` caused guard regressions, and all-platform UTF-8 byte-SWAR regressed Scala Native. This version uses byte-SWAR only where it measured well (JVM), keeps Native/JS allocation-free, and only takes fast paths when the marker is already proven, preserving Jsonnet codepoint semantics for all non-ASCII strings.

References:
- Source idea: https://github.com/databricks/sjsonnet/pull/776
- Narrowed from commit: https://github.com/He-Pin/sjsonnet/commit/a190a800

Result:
Draft PR for review. JVM substr benchmark is positive; Native is neutral/slightly positive with no-shell hyperfine; full `./mill --no-server -j 1 __.reformat && ./mill --no-server -j 1 __.test` passed locally.

### DIFF
--- a/sjsonnet/src/sjsonnet/Parser.scala
+++ b/sjsonnet/src/sjsonnet/Parser.scala
@@ -748,7 +748,10 @@ class Parser(
     // cost more than the potential memory savings for strings that are unlikely
     // to repeat (e.g., 600KB text block literals)
     val unique = if (s.length > 1024) s else internedStrings.getOrElseUpdate(s, s)
-    Val.Str(pos, unique)
+    val result = Val.Str(pos, unique)
+    if (unique.length > 1024 && CharSWAR.isAsciiJsonSafe(unique))
+      result._asciiSafe = true
+    result
   }
 
   // Any `expr` that isn't naively left-recursive
--- a/sjsonnet/src/sjsonnet/Val.scala
+++ b/sjsonnet/src/sjsonnet/Val.scala
@@ -433,11 +433,15 @@ object Val {
       if (ls != null && ls.isEmpty) return right
       if (rs != null && rs.isEmpty) return left
       // Small string eagerness: both flat and combined length <= 128
-      if (ls != null && rs != null && ls.length + rs.length <= 128)
-        return new Str(pos, ls + rs)
+      if (ls != null && rs != null && ls.length + rs.length <= 128) {
+        val result = new Str(pos, ls + rs)
+        if (left._asciiSafe && right._asciiSafe) result._asciiSafe = true
+        return result
+      }
       // Rope node: O(1)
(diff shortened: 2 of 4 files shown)


stephenamar-db:
rebase

He-Pin:
@stephenamar-db rebased

stephenamar-db (approved):

PR #861, opened 2026-05-16
perf: incremental std.join + asciiSafe coverage improvements

He-Pin:
## Motivation

Stacked on #858. Four small, independent perf wins around `std.join` and
`asciiSafe` propagation. None changes any user-visible semantics.

Together they shave a few percent off real-world manifest workloads on Native
where there is no JIT to recover the missed fast paths. Splitting from the
follow-up structural refactor (separate PR) so each lands on its own merits.

## Modification

Four atomic commits:

1. **SWAR-accelerate `isAsciiJsonSafe`** — replace the per-char range check with a
   word-at-a-time SWAR scan in `CharSWAR` (JVM/Native/JS variants). Same result,
   fewer branches per character on the hot ASCII-safety detection path.

2. **Drop `parts[]` allocation in `joinDirectStringArray`** — when every element of
   `std.join(sep, arr)` is already a `Val.Str`, append directly into the result
   buffer instead of materializing an intermediate `Array[String]` first.

3. **Cache `asciiSafe` in `joinedRepeatedString`** — when the input array is a
   single `Val.Str` repeated (the rare-but-hot single-element-join case), reuse
   its cached `asciiSafe` flag instead of re-scanning the produced string.

4. **Propagate `asciiSafe` through more `StringModule` builtins** — extend
   `Val.Str.asciiSafe(...)` factory usage to `std.char`, `std.asciiUpper`,
   `std.asciiLower`, `std.strReplace`, `std.stripChars`, `std.{l,r}stripChars`,
   `std.split`, `std.splitLimit`, `std.splitLimitR`. These outputs are now
   ByteRenderer fast-path eligible without re-scanning. New regression test
   `new_test_suite/string_asciisafe_propagation.jsonnet`.

### Hyperfine (Native, end-to-end)

```
kube-prometheus (jrsonnet/tests/realworld/vendor/.../kube-prometheus/example.jsonnet)
  base (PR #858 head):  150.1 ± 11.9 ms
  this PR:              138.8 ±  4.0 ms     1.08× faster
```

### JMH (JVM, steady-state, 2 forks × 5 iter × 3s)

```
RegressionBenchmark.main          base ms/op       this PR ms/op
large_string_join.jsonnet         0.310 ± 0.045    0.304 ± 0.018  ≈ flat
large_string_template.jsonnet     0.650 ± 0.035    0.734 ± 0.088  noisy (overlap)
repeat_format.jsonnet             0.144 ± 0.019    0.162 ± 0.047  noisy (overlap)
manifestJsonEx.jsonnet            0.056 ± 0.011    0.055 ± 0.013  flat
```

JVM JIT recovers most of the per-character work; the wins are concentrated on
Native, which matches PR #858. Per-bench JMH error bars overlap, so JVM
steady-state is best read as "no regression".

## Test plan

- `./mill 'sjsonnet.jvm[3.3.7]'.test` — green
- `./mill 'sjsonnet.native[3.3.7]'.test` — green
- New regression test covers all 10 newly-propagated builtins
- `./mill __.checkFormat` — green


He-Pin:
@stephenamar-db Rebased on the current master

PR #874, opened 2026-05-28
perf: speed up manifest JSON rendering

He-Pin:
## Motivation

`std.manifestJson`, `std.manifestJsonMinified`, and `std.manifestJsonEx` still routed through `StringWriter`, paying `StringBuffer` synchronization per `write` and per `flush` on the hot manifestation path. Source-built jrsonnet comparisons showed sjsonnet trailing on object-heavy manifest workloads.

## Modification

- Add `StringBuilderWriter`: an unsynchronized `Writer` over a `StringBuilder`.
- Add package-private `FastMaterializeJsonRenderer` backed by `StringBuilderWriter`; route the three `std.manifestJson*` builtins through it. Public `MaterializeJsonRenderer` ABI/shape unchanged.
- Fix codepoint comparison for raw surrogate prefixes: equal surrogate UTF-16 code units must be decoded before deciding ordering. `UnicodeHandlingTests` extended for the prefix-ordering case.

## Result

Scala Native hyperfine on kube-prometheus, `-N -w 4 -m 20`, jrsonnet HEAD `2d7eed05`:

| Workload (native) | Before | After | Δ |
|---|---:|---:|---:|
| kube-prometheus, sjsonnet | 158.4 ± 16.8 ms | 143.7 ± 3.2 ms | **−9.3%** |
| kube-prometheus, jrsonnet | 101.2 ± 4.4 ms | 97.4 ± 8.6 ms | reference |
| `manifestJsonEx`, sjsonnet | — | 5.09 ± 1.01 ms | new |
| `manifestJsonEx`, jrsonnet | — | 4.08 ± 1.40 ms | reference |

JMH regression post-PR: `manifestJsonEx` 0.055 ms/op, `realistic2` 43.6 ms/op, `gen_big_object` 0.842 ms/op.

Related: #666.

## Test plan

- [x] `./mill __.reformat`
- [x] `./mill -j 1 __.test` — 517/517 pass

---

## Follow-up stacked optimizations

Each commit below was verified for byte-identical output and measured before landing. Perf bar: JVM-positive and Native-non-regressing (changes that measured neutral/negative on Native — a YAML-renderer swap, a binary-operator Position deferral, and a first char-deboxing attempt — were measured and dropped).

| Commit | Change | JVM | Native |
|---|---|---:|---:|
| `skip escape scan for AsciiSafeStr` | char renderer emits `Val.AsciiSafeStr` without the SWAR escape scan | **+10%** render-only | neutral |
| `unsynchronized StringBuilderWriter in TomlRenderer` | drop `StringBuffer` sync on the `manifestTomlEx` path | +6–14% | **1.11× (≈+10%)** |
| `capture parse Position without boxing` | `Parser.Pos` writes the `Position` straight into fastparse `successValue` instead of `Index.map` (no Int box/unbox/closure per node) | **+5.4%** parse | **+4.5%** parse |
| `defer Position alloc in exprSuffix2` | allocate the suffix `Position` only on a matched suffix, not on every rep-terminating attempt | +1.9% parse | neutral |
| `flush FastMaterializeJsonRenderer only at root depth` | accumulate in-memory, emit once at `depth == 0`; 4 KB initial buffer | — | — |

Methodology: JVM via JMH (`ParserBenchmark`, plus isolated render benches added under `bench/`); Native via the binary's `--debug-stats` phase timing and interleaved hyperfine on kube-prometheus (cooled, min/p25). Render micro-wins (AsciiSafeStr) do not transfer to Native end-to-end because parse+eval dominate there; the parse-side and TOML changes do.


He-Pin:
Superseded — split into focused, independently-measured PRs off current master (each output byte-identical, no benchmark code):

- #875 — TomlRenderer → unsynchronized StringBuilderWriter (Native ~1.14×)
- #876 — Parser.Pos without boxing the offset Int (Native parse +6–8%, JVM +5.4%)
- #877 — defer Position alloc in exprSuffix2 (JVM +1.9%, Native neutral)

The manifest-JSON rendering work this PR was based on is already in master (da92dd15). Closing in favor of the smaller PRs above.

PR #875, opened 2026-05-30
perf: use unsynchronized StringBuilderWriter in TomlRenderer

He-Pin:
## Motivation

`std.manifestTomlEx` had three sources of avoidable overhead on the hot manifestation path:

1. **Synchronized writer.** `TomlRenderer` and `ManifestModule.evalRhs` rendered into a `java.io.StringWriter`, whose backing `StringBuffer` pays a monitor enter/exit on every `write`/`flush`. The `FastMaterializeJsonRenderer` already uses the unsynchronized `StringBuilderWriter` (#874); TOML did not.
2. **Redundant field lookups in `renderTableInternal`.** Each key's `Val.Obj.value(k)` was resolved twice — once to classify scalar vs section, then again to render or recurse. The cache deduplicates the result, but the lookup itself still costs.
3. **Wasted indexing work.** `visibleKeyNames` was iterated and each key binary-searched back into `sortedVisibleKeyNames` — `sortedVisibleKeyNames` can be iterated directly, skipping `O(n log n)` compares per table.

## Modification

Two commits:

- **`perf: use unsynchronized StringBuilderWriter in TomlRenderer`** — Swap `TomlRenderer` and the `manifestTomlEx` render path in `ManifestModule` from `java.io.StringWriter` to the package-private `StringBuilderWriter`. `std.deepJoin` keeps `StringWriter` (separate concern).
- **`perf: cache resolved field values and skip binary search in renderTableInternal`** — Resolve each field once into a `resolved: Array[Val]` during section classification and reuse it during render/recurse; iterate `sortedVisibleKeyNames` directly (removes the now-unused `sortedKeyIndex` binary search); hoist `childIndent = cumulatedIndent + indent` out of the section loop (was an identical allocation per sibling section); pre-size the output `StringBuilderWriter` to 1 KiB so small/medium outputs skip the first ~6 doublings.

Output is byte-identical (verified at 1,228,186 bytes on the benchmark workload).

## Result

Scala Native, hyperfine A/B against `master` (`fc292fa6`). Workload: object comprehension over 8000 small tables → ~1.2 MB TOML output (render-dominated). Four interleaved-order passes, `--warmup 10 --min-runs 100 --shell=none`:

| pass | order | before mean | after mean | before min | after min | **min ratio** |
|---|---|---:|---:|---:|---:|---:|
| 1 | before → after | 59.4 ±  2.7 ms | 53.2 ± 23.4 ms | 55.4 ms | 43.8 ms | **1.27×** |
| 2 | after → before | 64.1 ±  7.7 ms | 51.8 ± 12.2 ms | 56.4 ms | 43.7 ms | **1.29×** |
| 3 | before → after | 64.1 ±  8.1 ms | 53.2 ± 14.3 ms | 56.4 ms | 42.0 ms | **1.34×** |
| 4 | after → before | 63.3 ± 14.3 ms | 49.2 ±  3.7 ms | 57.2 ms | 42.8 ms | **1.34×** |

Mean is noisy on the host (1.12× – 1.29×), but **after is faster in every one of the 4 passes** and the **min values are tight at ~1.27–1.34× faster** (best observed: 42.0 ms vs 56.4 ms, ~25.5% reduction). Output byte-identical, 1,228,186 bytes both sides.

For comparison, the StringBuilderWriter swap alone (commit 1) measures ~1.08–1.14× min; the cache + binary-search elimination + childIndent hoist (commit 2) lifts that to ~1.27–1.34× min.

## Test plan

- [x] `./mill __.reformat`
- [x] `./mill 'sjsonnet.jvm[3.3.7]'.test` — 519/519 pass
- [x] Scala Native A/B hyperfine — 4 interleaved-order passes, all positive; output byte-identical

---

> Rebased onto current `master` (`fc292fa6`). The companion commit "speed up manifest JSON rendering" was merged separately as #879, so this PR now contains only the TomlRenderer / ManifestModule changes.

### DIFF
--- a/sjsonnet/src/sjsonnet/TomlRenderer.scala
+++ b/sjsonnet/src/sjsonnet/TomlRenderer.scala
@@ -2,22 +2,24 @@ package sjsonnet
 
 import upickle.core.{ArrVisitor, ObjVisitor, SimpleVisitor, Visitor}
 
-import java.io.StringWriter
-
+// Uses the unsynchronized [[StringBuilderWriter]] rather than java.io.StringWriter: the latter is
+// backed by a synchronized StringBuffer, paying a monitor enter/exit on every write/flush on the
+// hot manifestTomlEx path. Output is byte-identical. Same swap as the JSON renderer in #874.
 class TomlRenderer(
-    out: StringWriter = new java.io.StringWriter(),
+    out: StringBuilderWriter = new StringBuilderWriter(),
     cumulatedIndent: String,
     indent: String)
-    extends SimpleVisitor[StringWriter, StringWriter] {
+    extends SimpleVisitor[StringBuilderWriter, StringBuilderWriter] {
   override def expectedMsg: String = "unimplemented type in Materializer"
-  private object objectKeyRenderer extends upickle.core.SimpleVisitor[StringWriter, StringWriter] {
+  private object objectKeyRenderer
+      extends upickle.core.SimpleVisitor[StringBuilderWriter, StringBuilderWriter] {
     override def expectedMsg = "expected string"
 
-    override def visitNull(index: Int): StringWriter = {
+    override def visitNull(index: Int): StringBuilderWriter = {
       TomlRenderer.this.visitNull(index)
     }
 
-    override def visitString(s: CharSequence, index: Int): StringWriter = {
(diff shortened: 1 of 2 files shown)


He-Pin:
@stephenamar-db rebased

PR #881, opened 2026-05-30
perf: use commix concurrent GC for Native CLI binary

He-Pin:
## Summary

Switch the Native CLI release binary from the default **immix** GC to **commix** (concurrent immix). commix collects on background threads, overlapping collection with evaluation. This is a one-line `build.mill` change with **byte-identical output** and **unchanged RSS** — it still frees memory, so it stays bounded and safe on small machines.

## Why commix

While profiling the Native binary against [jrsonnet](https://github.com/CertainLach/jrsonnet), the remaining wall-clock gap turned out to be **GC/allocation cost, not interpreter dispatch**. immix's stop-the-world collections dominate on allocation-heavy configs and cause large latency variance. I benchmarked all four Scala Native GCs (`none`, `immix`, `commix`, `boehm`) on jrsonnet's realworld config suite (min ms, interleaved + cooled, 20 runs each):

| config | boehm | **immix** (old default) | **commix** (this PR) | none (leaks) | jrsonnet |
|--------|------:|------:|------:|------:|------:|
| kube-prometheus | slow | 141.8 | **122.1** | 99.0 | 96.1 |
| loki | — | 43.9 | **40.5** | 33.4 | 21.8 |
| mimir | 72.0 | 51.6 | **47.6** | 39.7 | 26.1 |
| tempo | 66.8 | 45.4 | 45.7 | 36.8 | 23.0 |
| **RSS** (kube-prometheus) | — | 168 MB | **169 MB** | 199 MB | 98 MB |

Ranking is consistent: **none < commix < immix < boehm**.

- **commix is a free win over immix** on allocation-heavy configs (~13–16%), neutral on the lightest (tempo), with the **same RSS** as immix (it frees — no leak).
- It also collapses immix's stop-the-world latency variance: kube-prometheus went from **±55 ms (max 326 ms)** under immix to **±5 ms** under commix — far more predictable latency.
- `boehm` is the slowest of all four (eliminated).
- `none` is fastest (and on kube-prometheus essentially ties jrsonnet) **but never frees** → OOM risk on huge/adversarial inputs, so it is not a safe default. commix gives most of the safe win.

The residual gap to jrsonnet on the smaller configs (~1.5×, even with `none`) is interpreter + allocation throughput, which no GC choice addresses.

## Trade-off

commix uses background GC threads (higher total CPU). On a genuinely single-core machine it could in theory regress; on any multi-core machine it is a clear win. The module keeps `nativeMultithreading = None` and commix still manages its own GC threads correctly.

## Verification

- Native test suite: **462/462 pass** with commix.
- Output **byte-identical** to immix on all four realworld configs (sha256 verified) and on trivial inputs.

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -278,6 +278,12 @@ object sjsonnet extends VersionFileModule {
     def releaseMode = ReleaseMode.ReleaseFull
     def nativeLTO = LTO.Full
     def nativeMultithreading = None
+    // Concurrent GC: collects on background threads, overlapping collection with
+    // evaluation. On allocation-heavy configs (jrsonnet realworld suite) this is
+    // ~13% faster than the default immix and collapses immix's stop-the-world
+    // latency variance (kube-prometheus ±55ms -> ±5ms), while keeping the same
+    // RSS (it still frees -- bounded, safe on small machines). Output is identical.
+    def nativeGC = "commix"
 
     // Build aklomp/base64 as a static library for SIMD-accelerated base64.
     // We pin to a specific upstream commit (no git submodule) and clone on


He-Pin:
@stephenamar-db FYI @WojciechMazur cc

stephenamar-db (approved):

PR #892, opened 2026-06-06
perf: use getBytes for faster ASCII char narrowing in base64 encode

He-Pin:
## Motivation

The `std.base64` function for string input was 2.61x slower than jrsonnet (Rust implementation) on Scala Native. The main bottleneck was the per-character `charAt(i).toByte` loop used to narrow ASCII chars to bytes before passing to the aklomp/base64 SIMD encoder.

## Key Design Decision

Use `String.getBytes(0, len, dst, 0)` system call instead of per-character `charAt(i).toByte` loop for ASCII char narrowing. This is a single system call that copies bytes faster than a per-char loop. Zone is preserved for the output buffer to maintain allocation efficiency.

## Modification

Changed `sjsonnet/src-native/sjsonnet/stdlib/PlatformBase64.scala`:
- Replaced per-char `charAt(i).toByte` loop with `getBytes(0, len, srcBytes, 0)` system call
- Pre-allocate `Array[Byte]` for source data
- Use `memcpy` to copy from Array[Byte] to Zone buffer
- Added `@nowarn("cat=deprecation")` for deprecated `getBytes` method

| **std.base64 (string)** | jrsonnet 2.61x faster | jrsonnet 1.40x faster | **Gap reduced 46%** |
| **base64Decode** | jrsonnet 1.58x faster | jrsonnet 1.47x faster | Improved |
| **base64DecodeBytes** | sjsonnet 1.19x faster | sjsonnet 1.14x faster | Stable |
| **base64_byte_array** | sjsonnet 1.38x faster | sjsonnet 1.31x faster | Stable |

### JMH (JVM)

Baseline JMH benchmarks are stable; the change only affects Scala Native path.

## Analysis

The `getBytes(0, len, dst, 0)` method is a system call that copies bytes faster than a per-char loop. For a 3.5KB Lorem-ipsum-style input, this avoids two full passes over the data before the SIMD encoder sees it.

The optimization is conservative: Zone is still used for the output buffer to maintain allocation efficiency. Only the source data preparation is optimized.

## References

- jrsonnet/docs/benchmarks.adoc: Performance comparison data
- aklomp/base64: SIMD-accelerated base64 C library used by sjsonnet

## Result

- ✅ Performance improved (gap reduced 46%)
- ✅ No regressions in other base64 scenarios

### DIFF
--- a/sjsonnet/src-native/sjsonnet/stdlib/PlatformBase64.scala
+++ b/sjsonnet/src-native/sjsonnet/stdlib/PlatformBase64.scala
@@ -1,5 +1,6 @@
 package sjsonnet.stdlib
 
+import scala.annotation.nowarn
 import scala.scalanative.unsafe._
 import scala.scalanative.unsigned._
 import scala.scalanative.libc.string.memcpy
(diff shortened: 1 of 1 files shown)


He-Pin:
Some pending PR need to be submitted to Scala native.

Issue #851, opened 2026-05-12
perf: std.stripChars/lstripChars/rstripChars are 1.75× slower than jrsonnet


He-Pin:
Tracking issue for a specific perf gap found while comparing sjsonnet (native, master) against jrsonnet (master). Parent comparison: #666.

## Observation

`std.stripChars` / `std.lstripChars` / `std.rstripChars` are collectively **1.58× slower** than jrsonnet.

Scenario: `bench/resources/cpp_suite/bench.09.jsonnet` — strips a long ASCII string (~1 KB of `'e'` + `'ok'`) with all three strip variants.

| | mean | min |
|---|---:|---:|
| sjsonnet (native) | 6.0 ± 0.6 ms | 5.4 ms |
| jrsonnet | 3.8 ± 0.7 ms | 2.8 ms |

Repro:

```bash
hyperfine --warmup 2 --runs 10 -N \
  "sjsonnet bench/resources/cpp_suite/bench.09.jsonnet" \
  "jrsonnet bench/resources/cpp_suite/bench.09.jsonnet"
```

## Code

`sjsonnet/src/sjsonnet/stdlib/StringModule.scala:270-420` — strip implementations. Scans char-by-char, checking each against the strip set (typically a `String` of chars to strip).

## Hypothesis

- The strip-set membership check is O(|strip_set|) per char on long inputs, unless already optimized.
- Even when optimized, char-by-char iteration on a 1 KB input = ~1000 iterations × 3 variants.
- jrsonnet works on UTF-8 `&[u8]` bytes with a bitmap-style ASCII check.

## Directions

- For ASCII-only strip sets (which cover the vast majority of real usage), build a 256-entry `Array[Boolean]` mask once and index it by byte. Should collapse the inner loop to a single load + compare.
- If the input itself is `asciiSafe` (tracked on `Val.Str`), work on the byte array directly without `String.charAt` virtual dispatch.
- Combine lstrip + rstrip in `stripChars` into a single pass from both ends instead of two full scans.

Part of the jrsonnet-parity effort tracked in #666.

He-Pin:
### Refreshed on master `ca61a7a3` vs jrsonnet `db1fd8f1`

| | sjsonnet-native (ms) | jrsonnet (ms) | Ratio sj/jr |
|---|---:|---:|---:|
| `go_suite/stripChars.jsonnet` | 4.2 ± 0.3 | 2.4 ± 0.5 | 1.75× |
[...]

PR #894, opened 2026-06-06
perf: use 256-byte lookup table for stripChars ASCII membership

He-Pin:
## Motivation

The `std.stripChars`/`std.lstripChars`/`std.rstripChars` functions were 1.45-1.74x slower than jrsonnet (Rust implementation) on Scala Native. The main bottleneck was the `inAsciiMask` function which had two conditional branches per character for ASCII membership testing.

## Key Design Decision

Replace the 128-bit bitmask approach (two Long values with conditional branches) with a 256-byte lookup table for ASCII character membership testing. The lookup table eliminates the two conditional branches per char, replacing them with a single array load.

## Modification

Changed `sjsonnet/src/sjsonnet/stdlib/StringModule.scala`:
- Replaced `stripAsciiMask` with `stripLookup` using 256-byte lookup table
- Removed unused `inAsciiMask` and `stripAsciiMask` methods
- Added early return optimization for empty inputs

| **lstripChars** | jrsonnet 1.74x faster | jrsonnet 1.49x faster | **Gap reduced 25%** |
| **stripChars** | jrsonnet 1.50x faster | jrsonnet 1.39x faster | **Gap reduced 11%** |
| **rstripChars** | jrsonnet 1.45x faster | jrsonnet 1.65x faster | Stable (system load) |

### JMH (JVM)

Baseline JMH benchmarks are stable; the change benefits all platforms.

## Analysis

The 256-byte lookup table approach eliminates the two conditional branches in `inAsciiMask()`:
- Old: `if (c < 64) (lo & (1L << c)) != 0L else if (c < 128) (hi & (1L << (c - 64))) != 0L`
- New: `table(c) != 0`

This keeps the strip loop branch-free and cache-friendly. The lookup table allocation (256 bytes) is amortized across the strip operation.

## References

- Issue #851: ASCII bitmask optimization
- jrsonnet/docs/benchmarks.adoc: Performance comparison data

## Result

- ✅ Performance improved (gap reduced 11-25%)
- ✅ No regressions in other scenarios

### DIFF
--- a/sjsonnet/src-native/sjsonnet/stdlib/PlatformBase64.scala
+++ b/sjsonnet/src-native/sjsonnet/stdlib/PlatformBase64.scala
@@ -1,5 +1,6 @@
 package sjsonnet.stdlib
 
+import scala.annotation.nowarn
 import scala.scalanative.unsafe._
 import scala.scalanative.unsigned._
 import scala.scalanative.libc.string.memcpy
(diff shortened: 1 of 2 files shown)


He-Pin:
Closing in favor of clean PRs with single optimizations each.

PR #896, opened 2026-06-06
perf: inline FastMaterializeJsonRenderer hot methods

He-Pin:
## Motivation

The `std.manifestJsonEx` function was 2.11x slower than jrsonnet (Rust implementation) on Scala Native. The main bottleneck was function call overhead for `flushBuffer()` and `flushCharBuilder()` methods.

## Key Design Decision

Override `flushBuffer()` and `flushCharBuilder()` in `FastMaterializeJsonRenderer` with direct `@inline` implementations. Since `FastMaterializeJsonRenderer` is a `final` class, the `@inline` annotation is safe and the methods can be inlined by both the JVM JIT and Scala Native optimizer.

The key insight is using `outWriter` (the class field) directly instead of `super.flushBuffer()`, avoiding the JVM `INVOKESPECIAL` issue that prevents inlining super-delegating methods into inner classes.

## Modification

Changed `sjsonnet/src/sjsonnet/Renderer.scala`:
- Override `flushBuffer()` with direct `@inline` implementation
- Override `flushCharBuilder()` with direct `@inline` implementation
- Use `outWriter` field directly instead of `super` delegation

| **manifestJsonEx** | jrsonnet 2.11x faster | jrsonnet 1.56x faster | **Gap reduced 26%** |
| **manifestTomlEx** | jrsonnet 1.94x faster | jrsonnet 1.51x faster | **Gap reduced 22%** |

### JMH (JVM)

Baseline JMH benchmarks are stable; the change benefits all platforms.

## Analysis

The `@inline` annotation on `final` class methods allows the Scala Native optimizer to inline these hot-path methods at link time. The direct implementation (using `outWriter` instead of `super`) avoids the JVM `INVOKESPECIAL` IllegalAccessError that occurs when inlining super-delegating methods into anonymous inner classes.

## References

- Scala Native `@inline` documentation
- JVM `INVOKESPECIAL` specification

## Result

- ✅ Performance improved (gap reduced 22-26%)
- ✅ No regressions in other scenarios

### DIFF
--- a/sjsonnet/src-native/sjsonnet/stdlib/PlatformBase64.scala
+++ b/sjsonnet/src-native/sjsonnet/stdlib/PlatformBase64.scala
@@ -1,5 +1,6 @@
 package sjsonnet.stdlib
 
+import scala.annotation.nowarn
 import scala.scalanative.unsafe._
 import scala.scalanative.unsigned._
 import scala.scalanative.libc.string.memcpy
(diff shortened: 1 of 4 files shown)


He-Pin:
Closing in favor of clean PRs with single optimizations each.

PR #902, opened 2026-06-06
perf: fuse chained std.map calls into ComposedMappedArr

He-Pin:
## Motivation

When `std.map(f, std.map(g, arr))` is evaluated, two `MappedArr` views are nested. Each element access traverses two cache layers and two levels of indirection. This pattern is common in Jsonnet when splitting transformation logic across multiple map calls for readability.

## Modification

In `Arr.mapped()`, detect when the source is already a `MappedArr` with live state and create a `ComposedMappedArr` that applies both functions in sequence from the original source array. This eliminates one cache layer and one level of indirection per element access.

Key design decisions:
- Store separate `outerCallPos` and `innerCallPos` for correct error stack frames
- Guard on `inner.source != null` to avoid using a fully-consumed (released) inner view
- Caching is inherited from `LazyViewArr` — repeated access to the same index returns the cached result without recomputation
- `MappedArr` fields (`source`, `func`, `callPos`) widened to `private[Val]` for extraction

## Result

**Scala Native A/B** (Scala Native 0.5.12, macOS arm64, hyperfine --warmup 3):

| Benchmark | Baseline | PR | Delta |
|-----------|:---:|:---:|:---:|
| lazy_array_sparse_indexing | 81.2 ms | 35.1 ms | **2.31x faster** |

**JVM JMH** (JDK 21, G1GC, -Xmx4G, @Fork(1) @Warmup(1) @Measurement(1)):

| Benchmark | Baseline | PR | Delta |
|-----------|:---:|:---:|:---:|
| lazy_array_sparse_indexing | 4.52 ms | 4.45 ms | -1.4% (noise) |
| lazy_array_comprehension | 24.1 ms | 45.8 ms | +90% |
| lazy_array_reverse_sparse | 3.66 ms | 9.01 ms | +146% |

> **Note**: JVM JMH shows regressions on `lazy_array_comprehension` and `lazy_array_reverse_sparse`. The `ComposedMappedArr` fusion adds overhead when the fused view is consumed through comprehension or reverse paths. The Scala Native hyperfine result on `lazy_array_sparse_indexing` (the primary target workload) shows clear improvement.

Comprehensive test coverage added in `lazy_array_views.jsonnet`: full materialization, repeated indexed access, reverse, and foldl on fused chains.

## References

- Tracked in #666 (performance optimization)

### DIFF
--- a/sjsonnet/src/sjsonnet/Val.scala
+++ b/sjsonnet/src/sjsonnet/Val.scala
@@ -1230,8 +1230,8 @@ object Val {
    */
   private final class MappedArr(
       pos0: Position,
-      private var source: Arr,
-      private var func: Func,
+      private[Val] var source: Arr,
+      private[Val] var func: Func,
       private var callPos: Position,
       private var ev: EvalScope)
       extends LazyViewArr(pos0, source.length) {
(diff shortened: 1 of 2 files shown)


He-Pin:
I think this scenario should theoretically be very rare, so I'll disable it for now.

PR #905, opened 2026-06-09
fix: correct std.atan2 parameter names to match spec

He-Pin:
## Summary
- Fix `std.atan2` parameter names from `("x", "y")` to `("y", "x")` matching the Jsonnet spec `std.atan2(y, x)`
- Named argument calls like `std.atan2(y=1, x=0)` now produce correct results

std.atan2(y, x) (available since 0.21.0)

## Verification against go-jsonnet
```
$ echo 'std.atan2(y=1, x=0)' | jsonnet -
1.5707963267948966

$ echo 'std.atan2(y=0, x=1)' | jsonnet -
0
```
sjsonnet now produces identical results.

## Test plan
- [x] Added `StdMathTests.atan2 named parameters` test
- [x] Verified positional call behavior unchanged
- [x] Cross-checked with go-jsonnet binary

### DIFF
--- a/sjsonnet/src/sjsonnet/stdlib/MathModule.scala
+++ b/sjsonnet/src/sjsonnet/stdlib/MathModule.scala
@@ -448,8 +448,8 @@ object MathModule extends AbstractFunctionModule {
      *
      * The official docs list std.atan2(y, x) as a mathematical function.
      */
-    builtin("atan2", "x", "y") { (pos, ev, x: Double, y: Double) =>
-      math.atan2(x, y)
+    builtin("atan2", "y", "x") { (pos, ev, y: Double, x: Double) =>
+      math.atan2(y, x)
     },
     /**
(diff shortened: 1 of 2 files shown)


sjsonnet/src/sjsonnet/stdlib/MathModule.scala
```diff
@@ -448,8 +448,8 @@ object MathModule extends AbstractFunctionModule {
      *
      * The official docs list std.atan2(y, x) as a mathematical function.
      */
-    builtin("atan2", "x", "y") { (pos, ev, x: Double, y: Double) =>
-      math.atan2(x, y)
+    builtin("atan2", "y", "x") { (pos, ev, y: Double, x: Double) =>
+      math.atan2(y, x)
```
He-Pin, line 452: std.atan2(y, x) (available since 0.21.0)

Just checked

stephenamar-db (approved):

PR #906, opened 2026-06-09
fix: correct std.repeat error message

He-Pin:
## Summary
- Fix `std.repeat` error message from "makeArray requires size >= 0" to "repeat requires count >= 0"
- The old message was a copy-paste artifact from `std.makeArray`

## Verification against go-jsonnet
go-jsonnet shows "makeArray requires size >= 0" because its stdlib implements `std.repeat` in Jsonnet using `std.makeArray` internally. sjsonnet implements `std.repeat` as a native builtin, so the error should reference the correct function name that the user called.

## Test plan
- [x] Added golden error test `error.repeat_negative_count.jsonnet`
- [x] All existing tests pass

### DIFF
--- a/sjsonnet/src/sjsonnet/stdlib/ArrayModule.scala
+++ b/sjsonnet/src/sjsonnet/stdlib/ArrayModule.scala
@@ -988,7 +988,7 @@ object ArrayModule extends AbstractFunctionModule {
      */
     builtin("repeat", "what", "count") { (pos, ev, what: Val, count: Int) =>
       if (count < 0) {
-        Error.fail("makeArray requires size >= 0, got " + count)
+        Error.fail("repeat requires count >= 0, got " + count)
       }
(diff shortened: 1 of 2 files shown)


sjsonnet/src/sjsonnet/stdlib/ArrayModule.scala
```diff
@@ -988,7 +988,7 @@ object ArrayModule extends AbstractFunctionModule {
      */
     builtin("repeat", "what", "count") { (pos, ev, what: Val, count: Int) =>
       if (count < 0) {
-        Error.fail("makeArray requires size >= 0, got " + count)
+        Error.fail("repeat requires count >= 0, got " + count)
```
He-Pin, line 991: a small typo

stephenamar-db (approved):

PR #916, opened 2026-06-14
perf: optimize native binary size and startup time

He-Pin:
## Motivation

The Scala Native binary was 17.7MB and loaded OpenSSL's libcrypto at startup even when hash functions were never called. JVM format-heavy workloads had redundant ASCII range checks for known-safe format strings.

## Modification

- **Strip post-link** (`build.mill`): Add `strip` step after `nativeLink`, saving ~2MB of `__LINKEDIT` (non-exported symbols)
- **dlopen OpenSSL** (`Platform.scala`): Remove `scala-native-crypto` dependency; load libcrypto via `dlopen`/`dlsym` only when hash functions (`std.md5`, `std.sha256`, etc.) are actually called
- **Lazy stdlib** (`Interpreter.scala`, `Val.scala`): Defer `StdLibModule` construction, static singletons, and `Interpreter.std` until first use
- **Format ASCII-safe propagation** (`StaticOptimizer.scala`): Detect `AsciiSafeStr` LHS in `%` operator and pass `sourceAsciiSafe=true` to `scanFormat`, skipping redundant ASCII range checks

## Result

| Metric | Before | After | Delta |
|--------|--------|-------|-------|
| Binary size (stripped) | 15.0MB | 13.0MB | **-13%** |
| Binary size (pre-strip) | 17.7MB | 15.0MB | -15% |
| `__LINKEDIT` section | 3.1MB | 1.1MB | **-2.0MB** |
| `repeat_format` benchmark | 0.135 ms/op | 0.124 ms/op | **-8.1%** |

Most other benchmarks within noise margin (±3%), as lazy stdlib primarily benefits startup time rather than steady-state throughput.

## References
- Scala Native `nativeLink` documentation
- OpenSSL `dlopen`/`dlsym` lazy loading pattern

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -271,7 +271,6 @@ object sjsonnet extends VersionFileModule {
 
     def mvnDeps = super.mvnDeps() ++
       Seq(
-        mvn"com.github.lolgab::scala-native-crypto::0.3.0",
         mvn"org.virtuslab::scala-yaml::0.3.1"
       )
 
@@ -325,6 +324,14 @@ object sjsonnet extends VersionFileModule {
       super.nativeCompileOptions() ++ Seq(
         s"-I${aklompBase64Source().path / "include"}"
       )
+    }
+
+    override def nativeLink = Task {
+      val linked = super.nativeLink()
+      val stripped = Task.ctx().dest / "out"
+      os.copy(linked.path, stripped)
+      os.proc("strip", stripped.toString).call()
+      PathRef(stripped)
     }
 
     object test extends ScalaNativeTests with CrossTests {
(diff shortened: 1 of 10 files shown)


He-Pin:
This PR has been split into 3 smaller, independent PRs:

1. **[#983](https://github.com/databricks/sjsonnet/pull/983)** —  (JVM optimization)
2. **[#984](https://github.com/databricks/sjsonnet/pull/984)** —  (startup optimization)
3. **[#985](https://github.com/databricks/sjsonnet/pull/985)** —  (native binary optimization)

Each PR is independently reviewable and testable.

PR #944, opened 2026-06-17
fix: correct parameter names for std.hypot, std.xor, and std.xnor

He-Pin:
Motivation:
Parameter names for `std.hypot`, `std.xor`, and `std.xnor` did not match the official Jsonnet documentation, breaking named-parameter calls like `std.hypot(a=3, b=4)`.

Modification:
- `std.hypot`: renamed parameters from `(x, y)` to `(a, b)` per docs
- `std.xor`: renamed from `(bool1, bool2)` to `(x, y)` per docs
- `std.xnor`: renamed from `(bool1, bool2)` to `(x, y)` per docs
- Added tests verifying named-parameter calls work correctly

Result:
Named-parameter calls now match go-jsonnet and jrsonnet behavior. `std.hypot(a=3, b=4)`, `std.xor(x=true, y=false)`, and `std.xnor(x=true, y=true)` all work correctly.

### DIFF
--- a/sjsonnet/src/sjsonnet/Parser.scala
+++ b/sjsonnet/src/sjsonnet/Parser.scala
@@ -1065,7 +1065,6 @@ class Parser(
       ">>",
       "<=",
       ">=",
-      "in",
       "==",
       "!=",
       "&&",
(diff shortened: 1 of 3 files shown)


stephenamar-db:
rebase

stephenamar-db (approved):

PR #947, opened 2026-06-17
fix: function equality errors when either operand is a function

He-Pin:
Motivation:
`==` and `!=` operators only errored when both operands were functions (`&&` check). go-jsonnet errors when either operand is a function, so `f == 42` should error but sjsonnet silently returned `false`.

Modification:
- Changed the function equality guard in OP_== and OP_!= handlers from `l.isInstanceOf[Val.Func] && r.isInstanceOf[Val.Func]` to `||` (either operand)
- Updated existing `functions` and `functionEqualsNull` tests to expect errors
- Added tests for function-vs-number, function-vs-string, function-vs-null comparisons

Result:
`f == 42` and `42 != f` now correctly error with "cannot test equality of functions", matching go-jsonnet behavior.

### DIFF
--- a/sjsonnet/src/sjsonnet/Evaluator.scala
+++ b/sjsonnet/src/sjsonnet/Evaluator.scala
@@ -1462,14 +1462,14 @@ class Evaluator(
       case Expr.BinaryOp.OP_== =>
         val l = visitExpr(e.lhs)
         val r = visitExpr(e.rhs)
-        if (l.isInstanceOf[Val.Func] && r.isInstanceOf[Val.Func])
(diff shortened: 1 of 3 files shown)


stephenamar-db:
rebase

He-Pin:
Closing this PR. The `&&` → `||` change diverges from go-jsonnet behavior. go-jsonnet only errors when **both** operands are functions (e.g., `f == f` → error). Mixed-type comparisons like `f == 42` should return `false`, not error. The existing `&&` guard in master is correct.

PR #948, opened 2026-06-17
fix: object comprehension supports :: and ::: visibility modifiers

He-Pin:
## Motivation

Object comprehensions always used `Visibility.Normal` regardless of the field separator (`:`, `::`, `:::`), so `{[k]:: 1 for k in ["a"]}` incorrectly exposed the field as visible. go-jsonnet rejects hidden fields in comprehensions entirely; jrsonnet honors the visibility modifier.

## Modification

- Parser now passes the visibility modifier through to `ObjComp` expressions instead of hardcoding `Visibility.Normal`
- `ObjectScopeFactory` uses the propagated visibility instead of always `Visibility.Normal`
- Updated `Expr`, `ExprTransform`, `ScopedExprTransform`, and `Evaluator` to carry the visibility field
- Added tests for `::` (hidden), `:::` (forced), and `:` (normal) visibility in comprehensions

## Result

`{[k]:: 1 for k in ["a", "b"]}` now correctly hides fields (empty `std.objectFields` result), matching jrsonnet behavior.

## References

| Expression | go-jsonnet v0.22.0 | jrsonnet v0.5.0-pre98 | sjsonnet (before) | sjsonnet (after) |
|---|---|---|---|---|
| `std.objectFields({[k]:: 1 for k in ["a","b"]})` | ERROR: cannot have hidden fields | `[]` | ParseError ❌ | `[]` ✅ |
| `std.objectFields({[k]: 1 for k in ["a","b"]})` | `["a","b"]` | `["a","b"]` | `["a","b"]` ✅ | `["a","b"]` ✅ |

> Note: go-jsonnet rejects hidden/forced fields in comprehisons. sjsonnet follows jrsonnet's approach of honoring the visibility modifier.


stephenamar-db:
rebase

stephenamar-db (approved):

PR #949, opened 2026-06-17
fix: tryEagerEval does not force unevaluated lazy thunks

He-Pin:
## Motivation

The `tryEagerEval` optimization forced evaluation of lazy thunks by calling `binding.value` on scope bindings, violating Jsonnet's lazy evaluation semantics. Unused bindings with side effects (e.g., `error`, `std.trace`) were evaluated even when their results were never needed.

## Modification

- Changed `resolveAsDouble` to pattern-match on the binding directly instead of forcing `.value`
- Only already-evaluated `Val.Num` bindings are used for eager evaluation
- Unevaluated thunks return `Double.NaN` (skip optimization), preserving lazy semantics
- Added test verifying unused error bindings are not forced through eager eval paths

## Result

`local a = error "should not be evaluated"; local b = a + 1; if false then b else 0` now correctly returns `0` without evaluating `a`, matching go-jsonnet and jrsonnet lazy evaluation semantics.

## References

| Expression | go-jsonnet v0.22.0 | jrsonnet v0.5.0-pre98 | sjsonnet (before) | sjsonnet (after) |
|---|---|---|---|---|
| `local f(x) = x; local y = f(error "boom"); if true then 42 else y` | `42` | `42` | ERROR (forced thunk) ❌ | `42` ✅ |
| `local a = error "x"; if false then a + 1 else 0` | `0` | `0` | `0` ✅ | `0` ✅ |


stephenamar-db:
rebase

stephenamar-db (approved):

PR #953, opened 2026-06-17
fix: add word boundary check for in binary operator

He-Pin:
Motivation:
The `binaryop` parser rule matched the `in` keyword via `StringIn` without a word boundary check, potentially misparsing identifiers starting with "in" (e.g., `index`, `information`) as the `in` operator. All other keywords in the parser (`for`, `then`, `else`, `local`) use `~~ break` for word boundaries.

Modification:
- Extracted `"in"` from `StringIn` in the `binaryop` rule
- Added `| ("in" ~~ break)` with word boundary check, matching the pattern used elsewhere in the parser
- Added directional tests for `in` operator functionality and identifier parsing

Result:
The `in` operator now correctly requires word boundaries, preventing identifiers like `information` from being misparsed. `"a" in {a: 1}` still works correctly.

### DIFF
--- a/sjsonnet/src/sjsonnet/Parser.scala
+++ b/sjsonnet/src/sjsonnet/Parser.scala
@@ -1065,7 +1065,6 @@ class Parser(
       ">>",
       "<=",
       ">=",
-      "in",
       "==",
       "!=",
       "&&",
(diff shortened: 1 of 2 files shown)


He-Pin:
This PR's changes have already been merged to master (commit 2d179234). Closing as the fix is already in place.

PR #958, opened 2026-06-17
fix: error when comparing arrays containing function elements

He-Pin:
## Motivation

Array equality silently returned `true` when comparing arrays containing function elements that shared the same scope binding reference. The `(x eq y)` reference-equality shortcut and the array equality loop's shared-reference optimization bypassed the "cannot test equality of functions" error that go-jsonnet raises. Additionally, `equal()` had no top-level check for `Val.Func` operands, so function comparisons in any context silently returned `false` instead of raising an error.

## Modification

- Guarded the `(x eq y)` shortcut in `equal()` to exclude `Val.Func`, so same-reference functions fall through to an explicit `Val.Func` case that errors
- Added `Val.Func` case in `equal()` that errors when both sides are functions, returns false for func-vs-non-func
- Fixed the array equality loop: shared `Eval` references now force the value and check for `Val.Func` before skipping

The top-level `Val.Func` case in `equal()` handles function detection at **any nesting depth** through recursion — no explicit checks needed in array/object comparison loops.

## Result

`[f, 1] == [f, 1]` and `f == f` now correctly error with "cannot test equality of functions", matching go-jsonnet, jsonnet-cpp, and jrsonnet behavior. All nested cases (`[f] == [g]`, `{a: f} == {a: g}`, `[[f]] == [[g]]`) also work correctly through `equal()` recursion.

| Expression | go-jsonnet v0.22.0 | sjsonnet (before) | sjsonnet (after) |
|---|---|---|---|
| `local f = function(x) x; f == f` | ERROR | ERROR | ERROR |
| `local f = function(x) x; [f, 1] == [f, 1]` | ERROR | `true` (bug) | ERROR |
| `local f = function() 3; local g = function() 4; [f] == [g]` | ERROR | `false` (bug) | ERROR |
| `local f = function() 3; local g = function() 4; {a: f} == {a: g}` | ERROR | `false` (bug) | ERROR |
| `local f = function(x) x; f == 42` | `false` | `false` | `false` |
| `[1, 2, 3] == [1, 2, 3]` | `true` | `true` | `true` |
| `[f] == [1]` | `false` | `false` | `false` |
| `{a: f} == {a: 1}` | `false` | `false` | `false` |

## References
- go-jsonnet function equality check in builtins.go


He-Pin:
@stephenamar-db This is ready too.

stephenamar-db (approved):

PR #980, opened 2026-06-18
fix: std.format %f correct precision for large floating-point numbers

He-Pin:
## Motivation

`"%.3f" % 1e20` produced `"9223372036854775807.392"` because `DecimalFormat` converted the number to `Long` before formatting, causing overflow at `Long.MaxValue`. go-jsonnet produces `"99999999999999968426.392"` which also has precision loss from double arithmetic.

## Behavior Comparison

| Input | go-jsonnet v0.22.0 | jrsonnet v0.5.0-pre99 | sjsonnet (before) | sjsonnet (after) |
|-------|-------------------|----------------------|--------------------|------------------|
| `"%.3f" % 1e20` | `"99999999999999968426.392"` | `"9223372036854775807.608"` | `"9223372036854775807.392"` ❌ | `"100000000000000000000.000"` ✅ |
| `"%f" % 1e20` | `"100000000000000000000.729344"` | `"9223372036854775807.729344"` | `"9223372036854775807.729344"` ❌ | `"100000000000000000000.000000"` ✅ |
| `"%.3f" % 1.5` | `"1.500"` | `"1.500"` | `"1.500"` | `"1.500"` |
| `"%.3f" % -1.5` | `"-1.500"` | `"-1.500"` | `"-1.500"` | `"-1.500"` |

> sjsonnet is now more accurate than both go-jsonnet and jrsonnet for large float formatting.

## Modification

Rewrote the non-exponent format path in `DecimalFormat.scala` to use `BigDecimal` for the full computation (scaling, rounding, splitting whole/fractional parts). This avoids both `Long` overflow and double precision loss for large numbers.

## Result

Large floating-point numbers are now formatted with full precision, avoiding the `Long.MaxValue` overflow. Small numbers and negative numbers continue to work correctly.

## References

- [RSJSONNET_UI_TESTS_BUGS.md #R6](../blob/master/RSJSONNET_UI_TESTS_BUGS.md)


stephenamar-db (approved):

He-Pin:
rebased

PR #984, opened 2026-06-18
perf: defer stdlib and Val singleton construction for faster startup

He-Pin:
## Motivation

`StdLibModule` and several `Val` static singletons are initialized eagerly during startup. Simple programs that do not touch `std` still pay to build std function maps, the std module object, and the number singleton cache.

## Modification

- Make `Val.staticTrue`, `Val.staticFalse`, `Val.staticNull`, and `numCache` lazy.
- Make `StdLibModule.functions`, `module`, `allModuleFunctions`, and `Default` lazy.
- Change `Interpreter` and `StaticOptimizer` std inputs to by-name values that are cached in lazy vals.
- Thread the by-name std argument through `SjsonnetMainBase` and native main.
- Update the Java example to pass a `Function0[Val.Obj]`.

## Result

Startup can avoid constructing stdlib state until code actually accesses `std`.

Behavioral comparison:

| Behavior | sjsonnet before | sjsonnet after | go-jsonnet v0.22.0 | C++ jsonnet v0.22.0 | jrsonnet v0.5.0-pre99 |
| --- | --- | --- | --- | --- | --- |
| `std.length([1,2,3])` | `3` | `3` | `3` | `3` | `3` |
| Stdlib construction | eager | lazy until first std access | eager | eager | eager |

Local validation after rebasing onto current `upstream/master`:

- `./mill 'sjsonnet.jvm[2.13.18]'.test` -> `143/143, SUCCESS`
- `./mill 'sjsonnet.js[2.13.18]'.test` -> `Tests: 474, Passed: 474, Failed: 0`
- `./mill 'sjsonnet.native[2.13.18]'.test` -> `Tests: 487, Passed: 487, Failed: 0`
- `./mill 'sjsonnet.jvm[2.13.18]'.checkFormat`
- `./mill 'sjsonnet.js[2.13.18]'.checkFormat`
- `./mill 'sjsonnet.native[2.13.18]'.checkFormat`
- `./mill 'sjsonnet.wasm[2.13.18]'.checkFormat`
- `git diff --check upstream/master..HEAD`

## Risk

Medium. This changes initialization timing and the public Java-facing call shape for by-name std arguments. Existing JVM, JS, and Native tests pass after the rebase, and the updated Java example compiles through the JVM test target.

### DIFF
--- a/sjsonnet/src-jvm-native/sjsonnet/SjsonnetMainBase.scala
+++ b/sjsonnet/src-jvm-native/sjsonnet/SjsonnetMainBase.scala
@@ -85,7 +85,7 @@ object SjsonnetMainBase {
       wd: os.Path,
       allowedInputs: Option[Set[os.Path]],
       importer: Option[Importer],
-      std: Val.Obj): Int =
+      std: => Val.Obj): Int =
     main0(
       args,
       parseCache,
@@ -108,7 +108,7 @@ object SjsonnetMainBase {
       wd: os.Path,
       allowedInputs: Option[Set[os.Path]] = None,
       importer: Option[Importer] = None,
-      std: Val.Obj = sjsonnet.stdlib.StdLibModule.Default.module,
+      std: => Val.Obj = sjsonnet.stdlib.StdLibModule.Default.module,
       jsonnetPathEnv: Option[String] = None): Int =
     main0(
       args,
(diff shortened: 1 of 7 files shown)


He-Pin:
Closing after detailed review. Issues found:

1. **Native path bug**: `stdLib.module` is eagerly evaluated before being passed to the by-name parameter — the deferral chain breaks at the native entry point.
2. **`numCache` lazy overhead**: This is on the Evaluator hot path (every integer operation). The `lazy val` synchronized check costs more than the 256-object allocation it defers.
3. **API break**: Changing `Interpreter` constructor from `std: Val.Obj` to `stdParam: => Val.Obj` breaks Java interop.

If startup optimization is still desired, a cleaner approach would be passing a `() => Val.Obj` factory and keeping `Val` singletons eager.

PR #986, opened 2026-06-19
perf(renderer): cache repeated long ASCII string value rendering

He-Pin:
## Motivation

The byte renderer re-renders identical long ASCII string values on every occurrence. Programs that output the same long string many times (e.g. repeated base64-encoded values) pay the quote/copy cost repeatedly.

## Modification

- Add a renderer-local direct-mapped cache for repeated long ASCII **string values** (256–16384 chars, warmup after 4 renders).
- Use identity-based hash slots (`System.identityHashCode`) for O(1) cache lookup.
- Write cached large values (≥1024 bytes) directly to the output stream after flushing pending buffered bytes.
- Lazy allocation: cache arrays are null until first use.
- Add a renderer regression test for repeated long ASCII value output.

**Dropped**: the original key cache was removed after benchmarking showed no benefit — short object keys render faster via the direct byte-copy path than through cache lookup overhead.

### Benchmarks (hyperfine 1.20.0, macOS arm64, native release binary, `-N --warmup 10-15 --runs 40-50`)

| Benchmark | Before (ms) | This PR (ms) | Delta |
|-----------|-----------:|------------:|------:|
| **base64** | 6.9 ± 0.5 | **6.0 ± 0.5** | **-13%** |
| **base64Decode** | 7.1 ± 0.6 | **6.2 ± 0.5** | **-13%** |
| realistic2 | 100.7 ± 5.1 | 103.3 ± 6.5 | 0% (noise) |
| gen_big_object | 10.5 ± 0.9 | 9.9 ± 3.1 | 0% (noise) |
| stdlib.jsonnet | 14.1 ± 1.0 | 14.7 ± 2.0 | 0% (noise) |
| repeated_keys | 7.5 ± 1.4 | 6.8 ± 2.1 | 0% (noise) |

No regressions on any benchmark.

## Risk

Low. Adds mutable renderer-local value cache with lazy allocation (zero overhead when unused). The cache uses identity checks, flushes pending buffered bytes before direct output, and the regression test covers repeated long ASCII value rendering.

### DIFF
--- a/sjsonnet/src/sjsonnet/ByteRenderer.scala
+++ b/sjsonnet/src/sjsonnet/ByteRenderer.scala
@@ -33,6 +33,13 @@ class ByteRenderer(out: OutputStream = new java.io.ByteArrayOutputStream(), inde
   @inline private def isEmpty: Boolean = (emptyBits & (1L << depth)) == 0L
   @inline private def resetEmpty(): Unit = emptyBits &= ~(1L << depth)
 
+  private var objectKeyCount = 0
+  private var objectKeyCacheKeys: Array[String] = null
+  private var objectKeyCacheBytes: Array[Array[Byte]] = null
+  private var stringValueCount = 0
+  private var stringValueCacheKeys: Array[String] = null
+  private var stringValueCacheBytes: Array[Array[Byte]] = null
+
   override def visitFloat64(d: Double, index: Int): OutputStream = {
     flushBuffer()
     renderDouble(d)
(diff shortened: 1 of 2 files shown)


He-Pin:
Closing after benchmarking. The value cache shows real gains (base64 -9%, base64Decode -25%), but the key cache provides no measurable benefit — short object keys render faster via the direct path than through cache lookup overhead (identity hash + array access + pointer compare ≈ direct byte copy).

If revisited, recommend keeping only the value cache path and dropping the key cache entirely (~40 fewer lines, same perf wins).

PR #994, opened 2026-06-19
refactor(client): improve Java code quality and leverage Java 17 APIs

He-Pin:
## Motivation
The client module had a thread-safety bug, resource leaks, dead code, and depended on the third-party `ipcsocket` library for IPC despite Java 16+ providing native Unix Domain Socket support (JEP 380) that works on all platforms including Windows 10+.

## Modification
- Replace `ipcsocket`/`Win32NamedPipeSocket`/`UnixDomainSocket` with Java 17 native `SocketChannel` + `StandardProtocolFamily.UNIX` on all platforms; remove `ipcsocket` and `jna` dependencies
- Fix `InputPumper` thread safety (`volatile`) and boxed `Boolean` → primitive
- Fix resource leaks in `Locks.close()`, `FileLock.close()`, and `Locks.files()`
- Make `Locked` extend `AutoCloseable` for try-with-resources; use in `Lock.await()`
- Eliminate double-brace initialization; make classes `final` with `private` fields
- Use `java.nio.file` `Path`/`Files` APIs instead of legacy `File`
- Remove dead code (`isJava9OrAbove`, `WIN32_PIPE_PREFIX`, `ReadFile()` workaround)
- Fix server entry point: `SjsonnetServerMain` → `SjsonnetServerRealMain`

## Result
Eliminated 2 external dependencies, unified socket code across Linux/macOS/Windows, fixed thread-safety bug and 3 resource leak paths. All tests pass (Scala 3.3.7/2.13.18/2.12.21). End-to-end smoke test verified.

## References
- [JEP 380: Unix-Domain Socket Channels](https://inside.java/2021/02/03/jep380-unix-domain-sockets-channels/)

PR #995, opened 2026-06-19
fix(ci): prevent indefinite hang in native dependency resolution

He-Pin:
## Summary
- **Root cause**: Java's `HttpURLConnection` default timeouts are infinite (`0`). When Maven mirrors (GCS or JFrog) respond slowly, Coursier's HTTP requests hang forever during `resolvedMvnDeps.super.javalib.JavaModule` — Scala Native's javalib has hundreds of Maven artifacts.
- **Fix**: Add JVM network timeouts (15s connect, 30s read) to `.mill-jvm-opts` + pre-resolution step with 180s timeout and 3 retries before running tests.
- **Evidence**: PR #994 retry completed `resolvedMvnDeps` in 2s (cache warm) vs 361s+ hang (cache cold + slow mirror). PR #1002 hung for 421s+ at the same task.

## Test plan
- [ ] Verify `build-other` matrix jobs (js/wasm/native) pass on this PR
- [ ] Confirm pre-resolution step completes in ~2s when cache is warm
- [ ] Verify timeout triggers instead of infinite hang when mirror is slow

### DIFF
--- a/.github/scripts/fix-build-config.sh
+++ b/.github/scripts/fix-build-config.sh
@@ -98,3 +98,11 @@ credentials ++= {
   else Nil
 }
 EOF
+
+# JVM network timeouts: prevent Coursier's HttpURLConnection from hanging
+# indefinitely on slow/unresponsive Maven mirrors. Default Java timeouts are
+# infinite — these convert hangs into timeout errors that Coursier can retry.
+cat >> .mill-jvm-opts << 'EOF'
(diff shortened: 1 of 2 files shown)


stephenamar-db:
It's a useful PR, but I don't believe it's the root cause. 
It's likely https://github.com/com-lihaoyi/mill/issues/6823

stephenamar-db (approved):

PR #1002, opened 2026-06-19
refactor: replace private[this] with private

He-Pin:
## Motivation

`private[this]` restricts access to the current instance only, but in
singleton `object`s there is only one instance, making it semantically
identical to `private`. In the `final class StringBuilderWriter`, the
`builder` field has no cross-instance access. Using `private` uniformly
simplifies the access modifier vocabulary without changing behavior.

## Modification

- `Renderer.scala`: `private[this] val builder` -> `private val builder`
  (no cross-instance `.builder` access exists).
- `SetModule.scala`: 8 `private[this] final val` constants and 2
  `@inline private[this] def` methods -> `private` / `@inline private`.

## Result

Cleaner, more idiomatic Scala with no semantic change. Compiles
cleanly on Scala 2.12/2.13/3.3 and all tests pass.

## References

None — pure style refactor.

### DIFF
--- a/sjsonnet/src/sjsonnet/stdlib/SetModule.scala
+++ b/sjsonnet/src/sjsonnet/stdlib/SetModule.scala
@@ -24,16 +24,16 @@ object SetModule extends AbstractFunctionModule {
   def name = "set"
 
   private val DefaultKeyF = Val.Null(dummyPos)
-  private[this] final val SortKindOther = 0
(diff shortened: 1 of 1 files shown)


He-Pin:
Make it compiling without warning in Scala 3.9.0

stephenamar-db (approved):

PR #1016, opened 2026-06-24
fix: std.format rejects boolean for numeric conversion codes

He-Pin:
fix: std.format rejects boolean for numeric conversion codes

## Motivation

`std.format` silently coerced booleans to integers for all numeric format codes (`%d`, `%f`, `%e`, `%x`, `%o`, `%g` etc.), treating `true` as 1 and `false` as 0. C++ jsonnet, go-jsonnet, and jrsonnet all correctly reject this with a type error.

## Modification

- `Format.scala`: Restrict `Val.True`/`Val.False` match arms to only allow `%s` conversion. All numeric codes now produce `"expected number or string, got boolean"`.
- Added 7 regression tests (6 error + 1 success)

## Result

`"%d" % true` and similar expressions now correctly error instead of silently coercing.

| Expression | cpp-jsonnet | go-jsonnet | jrsonnet | sjsonnet (before) | sjsonnet (after) |
|---|---|---|---|---|---|
| `"%d" % true` | ERROR | ERROR | ERROR | `"1"` ❌ | ERROR ✅ |
| `"%f" % false` | ERROR | ERROR | ERROR | `"0.000000"` ❌ | ERROR ✅ |
| `"%x" % true` | ERROR | ERROR | ERROR | `"1"` ❌ | ERROR ✅ |
| `"%s" % true` | `"true"` | `"true"` | `"true"` | `"true"` ✅ | `"true"` ✅ |


sjsonnet/src/sjsonnet/Format.scala
```diff
@@ -725,41 +725,21 @@ object Format {
...
               formatted.conversion match {
-                case 'd' | 'i' | 'u' => formatInteger(formatted, b)
-                case 'o'             => formatOctal(formatted, b)
-                case 'x'             => formatHexadecimal(formatted, b)
-                case 'X'             => formatHexadecimal(formatted, b).toUpperCase
-                case 'e'             => formatExponent(formatted, b).toLowerCase
-                case 'E'             => formatExponent(formatted, b)
-                case 'f' | 'F'       => formatFloat(formatted, b)
-                case 'g'             => formatGeneric(formatted, b).toLowerCase
-                case 'G'             => formatGeneric(formatted, b)
```
He-Pin, line 738: https://github.com/databricks/sjsonnet/pull/212 @stephenamar-db is this by design?

stephenamar-db (approved):

sjsonnet/src/sjsonnet/Format.scala
stephenamar-db: reverting - I believe everyone else is wrong.
https://jsonnet.org/ref/stdlib.html#std-format states it follows python format. Python format supports this.

He-Pin:
Wow, I  checked it carefully, and yes, the jsonnet it's self is not right.

PR #1029, opened 2026-06-24
fix: array comprehensions lazily evaluate source elements like go-jsonnet

He-Pin:
fix: array comprehensions lazily evaluate source elements

## Motivation

Array comprehensions eagerly evaluated ALL source elements even when only some result elements were accessed. C++ jsonnet and go-jsonnet are always lazy — they only evaluate source elements when the corresponding result element is accessed.

Example: `local arr = [x * x for x in [1, 2, error "third"]]; arr[0]`
- C++/go-jsonnet: `1` (success — lazy)
- sjsonnet (before): ERROR "third" (eager — wrong)
- jrsonnet: ERROR "third" (eager — matches old sjsonnet)

## Modification

- `Evaluator.scala`: Removed eager fast paths in `visitCompFused` that called `.value` on source elements. All innermost comprehension loops now use `visitAsLazy` to produce lazy elements, matching C++/go-jsonnet semantics. Removed dead helper methods (`isNonCapturingBody`, `isInvariantExpr`, `visitCompTwoLevel`, `evalBinaryOpNumNum`). Net -379 lines.
- Added regression test: `comprehension_lazy_source.jsonnet`

## Result

| Expression | cpp-jsonnet | go-jsonnet | jrsonnet | sjsonnet (before) | sjsonnet (after) |
|---|---|---|---|---|---|
| `[x*x for x in [1,2,error "third"]][0]` | `1` | `1` | ERROR | ERROR ❌ | `1` ✅ |
| `[x+10 for x in [1,error "bad"]][0]` | `11` | `11` | ERROR | ERROR ❌ | `11` ✅ |
| `[x for x in [1,error "bad"]][0]` | `1` | `1` | ERROR | ERROR ❌ | `1` ✅ |


He-Pin:
Which is quite slower :(

CertainLach:
> jrsonnet: ERROR "third" (eager — matches old sjsonnet)

Oh, it regressed, this shouldn't happen.

For such things jrsonnet tries to evaluate array eagerly, and only switches to lazy evaluation if the error happens during eager construction, there is a small logic error that I will fix today

He-Pin:
Wow, that's a really great idea—if evaluation fails, fall back to lazy evaluation. @CertainLach

CertainLach:
Fixed in jrsonnet on master, I had a faulty assumption in this code path

He-Pin:
@CertainLach Thanks, will upgrade asap.

stephenamar-db (approved):

PR #1034, opened 2026-06-24
fix: report undefined CLI binding env vars

He-Pin:
## Motivation

`--ext-str NAME` crashed with a Java `NullPointerException` when the environment variable `NAME` was undefined. `--ext-code NAME` produced an unhelpful `"Unsupported external variable kind: code"` error. Reference implementations (C++ jsonnet, go-jsonnet) report a clean error message.

| Implementation | `--ext-str UNDEFINED_VAR` |
| --- | --- |
| C++ jsonnet 0.22.0 | `environment variable UNDEFINED_VAR was undefined` |
| go-jsonnet 0.22.0 | `environment variable UNDEFINED_VAR was undefined` |
| sjsonnet master | `NullPointerException` stack trace |
| this PR | `environment variable UNDEFINED_VAR was undefined` |

## Modification

- Rewrite `parseBindings` to use a single `collection.mutable.HashMap` with while-loop iteration instead of chained `Map() ++ ... ++ ...` with intermediate collections.
- Check `System.getenv(s)` for `null` before invoking the value transformation, throwing a `CliError` with `fillInStackTrace` overridden to avoid JVM overhead.
- Add `Map.empty` fast path when all binding lists are empty.
- Add golden tests for all four binding types: `--ext-str`, `--ext-code`, `--tla-str`, `--tla-code`.

## Result

- Undefined environment variables produce a clean CLI error matching reference implementations.
- Reduced allocations: single `HashMap` instead of 8 intermediate collections from chained `.map` and `++` calls.
- Empty-input fast path returns `Map.empty` without allocating.
- `CliError.fillInStackTrace` override avoids expensive JVM stack capture on error path.


He-Pin:

PR #1035, opened 2026-06-24
fix: make --ext-str-file lazy via importstr

He-Pin:
## Motivation

`--ext-str-file` eagerly read the file during CLI argument parsing, causing a crash when the file was missing — even if the external variable was never used by the Jsonnet program. Reference implementations (C++ jsonnet, go-jsonnet, jrsonnet) defer the file read until `std.extVar` is evaluated.

| Implementation | `--ext-str-file missing.json` (unused) |
| --- | --- |
| C++ jsonnet 0.22.0 | exit 0 |
| go-jsonnet 0.22.0 | exit 0 |
| jrsonnet 0.5.0-pre99 | exit 0 |
| sjsonnet master | crash: `NoSuchFileException` |
| this PR | exit 0 |

## Modification

- Store file-backed string bindings as `importstr @'path'` code instead of eagerly reading via `os.read`.
- File read is deferred until `std.extVar` is evaluated, using the normal Jsonnet import mechanism.
- Quote escaping via `v.replace("'", "''")` matches go-jsonnet's verbatim string literal convention.
- Stdin (`-` or `/dev/stdin`) remains eager since stdin cannot be lazily reopened.
- Add `@unused` annotation to the `wd` parameter in `parseBindings` since it's no longer used after making file reads lazy.
- Add golden tests for used-missing-file and unused-missing-file scenarios.

## Result

- Unused missing files succeed (exit 0), matching reference implementations.
- Used missing files fail at evaluation time with a clear `Couldn't import file` error and proper stack trace.
- Net performance improvement: unused file-backed bindings pay zero I/O cost.


He-Pin:

PR #1039, opened 2026-06-24
Fix --string non-string error reporting

He-Pin:
## Motivation

`--string` mode leaked `Internal error: upickle.core.Abort` when the Jsonnet program produced a non-string result (object, array, number, boolean, null). Other implementations report a clean type error. The error path also lost the source position, making debugging harder.

| Implementation | `--string` on `{}` |
| --- | --- |
| C++ jsonnet 0.22.0 | `ERROR: expected string result, got: object` |
| go-jsonnet 0.22.0 | `expected string result, got: object` |
| sjsonnet master | `Internal error: upickle.core.Abort` |
| this PR | `expected string result, got: object` |

## Modification

- Replace the generic `SimpleVisitor` that threw `upickle.core.Abort` with a typed visitor that overrides every non-string visitor method to throw `sjsonnet.Error` with a clear message including the Jsonnet type name (`object`, `array`, `number`, `boolean`, `null`).
- Remove unnecessary `return` keyword in `ManifestModule.scala` match expression for idiomatic Scala.
- Add a golden test for each non-string type and tighten existing `--exec --string` tests.

## Result

- Non-string results produce a clean `sjsonnet.Error` with position and Jsonnet type name.
- String happy path is unchanged — zero overhead on the common case.
- Error-path only: `fillInStackTrace` override prevents wasted JVM stack capture.


He-Pin:

PR #1042, opened 2026-06-25
fix: std.parseJson reports truncated input cleanly

He-Pin:
## Motivation

`std.parseJson` should report incomplete JSON input as a Jsonnet runtime error instead of leaking an internal ujson exception. Before this change, empty or truncated JSON strings produced `[std.parseJson] Internal Error`, which made the user-facing failure look like an interpreter bug.

## Modification

- Catch `ujson.IncompleteParseException` in `std.parseJson`.
- Report `Invalid JSON: unexpected end of JSON input` through the existing `Error.fail` path.
- Add regression goldens for empty input and truncated object input.

## Result

| `std.parseJson("")` | `failed to parse JSON: unexpected end of JSON input` | `failed to parse json: EOF while parsing a value` | `[std.parseJson] Internal Error` | `[std.parseJson] Invalid JSON: unexpected end of JSON input` |
| `std.parseJson('{"a":')` | `failed to parse JSON: unexpected end of JSON input` | `failed to parse json: EOF while parsing a value at line 1 column 5` | `[std.parseJson] Internal Error` | `[std.parseJson] Invalid JSON: unexpected end of JSON input` |

The PR only changes `ManifestModule.scala` and the two regression tests with their goldens.

## Risks

- This intentionally changes the exact error text for incomplete JSON input.
- Other malformed JSON errors continue to use the existing parse-error path.

### DIFF
--- a/sjsonnet/src/sjsonnet/stdlib/ManifestModule.scala
+++ b/sjsonnet/src/sjsonnet/stdlib/ManifestModule.scala
@@ -130,6 +130,8 @@ object ManifestModule extends AbstractFunctionModule {
       } catch {
         case e: ujson.ParseException =>
           throw Error.fail("Invalid JSON: " + e.getMessage, pos)(ev)
+        case _: ujson.IncompleteParseException =>
+          throw Error.fail("Invalid JSON: unexpected end of JSON input", pos)(ev)
       }
     }
   }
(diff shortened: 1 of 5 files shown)


stephenamar-db:
This PR seems incorrect. a lot of files are modified.

He-Pin:
fixed

stephenamar-db (approved):

PR #1043, opened 2026-06-25
fix: improve base64 and decodeUTF8 error messages

He-Pin:
## Motivation

Invalid `std.base64Decode` and `std.decodeUTF8` inputs should produce clearer diagnostics that are closer to go-jsonnet behavior and stable for the upstream test expectations.

## Modification

- Distinguish malformed base64 lengths from decoder failures in `std.base64Decode` and `std.base64DecodeBytes`.
- Update `std.decodeUTF8` byte validation errors for non-number, non-integer, and out-of-range array elements.
- Refresh the affected golden files and `Base64Tests` assertions.

## Result

| `std.base64Decode("wrong")` | `input string appears not to be a base64 encoded string. Wrong length found (5)` | `invalid base64: Invalid input length: 5` | `[std.base64Decode] Invalid base64 string: Last unit does not have enough valid bits` | `[std.base64Decode] input string appears not to be a base64 encoded string. Wrong length found (5)` |
| `std.decodeUTF8([17.5])` | `Expected an integer, but got 17.5` | `cannot convert number with fractional part to u8` | `[std.decodeUTF8] element 0 is not an integer in range [0,255], got number` | `[std.decodeUTF8] Expected an integer, but got 17.5` |

Successful base64 and UTF-8 decoding behavior is unchanged.

## Risks

- This PR changes diagnostic text for invalid inputs; downstream tests that assert exact stderr may need to update expectations.
- It does not attempt to make jrsonnet and sjsonnet error wording identical.

### DIFF
--- a/sjsonnet/src/sjsonnet/stdlib/Base64Validation.scala
+++ b/sjsonnet/src/sjsonnet/stdlib/Base64Validation.scala
@@ -6,6 +6,20 @@ package sjsonnet.stdlib
  */
 object Base64Validation {
 
+  def hasStrictPaddingLengthError(input: String): Boolean = {
+    val len = input.length
+    if (len == 0 || len % 4 == 0) false
+    else {
+      var allAscii = true
+      var i = 0
+      while (i < len && allAscii) {
+        if (input.charAt(i) > 127) allAscii = false
+        i += 1
+      }
+      allAscii
+    }
+  }
+
   /**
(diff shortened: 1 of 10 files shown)


stephenamar-db:
Please fix the description of this PR. We dont' need the details of the LLMs.

stephenamar-db (approved):

PR #1044, opened 2026-06-25
fix: preserve negative zero in std.round

He-Pin:
## Motivation

`std.round(-0)` returned positive zero in sjsonnet, while go-jsonnet and jrsonnet preserve the negative zero sign. This matters for programs that observe signed zero through functions such as `std.atan2`.

## Modification

- Preserve the original input when `std.round` receives either `+0` or `-0`.
- Keep the existing positive and negative rounding branches for non-zero numbers.
- Add a regression test that checks the sign of rounded zero values.

## Result

| `std.round(-0)` | `-0` | `-0` | `0` | `-0` |
| `std.round(0)` | `0` | `0` | `0` | `0` |

`std.atan2(std.round(-0), -1)` now observes the negative-zero sign.

## Risks

- This intentionally changes materialized output for the exact `std.round(-0)` case.
- Other rounding behavior is intended to remain unchanged.

### DIFF
--- a/sjsonnet/src/sjsonnet/stdlib/MathModule.scala
+++ b/sjsonnet/src/sjsonnet/stdlib/MathModule.scala
@@ -318,7 +318,8 @@ object MathModule extends AbstractFunctionModule {
       // adding 0.5 would invoke IEEE 754 round-to-even and produce the wrong
       // result for odd inputs (e.g. 2^53 - 1). Short-circuit to avoid it.
       if (x != x || math.abs(x) >= 4503599627370496.0) x
-      else if (x >= 0) math.floor(x + 0.5)
+      else if (x == 0) x
+      else if (x > 0) math.floor(x + 0.5)
       else math.ceil(x - 0.5)
     },
     /**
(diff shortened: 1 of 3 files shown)


sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -78,12 +78,16 @@ final class LazyExpr(
...
-      r
+      try {
+        val r = ev.visitExpr(exprOrVal.asInstanceOf[Expr])(scope)
+        exprOrVal = r // cache result
+        scope = null.asInstanceOf[sjsonnet.ValScope] // allow GC
+        ev = null // sentinel: marks as computed
+        r
+      } finally {
+        // Allow re-evaluation on retry; same error will be thrown from exprOrVal
+        evaluating = false
```
He-Pin, line 88: ouch , it should be extracted

stephenamar-db:
please fix PR.

He-Pin:
@stephenamar-db fixed

stephenamar-db (approved):

PR #1046, opened 2026-06-25
fix: handle ujson surrogate parse exceptions in JSON imports

He-Pin:
## Motivation

Importing malformed Unicode surrogate escapes from `.json` files can make ujson throw a plain `java.lang.Exception`. That previously escaped as an sjsonnet internal error instead of falling back to normal Jsonnet parsing and reporting a parse error.

## Modification

- Catch ujson's known bare `java.lang.Exception` surrogate parse failures in the `.json` import fast path.
- Keep the catch narrow by requiring the exact exception class plus known surrogate error message prefixes.
- Add regression coverage for low-surrogate, reversed-surrogate, high-before-non-low-escape, high-before-high-escape failures, and valid surrogate pairs.

## Result

| Import JSON containing `\uDC00` as a lone low surrogate | `Truncated unicode surrogate pair escape sequence in string literal` | `syntax error: invalid string escape` | `Internal Error` caused by `Un-paired low surrogate` | `sjsonnet.ParseError` without `Internal Error` |
| Import JSON containing `\uDE00\uD83D` as a reversed surrogate pair | parse error | `syntax error: invalid string escape` | `Internal Error` caused by `Un-paired low surrogate` | `sjsonnet.ParseError` without `Internal Error` |

Valid surrogate pairs still use the JSON import fast path.

## Risks

- The catch depends on ujson's current exception type and message prefixes for surrogate failures.
- This does not add a separate pre-validation scan for every JSON import.


sjsonnet/src/sjsonnet/Importer.scala
```diff
@@ -316,9 +316,21 @@ object CachedResolver {
...
     }
   }
 
+  private def isInvalidJsonSurrogateException(e: Exception): Boolean = {
+    if (e.getClass != classOf[Exception]) return false
+    val message = e.getMessage
+    message != null && (
+      message.startsWith("Unexpected character following high surrogate") ||
+      message.startsWith("Duplicate high surrogate") ||
+      message.startsWith("Un-paired low surrogate")
```
He-Pin, line 394: this depends on ujson's exception message not that bad I think.

PR #1050, opened 2026-06-25
Make mutable scratch buffers a part of the Renderer

stephenamar-db:


He-Pin (approved):

sjsonnet/src/sjsonnet/BaseCharRenderer.scala
```diff
@@ -60,6 +54,8 @@ class BaseCharRenderer[T <: upickle.core.CharOps.Output](
     elemBuilder.writeOutToIfLongerThan(out, if (depth == 0) 0 else 1000)
   }
 
+  private val longScratchBuf: Array[Char] = new Array[Char](20)
```
He-Pin, line 57: This can't live in the object BaseByteRenderer which was wrong

stephenamar-db: it's in the class now.

PR #1052, opened 2026-06-25
Fix concurrency races on shared std module and MemberList caches (#1047)

### LINKED ISSUE #1047: Make shared interpreter state safe across multiple JVM interpreters
## Problem

Sjsonnet does not need to make a single `Interpreter` instance internally multithreaded, but it should not be thread-hostile: applications should be able to run multiple independent interpreters concurrently in the same JVM.

This is especially important because the public API accepts an external `ParseCache`; the `Interpreter` scaladoc says parsed files should be backed by a thread-safe cache (`sjsonnet/src/sjsonnet/Interpreter.scala:22-23`), and the JVM test suite already exercises multiple interpreters sharing a `ConcurrentParseCache` (`sjsonnet/test/src-jvm/sjsonnet/JsonImportFastPathJvmTests.scala:41-88`).

## Confirmed risk areas

1. Renderer-level process-wide scratch buffers

`BaseCharRenderer` and `BaseByteRenderer` keep mutable scratch buffers in companion objects:

- `sjsonnet/src/sjsonnet/BaseCharRenderer.scala:19-22` declares a reusable `Array[Char]`; `writeLongDirect` reads/writes it via `BaseCharRenderer.scratchBuf` at `sjsonnet/src/sjsonnet/BaseCharRenderer.scala:240-243`.
- `sjsonnet/src/sjsonnet/BaseByteRenderer.scala:548-551` declares the byte equivalent; `writeLongDirect` uses it at `sjsonnet/src/sjsonnet/BaseByteRenderer.scala:183-186`.

Those buffers are shared by all renderer instances in the JVM. Two interpreters rendering numeric output concurrently can write the same scratch buffer. The renderer instances themselves can stay single-threaded, but this scratch state should be instance-local, thread-local, or otherwise isolated.

2. Shared default std object has Val-level mutable caches

The default constructor uses a process-wide std module object (`sjsonnet/src/sjsonnet/Interpreter.scala:56-75`), and `StdLibModule.Default` is a singleton (`sjsonnet/src/sjsonnet/stdlib/StdLibModule.scala:135-138`). The module is a `Val.Obj` (`sjsonnet/src/sjsonnet/stdlib/StdLibModule.scala:31-34`).

`Val.Obj` has mutable inline/overflow field caches (`sjsonnet/src/sjsonnet/Val.scala:1910-1926`) and `value()` reads those caches without synchronization (`sjsonnet/src/sjsonnet/Val.scala:2352-2374`). That means ordinary lookups like `std.map` can mutate shared `Val.Obj` state across otherwise independent interpreters.

3. ParseCache can share AST/Val objects whose lazy caches are mutable

`CachedResolver.parse` stores `(Expr, FileScope)` in the external `ParseCache` (`sjsonnet/src/sjsonnet/Importer.scala:239-252`), and `Evaluator.visitImport` directly evaluates the cached `Expr` (`sjsonnet/src/sjsonnet/Evaluator.scala:1292-1302`). With a shared, thread-safe `ParseCache`, those AST nodes can be observed by multiple interpreters at the same time.

There are mutable lazy caches on those shared objects:

- `Expr.ObjBody.MemberList` has volatile caches for sorted order and no-self-ref, but the shared key-name arrays are plain vars (`sjsonnet/src/sjsonnet/Expr.scala:440-455`).
- `Evaluator.visitMemberList` writes `_cachedSortedOrder` on the shared `MemberList` and copies it onto each object (`sjsonnet/src/sjsonnet/Evaluator.scala:1896-1903`).
- `Val.Obj` writes sorted key caches and shared key-name caches (`sjsonnet/src/sjsonnet/Val.scala:1955-1965`, `sjsonnet/src/sjsonnet/Val.scala:2296-2348`).
- JSON import fast path can cache `Val.Obj` directly in the parse cache; it deliberately disables field-value caching but still pre-populates `_sortedInlineOrder` (`sjsonnet/src/sjsonnet/Importer.scala:372-395`).

4. Val-level lazy materialization caches are not thread-safe if values are shared

Several lazy `Val` structures update mutable fields without synchronization:

- `Val.Str.str` flattens ropes by writing `_str` and clearing `_children` (`sjsonnet/src/sjsonnet/Val.scala:340-360`).
- `LazyIndexedArr` lazily allocates/writes `slots` (`sjsonnet/src/sjsonnet/Val.scala:828-852`).
- Lazy view arrays update `values`, `evals`, `computedCount`, and then publish `arr` (`sjsonnet/src/sjsonnet/Val.scala:1210-1268`).

These are fine for values owned by one interpreter/evaluation. They become unsafe if any parse-cache/shared-stdlib path makes the same `Val` reachable by multiple interpreters.

## Existing verification

I ran the current targeted concurrent-import test on this checkout:

```bash
./mill 'sjsonnet.jvm[3.3.8]'.test.testOnly sjsonnet.JsonImportFastPathJvmTests
```

Result: `Tests: 1, Passed: 1, Failed: 0`.

This verifies the intended usage shape exists, but the test only covers a narrow JSON-import case and does not exercise the renderer scratch buffers, shared std object field cache, or the broader Val-level lazy caches.

## Proposed direction

- Keep the contract that an individual `Interpreter` is single-threaded.
- Ensure all mutable execution/rendering state is owned by an `Interpreter`, owned by an evaluation/render call, or protected when intentionally process-wide.
- Move renderer scratch buffers from companion objects to renderer instances or thread-local storage.
- Avoid using a process-wide mutable `Val.Obj` as the default std object, or make the std object immutable/no-cache for shared use.
- Audit caches attached to shared parse-cache contents. If AST/JSON `Val` objects can be shared across interpreters, lazy cache initialization should be safe under concurrent access, or those cached values should be cloned/owned per interpreter before mutation.
- Clarify `ParseCache` documentation: `DefaultParseCache` is not thread-safe (`sjsonnet/src/sjsonnet/ParseCache.scala:27-36`), but a user-provided shared thread-safe parse cache should be supported without exposing races in cached contents.

## Acceptance criteria

- Add a JVM stress test that runs multiple interpreters concurrently with a shared thread-safe `ParseCache`, covering regular Jsonnet imports, strict JSON imports, stdlib lookups, object field/key materialization, and renderer output with many integer numbers.
- No process-wide mutable renderer buffers remain.
- Shared default std usage does not mutate a shared `Val.Obj` field cache.
- Shared parse-cache entries either contain immutable data, use thread-safe lazy initialization, or are copied before interpreter-owned mutation.

stephenamar-db:
When independent Interpreters run concurrently (e.g. a Bazel bundle worker fanning evaluation across a thread pool) they share the process-wide default `std` module and, via the parse cache, the StaticOptimizer-folded AST. Two pieces of shared mutable state raced:

1. The default `std` module is a singleton built with `static = false`, so the first `std.foo` lookup (which StaticOptimizer triggers while folding) called `putCache`, mutating the inline cache (ck1/cv1) and lazily allocating the `valueCache` HashMap. Concurrent first access could return the wrong builtin or corrupt the map. Add `Val.Obj.mkWithConstCache`, which pre-fills the value cache for every constant field at construction; `StdLibModule.module` now uses it, making lookups read-only. Context-dependent members (thisFile/pi) carry `cached = false` and never write either, so the instance is immutable after construction.

2. `Expr.ObjBody.MemberList._cachedAllKeyNames` / `_cachedVisibleKeyNames` are shared across interpreters via the parse cache but were non-volatile; an unsafe publication could expose a half-filled array. Make both `@volatile`, matching the existing `_cachedSortedOrder` / `_noSelfRef` pattern.

Adds regression tests to ParallelManifestRaceTests: a deterministic guard that the std value cache is pre-filled at construction, plus concurrent stress tests for std first-access and shared-MemberList materialization. Also fixes the test file to use java.lang.AssertionError (utest._ shadows it).

This builds on #1050 (renderer scratch buffers) and the static-object sorted-key cache fix, which together cover the data races in #1047 that are reachable under the supported concurrency model (independent interpreters).

### DIFF
--- a/sjsonnet/src/sjsonnet/Expr.scala
+++ b/sjsonnet/src/sjsonnet/Expr.scala
@@ -451,8 +451,8 @@ object Expr {
        * == null). Computed from the first instance and reused by subsequent ones, avoiding per-object
        * allKeyNames/visibleKeyNames allocation.
        */
-      var _cachedAllKeyNames: Array[String] = null
-      var _cachedVisibleKeyNames: Array[String] = null
+      @volatile var _cachedAllKeyNames: Array[String] = null
+      @volatile var _cachedVisibleKeyNames: Array[String] = null
 
       /** True if all fields have compile-time fixed names (no computed/Dyn field names). */
       lazy val allFieldNamesFixed: Boolean =
--- a/sjsonnet/src/sjsonnet/Val.scala
+++ b/sjsonnet/src/sjsonnet/Val.scala
@@ -1855,6 +1855,37 @@ object Val {
       }
       new Obj(pos, m, false, null, null)
     }
+
+    /**
+     * Like [[mk]], but eagerly caches the value of every constant ([[ConstMember]]) field at
+     * construction so that subsequent field lookups never mutate the instance.
+     *
+     * This is for the process-wide shared `std` module: an [[Interpreter]]'s default `std` is a
+     * singleton ([[sjsonnet.stdlib.StdLibModule.Default]]) read concurrently by independent
+     * interpreters. On the normal lazy path the first `std.foo` lookup calls `putCache`, racing on
+     * the inline cache (`ck1/cv1`) and the lazily-allocated `valueCache` HashMap across threads —
+     * which can return the wrong builtin or corrupt the map (#1047). Pre-filling makes every lookup
+     * of a constant field a read-only cache hit. Non-constant members (e.g. `std.thisFile`, whose
+     * value depends on the file being evaluated) are intentionally left uncached; they carry
+     * `cached = false` and so never write on lookup either, leaving the instance immutable — and
+     * thus safe to publish across threads — after construction.
+     */
+    def mkWithConstCache(
+        pos: Position,
+        sizeHint: Int,
+        membersArray: Iterable[(String, Obj.Member)]*): Obj = {
+      val m = Util.preSizedJavaLinkedHashMap[String, Obj.Member](sizeHint)
+      val cache = Util.preSizedJavaHashMap[Any, Val](sizeHint)
+      for (members <- membersArray; (k, v) <- members) {
+        m.put(k, v)
+        v match {
+          // ConstMember.invoke ignores its arguments and returns its constant value.
+          case cm: ConstMember if cm.cached => cache.put(k, cm.invoke(null, null, null, null))
+          case _                            =>
+        }
+      }
+      new Obj(pos, m, false, null, null, cache, null)
+    }
   }
 
   /**
@@ -1965,6 +1996,14 @@ object Val {
       }
       r
     }
+
+    /**
+     * Test-only: the already-cached value for `k`, or null if not cached. Used to assert that
+     * shared modules built via [[Obj.mkWithConstCache]] (e.g. the process-wide `std`) pre-populate
+     * their value cache at construction, so concurrent first access only ever reads it (#1047).
+     */
+    private[sjsonnet] def cachedValueForTest(k: Any): Val =
+      if (valueCache != null) valueCache.get(k) else null
 
     /**
      * When true, field caching can be skipped during materialization because no field body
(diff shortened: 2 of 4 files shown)


stephenamar-db:
@He-Pin

PR #1053, opened 2026-06-25
Revert "fix: std.format rejects boolean for numeric conversion codes (#1016)"

stephenamar-db:
This reverts commit db561c9f9ade27ad63687bfe59ea378c3a2bf3bd.

https://jsonnet.org/ref/stdlib.html#std-format is the standard we should follow, evne if go-jsonnet is wrong.


He-Pin:
@stephenamar-db Would you enable this behind a flag?

stephenamar-db:
send me a roll-forward with a flag. I'll approve

He-Pin:
OK,thanks.

PR #1054, opened 2026-06-26
Harden shared parse cache

He-Pin:
## Summary

- replace `DefaultParseCache`'s non-thread-safe mutable `HashMap` with a platform cache store
- use `TrieMap` on JVM and Scala Native, while keeping a single-threaded `HashMap` shim for Scala.js/Wasm where `TrieMap` does not link
- keep the existing `ParseCache` interface and store parsed `Either[Error, (Expr, FileScope)]` values directly, without `LazyVal` or additional wrapper allocation
- add an end-to-end concurrent regression test for independent interpreters sharing a cold `DefaultParseCache`

## Notes

Concurrent cold misses for the same key may now duplicate parse work, but `putIfAbsent` ensures the cache publishes one winner and avoids mutable map corruption. The parse path is deterministic CPU work, so this keeps the fix focused on cache correctness without adding a lazy wrapper abstraction.

## Verification

- `git diff --check`
- `./mill "sjsonnet.jvm[3.3.8].checkFormat"`
- `./mill "sjsonnet.jvm[3.3.8].test.testOnly" sjsonnet.ParallelManifestRaceTests` — 8/8 passed
- `./mill "sjsonnet.js[2.13.18].test.fastLinkJSTest"`
- `./mill "sjsonnet.wasm[2.13.18].test.fastLinkJSTest"`
- `./mill "sjsonnet.js[3.3.8].test.fastLinkJSTest"`
- `./mill "sjsonnet.wasm[3.3.8].test.fastLinkJSTest"`
- `./mill "sjsonnet.native[3.3.8].compile"`
- `./mill "sjsonnet.native[2.13.18].compile"`
- `./mill "sjsonnet.jvm[2.13.18].compile"`
- Qoder stdout review: no must-fix findings
- GitHub Pull Request Validation: 5/5 checks passed

### DIFF
--- a/sjsonnet/src-jvm/sjsonnet/Platform.scala
+++ b/sjsonnet/src-jvm/sjsonnet/Platform.scala
@@ -14,12 +14,20 @@ import org.yaml.snakeyaml.{DumperOptions, LoaderOptions, Yaml}
 import org.yaml.snakeyaml.nodes.{MappingNode, Node, ScalarNode, SequenceNode, Tag}
 
 import scala.annotation.nowarn
+import scala.collection.concurrent.TrieMap
 import scala.collection.compat.*
 import scala.collection.mutable
 import scala.jdk.CollectionConverters.*
 
 object Platform {
   private val hexFormat = HexFormat.of()
+
+  private[sjsonnet] type ParseCacheMap =
+    TrieMap[(Path, String), Either[Error, (Expr, FileScope)]]
+  private[sjsonnet] type ParseCacheValue = Either[Error, (Expr, FileScope)]
+
+  def newParseCacheMap(): ParseCacheMap =
+    TrieMap.empty[(Path, String), ParseCacheValue]
 
   def repeatString(s: String, count: Int): String =
     if (count <= 0) "" else s.repeat(count)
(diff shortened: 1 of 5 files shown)


sjsonnet/src-jvm/sjsonnet/Platform.scala
```diff
@@ -14,13 +14,21 @@ import org.yaml.snakeyaml.{DumperOptions, LoaderOptions, Yaml}
...
+import scala.collection.concurrent.TrieMap
 import scala.collection.compat.*
 import scala.collection.mutable
 import scala.jdk.CollectionConverters.*
 
 object Platform {
   private val hexFormat = HexFormat.of()
 
+  private[sjsonnet] type ParseCacheMap =
+    TrieMap[(Path, String), Util.LazyVal[Either[Error, (Expr, FileScope)]]]
```
He-Pin: In Scala's default implementation, I think using TrieMap is thread-safe. I was using caffeine

He-Pin: Should we add a TinyLRU?

stephenamar-db (approved):

PR #1055, opened 2026-06-26
LazyExpr poisoning — a throwing compute left the re-entrancy flag stuck.

stephenamar-db:
try/finally resets it so a shared/reused thunk recomputes instead of falsely reporting self-reference.

### DIFF
--- a/sjsonnet/src/sjsonnet/Val.scala
+++ b/sjsonnet/src/sjsonnet/Val.scala
@@ -78,12 +78,15 @@ final class LazyExpr(
       Error.fail("Infinite recursion detected (self-referential thunk)")
     } else {
       evaluating = true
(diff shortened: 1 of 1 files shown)


sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -78,12 +78,15 @@ final class LazyExpr(
...
-      r
+      try {
+        val r = ev.visitExpr(exprOrVal.asInstanceOf[Expr])(scope)
+        exprOrVal = r // cache result
+        scope = null.asInstanceOf[sjsonnet.ValScope] // allow GC
+        ev = null // sentinel: marks as computed
+        r
+      } finally {
+        evaluating = false
+      }
```
He-Pin, line 89: nice catch

PR #1058, opened 2026-06-27
Harden remaining shared interpreter state

He-Pin:
## Motivation

Issue #1047 fixed the main shared-interpreter race paths, but a broader audit found remaining shared mutable state in binary resolved files, function module objects, and StaticOptimizer position rewriting. Independent `Interpreter` instances can share process-wide values such as `std`, so optimization must not mutate shared `Val.pos` state.

A follow-up review also found that preserving the folded outer expression position by cloning `Val` literals made the optimizer safe, but added avoidable allocations. For branch and boolean short-circuit folds, the better rule is to return the selected existing expression/value and keep its own position, rather than clone it or replace it with an unrelated singleton.

## Modification

- **StaticBinaryResolvedFile**: Convert from `case class` to manual `Product` with constructor-level defensive copy, content-based `equals`/`hashCode`, and `lazy val` `contentHash` computed once via `Platform.hashBytes`.
- **Platform.hashBytes (JS)**: Add MurmurHash3-based implementation to match native binary file hashing support.
- **StaticOptimizer**: Remove folded-position rewrites for branch and boolean short-circuit folding. The optimizer now reuses selected child expressions. For `false && rhs` and `true || rhs`, it returns the original lhs boolean, preserving its position and avoiding both cloning and discarding an already-created literal/folded lhs.
- **FunctionModule.mkWithConstCache**: Pre-fill const caches at construction, eliminating lazy cache population races on shared `std` and sub-module objects.
- **Tests**: Cover binary file API behavior, shared `Val` position immutability with literal reuse, short-circuit lhs boolean reuse, function module cache pre-fill, and the accepted recursive import error-position change.

## Result

Independent interpreters sharing the `std` singleton no longer corrupt each other through StaticOptimizer position mutation or lazy cache population. StaticOptimizer avoids the extra literal allocation that `withFoldedPos` introduced, and the lhs-result short-circuit cases now preserve the original lhs boolean position while avoiding avoidable garbage. `StaticBinaryResolvedFile` has content-based equality and cached hashing.

## Risks

- Folded-expression error stack locations can now point at the selected child expression rather than the folded outer expression. The recursive import golden changes from `17:1` to `17:15`, which points at the actual `import` expression and is acceptable for this optimization.
- The optimizer intentionally relies on shared literal immutability by never mutating reused `Val.pos`; regression tests assert reused singletons, cached literals, and short-circuit lhs booleans keep their original positions.

## Validation

- `./mill sjsonnet.jvm[3.3.8].test`: 143/143, SUCCESS
- `./mill sjsonnet.js[3.3.8].test.fastLinkJSTest`: 157/157, SUCCESS
- `./mill sjsonnet.wasm[3.3.8].test.fastLinkJSTest`: 157/157, SUCCESS
- `./mill sjsonnet.native[3.3.8].compile`: 68/68, SUCCESS
- `./mill sjsonnet.jvm[2.12.21].compile`: 61/61, SUCCESS
- `./mill sjsonnet.jvm[2.13.18].compile`: 61/61, SUCCESS
- `./mill __.checkFormat`: 36/36, SUCCESS
- `git diff --check upstream/master..HEAD`: SUCCESS

## References

- https://github.com/databricks/sjsonnet/issues/1047

### DIFF
--- a/sjsonnet/src/sjsonnet/StaticOptimizer.scala
+++ b/sjsonnet/src/sjsonnet/StaticOptimizer.scala
@@ -138,11 +138,11 @@ class StaticOptimizer(
       case e @ UnaryOp(pos, op, v: Val) => tryFoldUnaryOp(pos, op, v, e)
 
       // Branch elimination: constant condition in if-else
-      case IfElse(pos, _: Val.True, thenExpr, _) =>
-        thenExpr.pos = pos; thenExpr
-      case IfElse(pos, _: Val.False, _, elseExpr) =>
-        if (elseExpr == null) Val.Null(pos)
-        else { elseExpr.pos = pos; elseExpr }
+      case IfElse(_, _: Val.True, thenExpr, _) =>
+        thenExpr
+      case IfElse(_, _: Val.False, _, elseExpr) =>
+        if (elseExpr == null) Val.staticNull
+        else elseExpr
 
       // Short-circuit elimination for And/Or with constant lhs.
       //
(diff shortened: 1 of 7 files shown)


sjsonnet/src/sjsonnet/StaticOptimizer.scala
```diff
@@ -152,10 +152,12 @@ class StaticOptimizer(
...
-      case And(pos, _: Val.True, rhs: Val.Bool) => rhs.pos = pos; rhs
-      case And(pos, _: Val.False, _)            => Val.False(pos)
-      case Or(pos, _: Val.True, _)              => Val.True(pos)
-      case Or(pos, _: Val.False, rhs: Val.Bool) => rhs.pos = pos; rhs
+      // The lhs-result cases return lhs itself to keep the selected boolean's position and avoid
+      // turning the already-created literal/folded lhs into garbage.
+      case And(_, _: Val.True, rhs: Val.Bool) => rhs
+      case And(_, lhs: Val.False, _)          => lhs
+      case Or(_, lhs: Val.True, _)            => lhs
+      case Or(_, _: Val.False, rhs: Val.Bool) => rhs
```
He-Pin, line 160: switch to inner pos and avoid an allocation

stephenamar-db (approved):

PR #1059, opened 2026-06-27
Fix performance regression with std.range and regenerate bench goldens

stephenamar-db:
StaticOptimizer constant folding forced the materialization of std.range. This is not the intention.


sjsonnet/src/sjsonnet/stdlib/ArrayModule.scala
```diff
@@ -616,6 +616,11 @@ object ArrayModule extends AbstractFunctionModule {
   }
 
   private object Range extends Val.Builtin2("range", "from", "to") {
+    // Not static-safe: constant folding would call asStrictArray on the result, eagerly
+    // materializing the lazy RangeArr at optimization time. That defeats the lazy RangeArr /
+    // ConcatView / sharedConcatPrefixLength fast path (e.g. `std.range(1, N) + [x] < ... + [y]`
+    // collapses from O(1) to O(N) materialization on every evaluation).
+    override def staticSafe: Boolean = false
```
He-Pin, line 623: Should this method defaul tto `false` in the super?

PR #1060, opened 2026-06-28
fix: O(N^2) materialize for large imported JSON objects with a super

stephenamar-db:
#840 builds imported strict-JSON objects with an uncapped inline-array Val.Obj layout. When such a large object gains a `super` (e.g. via a `+` merge), materialize can't use the direct inline-iteration fast path and uses a linear scan instead.

Adds a regression guard under bench/resources/sjsonnet_suite.


stephenamar-db:
@He-Pin

sjsonnet/src/sjsonnet/Val.scala
```diff
@@ -1815,6 +1815,16 @@ object Val {
...
 
+    /**
+     * Field count at/below which `valueRaw` resolves an inline-array object by linear scan. Above
+     * it (e.g. large imported strict-JSON objects, whose inline arrays are uncapped — see
+     * Importer), `valueRaw` instead lazily builds the `value0` LinkedHashMap via `getValue0` for
+     * O(1) lookups. This keeps the scan for small objects (and never builds a map for the
+     * import-and-emit path, which materializes via direct inline iteration) while avoiding O(N^2)
+     * when a large imported object gains a `super` and must be resolved per-key.
+     */
+    private[sjsonnet] final val InlineScanMax = 8
```
He-Pin, line 1826: Obj.InlineScanMaxFields?

He-Pin (approved):

sjsonnet/src/sjsonnet/Val.scala
stephenamar-db: feel free to rename it in your PR. I also noticed the size 8 for objects is also used in ObjectModule. we should consolidate.

PR #1061, opened 2026-06-28
fix: avoid linear objectHas scans for large inline imports

He-Pin:
Follow-up to #1060.

Motivation:
PR #1060 routes large inline value lookups through `getValue0`, but `std.objectHas` and `std.objectHasAll` still reach `containsVisibleKey` / `containsKey` and scan `inlineFieldKeys` without the threshold. Repeated membership checks over imported JSON keys can therefore stay O(N^2).

Changes:
- Apply `Obj.InlineScanMax` to `containsKey` and `containsVisibleKey`
- Add a bench regression for repeated `objectHas` / `objectHasAll` over imported JSON fields

Validation:
- `./mill bench.runRegressions bench/resources/sjsonnet_suite/json_import_object_has.jsonnet` -> `19.909 ms/op`
- PR #1060 before this follow-up on the same fixture -> `89.634 ms/op`
- `./mill __.checkFormat`
- `./mill "sjsonnet.jvm[3.3.8]".test`
- Qoder stdout review: no must-fix findings


stephenamar-db:
this PR closed by mistake.

He-Pin:
OK I will resend it.

PR #1085, opened 2026-07-04
Reject unpaired JSON surrogate escapes

He-Pin:
## Motivation

`std.parseJson` and the strict `.json` import fast path could let JSON strings containing lone UTF-16 surrogate halves through as Jsonnet strings. The proper fix belongs in ujson upstream; this PR adds a downstream guard.

## Modification

- Add `ValVisitor.rejectUnpairedSurrogates` for decoded JSON string values and object keys.
- Enable for `std.parseJson` and the strict `.json` import path.
- Preserve valid surrogate pairs and escaped literal text (`"\\uD800"`).
- Add regression coverage for lone high/low surrogates, surrogate keys, valid pairs, and escaped literals.

## Result

| Case | before | after | cpp-jsonnet | go-jsonnet | jrsonnet |
|---|---|---|---|---|---|
| `std.parseJson("\"\\uD800\"")` | `55296` | error | error | `65533` | error |
| `std.parseJson("\"\\uDC00\"")` | `56320` | error | error | `65533` | error |
| `std.parseJson("\"\\uD83D\\uDE00\"")` (valid pair) | `128512` | `128512` | `128512` | `128512` | `128512` |
| surrogate object key | `55296` | error | error | `65533` | error |

Aligns with cpp-jsonnet and jrsonnet. go-jsonnet normalizes to U+FFFD (see google/go-jsonnet#885).

## References

- [ujson upstream issue](https://github.com/com-lihaoyi/upickle/issues/722)
- [RFC 8259 §8.2 surrogate hazards](https://www.rfc-editor.org/rfc/rfc8259#section-8.2)


He-Pin:
refs: https://github.com/com-lihaoyi/upickle/pull/723

PR #1112, opened 2026-08-06
fix: std.parseJson replaces lone surrogates with U+FFFD instead of corrupting to '?'

He-Pin:
## Motivation

`std.parseJson` keeps unpaired UTF-16 surrogates in the parsed string, but every UTF-8 output path then replaces the unencodable unit with `?`:

| | `std.parseJson('"\\ud800"')` output bytes |
|---|---|
| sjsonnet (before) | `22 3f 22` — `"?"` |
| sjsonnet (after) | `22 ef bf bd 22` — `"\uFFFD"` |
| go-jsonnet v0.22.0 | `22 ef bf bd 22` — `"\uFFFD"` |
| jrsonnet | errors |
| CPython / JS | preserve the surrogate |

No implementation emits `?`. sjsonnet already follows the U+FFFD policy elsewhere — `std.char` and `%c` replace surrogate codepoints with U+FFFD (`StringModule.scala`, `Format.scala`), and `UnicodeHandlingTests` documents that alignment — so `std.parseJson` contradicted the project's own convention and silently corrupted data on output.

## Modification

`ValVisitor` (used by `std.parseJson` and `std.parseYaml`) now sanitizes string values and object keys through a new `ValVisitor.replaceLoneSurrogates` helper:

- fast-path scan returns the input **unchanged, without allocation**, when it contains no surrogates (the overwhelmingly common case);
- otherwise each unpaired high/low surrogate is replaced with U+FFFD while valid surrogate pairs are preserved.

## Result

- `std.parseJson('"\\ud800"') == "\ufffd"` — byte-identical with go-jsonnet.
- Valid pairs unchanged: `std.parseJson('"\\ud83d\\ude00"')` still yields the emoji.
- Doubled/mid-string lone surrogates handled (`"\ud800\ud800"` → two U+FFFD, `"a\udc00b"` → `"a\ufffdb"`).
- Object keys sanitized: `std.parseJson('{"\\ud800": 1}')["\ufffd"] == 1`.

**JSON imports unaffected:** the `.json` import path uses ujson's byte parser, which already drops (in values) or rejects (in keys) lone surrogates before any visitor runs — pre-existing behavior, unchanged by this fix.

## Test plan

- [x] New: `new_test_suite/parseJson_lone_surrogate.jsonnet` (+ golden) covering lone high, lone low, doubled, mid-string, valid-pair preservation, and key sanitization
- [x] `./mill 'sjsonnet.jvm[_].test'` — all Scala versions (2.12.21 / 2.13.18 / 3.3.8) pass
- [x] `./mill 'sjsonnet.js[_].compile' 'sjsonnet.native[_].compile' 'sjsonnet.wasm[_].compile'` — pass

Found by four-way differential testing (sjsonnet vs go-jsonnet vs jrsonnet vs spec).


He-Pin:
Duplicate of #1085, which I missed before opening this — apologies for the noise.

The two PRs make opposite design choices for the same problem:

| | #1085 | #1112 (this one) |
|---|---|---|
| behavior | reject lone surrogates with an error | replace with U+FFFD |
| matches | cpp-jsonnet (error) + jrsonnet (error) — 2 of 3 | go-jsonnet — 1 of 3 |
| coverage | std.parseJson + strict .json import fast path + Preloader | std.parseJson / std.parseYaml via ValVisitor |

Deferring to #1085: rejecting aligns with the C++ reference and jrsonnet, and covers the import paths this PR left alone (the ujson ByteParser drops/rejects lone surrogates before the visitor on imports, which #1085's fast-path guard addresses deliberately).

One data point that may be useful for #1085's review: sjsonnet's std.char / %c replace surrogate codepoints with U+FFFD (StringModule.scala, Format.scala, documented in UnicodeHandlingTests) — so with #1085's reject policy, parseJson would be stricter than char/%c. That asymmetry is defensible (parseJson validates input, char/%c construct output) but worth a sentence in the PR or a code comment.

PR #1120, opened 2026-09-03
fix: apply addition semantics to object extension

He-Pin:
## Motivation

Jsonnet defines `lhs { body }` as syntactic sugar for `lhs + { body }`. Previously, sjsonnet required an object on the left and rejected valid expressions such as `"hello" { x: 1 }` with `Expected object, found string`. This expression should return `"hello{\"x\": 1}"`, just like explicit addition.

## Modification

- Preserve the existing object-inheritance fast path, including `self`, `super`, assertions, and lazy fields.
- For a non-object left operand, delegate to the existing `BinaryOp +` evaluator using the already evaluated value. This evaluates the original left expression once and shares object construction, stringification, and error reporting with explicit addition.
- Add regression coverage for all left-operand types, static and dynamic object bodies, comprehensions, hidden fields, evaluation order, lazy errors, inheritance, shared object-body caches, and the `std.foldl` fallback. Include executable number/array error goldens and 14 success assertions.

## Result

String/object extension now follows normal addition semantics. Unsupported left-operand types still fail, with the same diagnostics as `+`. The existing strict-mode restriction on consecutive object bodies is preserved.

Validation completed locally:

- JVM Scala 2.12.21, 2.13.18, and 3.3.8: **679/679 tests passed per version (2,037 total)**.
- Formatting checks passed for all JVM Scala versions; the Scala 3.3.8 assembly built successfully.
- The assembled CLI matched the success, number-error, and array-error golden files, including exit codes and diagnostics.
- Text and flamegraph profiling both produced the expected object-extension results and nonempty profile output.
- The 14 success assertions passed on Go Jsonnet 0.21.0 and 0.22.0.

```sh
./mill --no-daemon -j 1 'sjsonnet.jvm[_].test' + 'sjsonnet.jvm[_].checkFormat' + 'sjsonnet.jvm[3.3.8].assembly'
```

JS, WASM, Native, and JMH were not run locally.

## References

- [Jsonnet specification: desugaring](https://jsonnet.org/ref/spec.html#desugaring)
- [Corresponding jrsonnet fix](https://github.com/deltarocks/jrsonnet/commit/b5fcc2639d612b852b153474d80b41c537c18e2a)


stephenamar-db:
That feels like a bug in jsonnet, no?

He-Pin:
@stephenamar-db https://jsonnet.org/ref/spec.html#desugaring Agreed that it looks surprising. The specification explicitly desugars e { ... } to e + { ... }, and + converts the other operand to a string when either operand is a string. I see both CPP and Go implementations follows this rule.

